PR #20, opened 2015-09-23
closes #19, removed dependency on heketi

### LINKED ISSUE #19: Remove heketi dependency
heketi is pulled in as a dependency for tests, just to make use of an Assert function.

krisis:
Removed the dependency on github.com/ heketi/heketi just for the
Assert() function. 'Borrowed' the implementation from
github.com/heketi/heketi pkg with attribution placed in tests/assert.go

Signed-off-by: Krishnan Parthasarathi <kparthas@redhat.com>

### DIFF
--- a/volume/volume_test.go
+++ b/volume/volume_test.go
@@ -3,8 +3,7 @@ package volume
 import (
 	"testing"
 
-	"github.com/heketi/heketi/tests"
-	"github.com/heketi/heketi/utils"
+	"github.com/kshlm/glusterd2/tests"
 )
 
 func TestNewVolumeEntry(t *testing.T) {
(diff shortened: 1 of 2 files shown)


volume/volume_test.go
```diff
@@ -82,12 +81,23 @@ func TestNewVolumeEntryFromRequestRedundancy(t *testing.T) {
...
 func TestNewVolumeEntryFromRequestBricks(t *testing.T) {
 
 	req := VolCreateRequest{}
 	req.Bricks = []string{"abc", "def"}
 
 	v := NewVolumeEntry(req)
-	tests.Assert(t, utils.SortedStringHas(v.Bricks, "abc"))
-	tests.Assert(t, utils.SortedStringHas(v.Bricks, "def"))
+	tests.Assert(t, find(v.Bricks, "abc"))
+	tests.Assert(t, find(v.Bricks, "def"))
```
kshlm: I think it would be better if you just looped over the req.Bricks slice.

atinmu: req.Bricks is the input here and the intention of this unit test it to find whether the bricks are propagated in volume object. So I feel this is correct.

krisis: @atinmu Iterating over ``req.Bricks`` using ``range`` doesn't change the tests performed, they just change the way it is expressed.

atinmu: @krisis : Now I got it after looking at the change, sorry for the confusion.

PR #24, opened 2015-10-15
handle volume create command

### LINKED ISSUE #14: volume create to work on a single node cluster

atinmu:
volume create support on a single node cluster
Closes #14
This patch adds the following changes:
- Add utility functions in utils/util.go required for volume create validation
- Introduce Brickinfo object
- return errors instead of panics in volgen packages
- add consul as backend and register for a temporary purpose

Note: this patch is still WIP, the tests are not stable.


client/struct.go
```diff
@@ -14,17 +14,24 @@ type GenericJSONResponse struct {
 	Data     interface{}
 }
 
-// SendResponse Warapper function to send Generic JSON response back to the rest client
-func SendResponse(w http.ResponseWriter, opRet int, opErrno int, opErrStr string, status int, v interface{}) {
-	var rsp GenericJSONResponse
+// FormResponse Warapper function to construct Generic JSON response to be sent
```
kshlm: Why did you split this into two? It was much simpler just calling SendResponse once, instead of making two calls.

commands/volume-create/volume-create.go
```diff
@@ -20,46 +20,71 @@ import (
...
-
-	var msg volume.VolCreateRequest
-
-	e := utils.GetJSONFromRequest(r, &msg)
+func validateVolumeCreateRequest(msg *volume.VolCreateRequest, r *http.Request) *client.GenericJSONResponse {
+	e := utils.GetJSONFromRequest(r, msg)
 	if e != nil {
-		client.SendResponse(w, -1, 422, errors.ErrJSONParsingFailed.Error(), 422, "")
-		return
+		log.Error("Invalid JSON Request")
```
kshlm: Please also log `e` using log.WithFields.

commands/volume-create/volume-create.go
```diff
@@ -20,46 +20,71 @@ import (
...
 		log.Error("Brick list is empty")
-		client.SendResponse(w, -1, http.StatusBadRequest, errors.ErrEmptyBrickList.Error(), http.StatusBadRequest, "")
-		return
+		return client.FormResponse(-1, http.StatusBadRequest, errors.ErrEmptyBrickList.Error(), "")
 	}
+	return nil
+
+}
+
+func createVolume(msg *volume.VolCreateRequest) *volume.Volinfo {
```
kshlm: This and the function below, sound too similar and are confusing. Please rename one of the two.

commands/volume-create/volume-create.go
```diff
@@ -20,46 +20,71 @@ import (
...
+	vol := volume.NewVolumeEntry(msg)
+	return vol
+}
+
+func (c *Command) volumeCreate(w http.ResponseWriter, r *http.Request) {
```
kshlm: You could rename this to volumeCreateHandler

commands/volume-create/volume-create.go
```diff
@@ -20,46 +20,71 @@ import (
...
+
+	msg := new(volume.VolCreateRequest)
 
+	rsp := validateVolumeCreateRequest(msg, r)
+	if rsp != nil {
+		client.SendResponse(w, http.StatusBadRequest, rsp)
+		return
+	}
 	if context.Store.VolumeExists(msg.Name) {
```
kshlm: You've changed all other context.Store.Vol* functions to volume.* functions but missed this.

commands/volume-create/volume-create.go
```diff
@@ -20,46 +20,71 @@ import (
...
 		log.WithField("Volume", msg.Name).Error("Volume already exists")
```
kshlm: Keep the field keys in lower-case.

commands/volume-create/volume-create.go
```diff
@@ -20,46 +20,71 @@ import (
...
 		return
 	}
 
-	vol := volume.NewVolumeEntry(msg)
-	e = context.Store.AddOrUpdateVolume(vol)
+	//TODO : Error handling for volgen
+	// Creating client  and server volfile
+	e := volgen.GenerateVolfile(vol)
 	if e != nil {
-		log.WithField("error", e).Error("Couldn't add volume to store")
```
kshlm: Retain the log here.

commands/volume-create/volume-create.go
```diff
@@ -20,46 +20,71 @@ import (
...
+		rsp = client.FormResponse(-1, http.StatusInternalServerError, e.Error(), "")
+		client.SendResponse(w, http.StatusInternalServerError, rsp)
 		return
 	}
 
-	// Creating client  and server volfile
-	volgen.GenerateVolfile(vol)
+	e = volume.AddOrUpdateVolume(vol)
+	if e != nil {
+		rsp = client.FormResponse(-1, http.StatusInternalServerError, e.Error(), "")
```
kshlm: Log the error here.

commands/volume-delete/volume-delete.go
```diff
@@ -20,23 +21,24 @@ type Command struct {
 
 func (c *Command) volumeDelete(w http.ResponseWriter, r *http.Request) {
 	p := mux.Vars(r)
-	/* TODO : As of now we consider the request as volname, later on we need
-	* to consider the volume id as well */
 	volname := p["volname"]
 
 	log.Info("In Volume delete API")
 
 	if context.Store.VolumeExists(volname) {
```
kshlm: Same comment as before.

store/volume.go
```diff
@@ -4,9 +4,7 @@ package store
 // volumes stored in the store
```
kshlm: It's better to remove this file and the functions here all together. Having empty wrapper functions doesn't serve any purpose.

utils/utils.go
```diff
@@ -0,0 +1,189 @@
...
+				return found
+			}
+		}
+	}
+	return found
+}
+
+// ParseHostAndBrickPath parses the host & brick path out of req.Bricks list
+func ParseHostAndBrickPath(brickPath string) (string, string) {
+	idx := strings.LastIndex(brickPath, ":")
```
kshlm: Just use `i` for index. It's shorter, and is much easier when reading. Also, everyone understands what `i` is.

utils/utils.go
```diff
@@ -0,0 +1,189 @@
...
+		return "", ""
+	}
+	hostname := brickPath[0:idx]
+	path := brickPath[idx+1 : len(brickPath)]
+
+	return hostname, path
+}
+
+//ValidateBrickPathLength validates the length of the brick path
+func ValidateBrickPathLength(brickPath string) int {
```
kshlm: I don't think we need to be validating path lengths. As we discussed, it's better to leave that to the underlying OS.

utils/utils.go
```diff
@@ -0,0 +1,189 @@
...
+	if len(filepath.Clean(brickPath)) >= syscall.PathMax {
+		log.Error("brickpath is too long")
+		return -1
+	}
+	return 0
+}
+
+//ValidateBrickSubDirLength validates the length of each sub directories under
+//the brick path
+func ValidateBrickSubDirLength(brickPath string) int {
```
kshlm: Same as above

volgen/core.go
```diff
@@ -14,59 +15,84 @@ const (
...
+	//TODO : take the hostname from brickinfo
+	hname := brickinfo.Hostname
 	// Create volume directory (/var/lib/glusterd/vols/<VOLNAME>)
 	err := os.MkdirAll(vdir, 0666)
 	if err != nil {
-		panic(err)
+		return err
+	}
+
+	slash_to_hyphen := func(r rune) rune {
```
kshlm: Use mixed caps / camel case for names in Go. Don't use underscores.

volume/struct.go
```diff
@@ -112,12 +119,142 @@ func NewVolumeEntry(req VolCreateRequest) *Volinfo {
...
+	//TODO : Generate internal username & password
 
-	v.Bricks = req.Bricks
-
+	v.Bricks = NewBrickEntries(req.Bricks, req.Force)
+	if v.Bricks == nil {
+		return nil
+	}
 	return v
 }
```
kshlm: You could move all the store functions into a new file in the same package. This keeps code nice and segmented.

volume/struct.go
```diff
@@ -112,12 +119,142 @@ func NewVolumeEntry(req VolCreateRequest) *Volinfo {
...
+// volinfo object
+func GetVolume(name string) (*Volinfo, error) {
+	var v Volinfo
+	b, e := context.Store.GetVolumeFromStore(name)
+	if e != nil {
+		log.WithField("error", e).Error("Couldn't retrive volume from store")
+		return nil, e
+	}
+	if e = json.Unmarshal(b, &v); e != nil {
+		log.Error("Failed to unmarshal the data into volinfo object")
```
kshlm: Please log `e` as wel.

volume/struct.go
```diff
@@ -112,12 +119,142 @@ func NewVolumeEntry(req VolCreateRequest) *Volinfo {
...
+		}
+		volumes[index] = *vol
+	}
+	return volumes, nil
+
+}
+
+// NewBrickEntries returns list of initialized Brickinfo objects using list of
+// bricks
+func NewBrickEntries(bricks []string, force bool) []Brickinfo {
```
kshlm: This function is not used outside of this package. Keep it unexported.

volume/struct.go
```diff
@@ -112,12 +119,142 @@ func NewVolumeEntry(req VolCreateRequest) *Volinfo {
...
+		//TODO : Add validation to check whether file system support
+		//extended attributes
+		b = append(b, b1)
+	}
+	return b
+}
+
+// IsBrickPathAvailable validates whether the brick is consumed by other
+// volume
+func IsBrickPathAvailable(hostname string, brickPath string) int {
```
kshlm: Same here.

utils/utils.go
```diff
@@ -0,0 +1,189 @@
...
+	subdirs := strings.Split(brickPath, string(os.PathSeparator))
+	// Iterate over the sub directories and validate that they don't breach
+	//  _POSIX_PATH_MAX validation
+	for _, subdir := range subdirs {
+		if len(subdir) >= C._POSIX_PATH_MAX {
```
kshlm: Use `syscall.PathMax` instead of importing limits.h

kshlm:
I've done just a quick review, and most of the comments are about style.

I'd like to know why the store utility functions for volumes were moved out of the store package.

commands/volume-create/volume-create.go
atinmu: Done

commands/volume-create/volume-create.go
atinmu: Done

commands/volume-create/volume-create.go
atinmu: Done

commands/volume-create/volume-create.go
atinmu: Done

commands/volume-create/volume-create.go
atinmu: Done

commands/volume-create/volume-create.go
atinmu: Done

commands/volume-create/volume-create.go
atinmu: done

commands/volume-delete/volume-delete.go
atinmu: done

store/volume.go
atinmu: IMO, store/volume.go provides the helper function in respect to centralized store where as volume/struct.go has higher abstraction where a volume object is converted into bytes and then subsequent store function is called. This abstraction was also brought in to solve cyclic dependency. as store package was importing volume which is not correct as store should have lower level utility functions.

utils/utils.go
atinmu: done

utils/utils.go
atinmu: No, this is what I already discussed with you. Even if OS does validate about the PATH_MAX but it will create the subsequent sub directories while attempting a mkdir which we don't want.

volgen/core.go
atinmu: Done

volume/struct.go
atinmu: Done

volume/struct.go
atinmu: done

volume/struct.go
atinmu: done

volume/struct.go
atinmu: done

utils/utils.go
atinmu: syscall.PathMax is 4096, here the requirement is POSIX_PATH_MAX which is 256 chars. I haven't come across any constant which represents POSIX'es PATH_MAX.

volgen/core.go
kshlm: The `filepath` package has a function that does this already, [`filepath.FromSlash()`](https://golang.org/src/path/filepath/path.go?s=4601:4635#L161) . Use this instead.

utils/utils.go
```diff
@@ -0,0 +1,190 @@
...
+	"path/filepath"
+	"strings"
+	"syscall"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/gluster/glusterd2/errors"
+)
+
+// IsLocalAddress checks whether a given host/IP is local
+func IsLocalAddress(host string) bool {
```
kshlm: This does a much better job of identifying local addrs.

```go
func IsLocalAddr(host string) (bool, error) {
        laddrs, e := net.InterfaceAddrs()
        if e != nil {
                return false, e
        }
        var lips []net.IP
        for _, laddr := range laddrs {
                lipa := laddr.(*net.IPNet)
                lips = append(lips, lipa.IP)
        }

        rips, e := net.LookupIP(host)
        if e != nil {
                return false, e
        }
        for _, rip := range rips {
                for _, lip := range lips {
                        if lip.Equal(rip) {
                                return true, nil
                        }
                }
        }
        return false, nil
}
```

volume/volume_test.go
```diff
@@ -14,31 +29,67 @@ func TestNewVolumeEntry(t *testing.T) {
...
+		if hay == needle {
+			return true
+		}
+	}
+
+	return false
 }
 
-func TestNewVolumeEntryFromRequest(t *testing.T) {
+func TestNewVolumeEntryFromRequestBricksRootPartition(t *testing.T) {
```
kshlm: This is testing newBrickEntries no NewVolumeEntry. The name for this test is incorrect.

volume/volume_test.go
```diff
@@ -14,31 +29,67 @@ func TestNewVolumeEntry(t *testing.T) {
...
+	context.Init()
```
kshlm: Where would you need the use of the context package?

volume/volume_test.go
```diff
@@ -14,31 +29,67 @@ func TestNewVolumeEntry(t *testing.T) {
...
+	bricks := getSampleBricks("/b1", "/b2")
 
-	req := VolCreateRequest{}
-	req.Name = "vol1"
+	b := newBrickEntries(bricks, uuid.NewUUID().String(), true)
+	tests.Assert(t, b != nil)
+
+}
+
+func TestNewVolumeEntryFromRequestBricks(t *testing.T) {
```
kshlm: The same comment as above on the naming.

kshlm:
@atinmu please add a comment describing the test for each unit test.

kshlm:
@atinmu Your unit tests depend on the environment to do their tests. They currently depend on things like directory locations in the test host and host capabilities for xattrs.

A unit test shouldn't depend on the external environment. This will cause the unit test results to differ in different environment.

If a unit test requires a particular environment, then it should set it up for itself. If the the unit being tested is calling functions whose result depends on the external environment, then these functions should be mocked, so that they provide consistent results.

I don't know how we can mock out functions yet, especially the stdlib ones. I found this [mock][1] package and I'm trying to understand how it works and see if we can use it.

[1]: https://github.com/golang/mock

utils/utils.go
```diff
@@ -0,0 +1,261 @@
...
+	log "github.com/Sirupsen/logrus"
+	"github.com/gluster/glusterd2/errors"
+)
+
+const testXattr = "trusted.glusterfs.test"
+const volumeIDXattr = "trusted.glusterfs.volume-id"
+const gfidXattr = "trusted.gfid"
+
+// PosixPathMax represents POSIX_PATH_MAX
+var PosixPathMax int
```
kshlm: You could just do `var PosixPathMax int = C._POSIX_PATH_MAX` instead of having an `init` function. Better still make it a `const`.

utils/utils.go
kshlm: This is know behaviour, even the `mkdir` command behaves the same way. But I'll drop discussion on this for now.

store/volume.go
kshlm: This extra level of abstraction doesn't give any gains. The functions exported here are only used by the `volume` package. I don't believe anyone else would want to use these helper functions. It's better to just not have these at all.

volume/volume_test.go
```diff
@@ -1,11 +1,46 @@
+// +build !volumecreate
```
kshlm: Using a negative tag here makes the test command confusing. You run `go test -tags volumecreate`, which would cause one to assume that volume create tests would be run, but instead volume create tests aren't run with this.

Why not jut use `// +build volumecreate`? This way a plaing `go test` wouldn't run this, but `go test -tags volumecreate` would run this tests.

volume/volume_test.go
```diff
@@ -1,11 +1,46 @@
...
+	"os"
 	"testing"
 
+	"github.com/gluster/glusterd2/context"
 	"github.com/gluster/glusterd2/tests"
+	"github.com/pborman/uuid"
 )
 
+func init() {
+	context.Init()
```
kshlm: Where exactly is `context` being used here? I cannot find it, how much ever I search.

utils/utils.go
atinmu: Done

volume/volume_test.go
atinmu: Done

volume/volume_test.go
atinmu: The reason of initing the context is to initialize the store as volume create uses the store. Without initing the context the tests fail.

kshlm:
The changes requested in `store/volume.go` haven't been done. But I'm going ahead and merging this.

PR #26, opened 2015-10-20
Add ReST endpoints for Peer commands

### LINKED ISSUE #25: Add rest endpoints for peer commands
This also requires description of the ReST api for peer commands

kshlm:
Closes #25

[Review on Reviewable](https://reviewable.io/reviews/gluster/glusterd2/26)


atinmu:
<sup>**[commands/command.go, line 33 \[r1\]](https://reviewable.io:443/reviews/gluster/glusterd2/26#-K1bm8b5_OSUPjZyCtrp)** ([raw file](https://github.com/gluster/glusterd2/blob/5ad37fe52b72e9670675f41d6e6819d5771916a0/commands/command.go#L33)):</sup>
Earlier we had separate packages for each of the volume commands. This approach looks better though. We need to do the similar changes for existing volume commands as well.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/gluster/glusterd2/26)

atinmu:
<sup>**[commands/peers/addpeer.go, line 19 \[r1\]](https://reviewable.io:443/reviews/gluster/glusterd2/26#-K1bmVBezHkvXioRM_pa)** ([raw file](https://github.com/gluster/glusterd2/blob/5ad37fe52b72e9670675f41d6e6819d5771916a0/commands/peers/addpeer.go#L19)):</sup>
rename it to addPeerHandler to maintain uniformity


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/gluster/glusterd2/26)

atinmu:
<sup>**[commands/peers/addpeer.go, line 28 \[r1\]](https://reviewable.io:443/reviews/gluster/glusterd2/26#-K1bPaghfHffMeP8mzyz)** ([raw file](https://github.com/gluster/glusterd2/blob/5ad37fe52b72e9670675f41d6e6819d5771916a0/commands/peers/addpeer.go#L28)):</sup>
Can you have this error "no hostnames present" defined in errors.go?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/gluster/glusterd2/26)

atinmu:
<sup>**[store/store.go, line 57 \[r1\]](https://reviewable.io:443/reviews/gluster/glusterd2/26#-K1bSQg0_wbr8A0IspL-)** ([raw file](https://github.com/gluster/glusterd2/blob/5ad37fe52b72e9670675f41d6e6819d5771916a0/store/store.go#L57)):</sup>
On what case a GET will have empty prefix?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/gluster/glusterd2/26)

atinmu:
<sup>**[context/statedir.go, line 17 \[r1\]](https://reviewable.io:443/reviews/gluster/glusterd2/26#-K1boF3AmbJKrpk_IwcZ)** ([raw file](https://github.com/gluster/glusterd2/blob/5ad37fe52b72e9670675f41d6e6819d5771916a0/context/statedir.go#L17)):</sup>
As discussed earlier, do we really need to maintain this. IIRC, you need to maintain the local state dir as you claimed that maintaining UUID in central store might not be possible, but IMO this is not true. Please validate.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/gluster/glusterd2/26)

atinmu:
<sup>**[store/peer.go, line 9 \[r1\]](https://reviewable.io:443/reviews/gluster/glusterd2/26#-K1boeyopjKWwQ7CBqlV)** ([raw file](https://github.com/gluster/glusterd2/blob/5ad37fe52b72e9670675f41d6e6819d5771916a0/store/peer.go#L9)):</sup>
Store should not import any higher level packages like peer, volume etc.  The goal is to maintain the store package independent of the internals of the glusterd objects. It can work on the raw bytes of data where the respective higher level packages can marshal/unmarshall the data. Refer to latest volume-create patch for more details.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/gluster/glusterd2/26)

krisis:
Reviewed 14 of 14 files at r1.
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks broke.

---

<sup>**[commands/peers/addpeer.go, line 36 \[r1\]](https://reviewable.io:443/reviews/gluster/glusterd2/26#-K1biLP8AMTslT193qDX)** ([raw file](https://github.com/gluster/glusterd2/blob/5ad37fe52b72e9670675f41d6e6819d5771916a0/commands/peers/addpeer.go#L36)):</sup>
Could we have issues opened for TODOs and FIXMEs? 1) We can make them (more) actionable. 2) We can describe in detail what is remaining to be done or fixed as opposed to limiting ourselves to short summaries in 
comments. If you respond with what part of the validation you intend to defer to a later commit, I could open issues for the same. 


---

<sup>**[context/myuuid.go, line 44 \[r1\]](https://reviewable.io:443/reviews/gluster/glusterd2/26#-K1Xn06VQ7M6UOv-t0BV)** ([raw file](https://github.com/gluster/glusterd2/blob/5ad37fe52b72e9670675f41d6e6819d5771916a0/context/myuuid.go#L44)):</sup>
s/new MyUUID/new UUID


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/gluster/glusterd2/26)

kshlm:
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks broke.

---

<sup>**[commands/command.go, line 33 \[r1\]](https://reviewable.io:443/reviews/gluster/glusterd2/26#-K1bm8b5_OSUPjZyCtrp)** ([raw file](https://github.com/gluster/glusterd2/blob/5ad37fe52b72e9670675f41d6e6819d5771916a0/commands/command.go#L33)):</sup>
Let's keep this refactoring as a later task.


---

<sup>**[commands/peers/addpeer.go, line 19 \[r1\]](https://reviewable.io:443/reviews/gluster/glusterd2/26#-K1bmVBezHkvXioRM_pa)** ([raw file](https://github.com/gluster/glusterd2/blob/5ad37fe52b72e9670675f41d6e6819d5771916a0/commands/peers/addpeer.go#L19)):</sup>
Done.


---

<sup>**[commands/peers/addpeer.go, line 28 \[r1\]](https://reviewable.io:443/reviews/gluster/glusterd2/26#-K1bPaghfHffMeP8mzyz)** ([raw file](https://github.com/gluster/glusterd2/blob/5ad37fe52b72e9670675f41d6e6819d5771916a0/commands/peers/addpeer.go#L28)):</sup>
Done.


---

<sup>**[commands/peers/addpeer.go, line 36 \[r1\]](https://reviewable.io:443/reviews/gluster/glusterd2/26#-K1biLP8AMTslT193qDX)** ([raw file](https://github.com/gluster/glusterd2/blob/5ad37fe52b72e9670675f41d6e6819d5771916a0/commands/peers/addpeer.go#L36)):</sup>
The validations I referred to were things like checking if a peer with given addresses or name exists already.


---

<sup>**[context/myuuid.go, line 44 \[r1\]](https://reviewable.io:443/reviews/gluster/glusterd2/26#-K1Xn06VQ7M6UOv-t0BV)** ([raw file](https://github.com/gluster/glusterd2/blob/5ad37fe52b72e9670675f41d6e6819d5771916a0/context/myuuid.go#L44)):</sup>
We need a better name this. MyUUID is okay to use in code, but doesn't make too much sense in logs. And having it as just 'UUID' is not specific enough.


---

<sup>**[context/statedir.go, line 17 \[r1\]](https://reviewable.io:443/reviews/gluster/glusterd2/26#-K1boF3AmbJKrpk_IwcZ)** ([raw file](https://github.com/gluster/glusterd2/blob/5ad37fe52b72e9670675f41d6e6819d5771916a0/context/statedir.go#L17)):</sup>
Yes. We do need to maintain some local information. Especially the information which identifies each peer uniquely. This sort of information cannot be stored in the central store, we wouldn't be able to identify which information we stored.


---

<sup>**[store/peer.go, line 9 \[r1\]](https://reviewable.io:443/reviews/gluster/glusterd2/26#-K1boeyopjKWwQ7CBqlV)** ([raw file](https://github.com/gluster/glusterd2/blob/5ad37fe52b72e9670675f41d6e6819d5771916a0/store/peer.go#L9)):</sup>
This is not something I agree with a 100%. Why does store package need to be independent of the internals of GlusterD? This is a specialized package internal to GlusterD, which is not supposed to be used by any one else external. So IMO it is okay for the store package to have knowledge of the internals. The actual store itself is agnostic to the data being stored.


---

<sup>**[store/store.go, line 57 \[r1\]](https://reviewable.io:443/reviews/gluster/glusterd2/26#-K1bSQg0_wbr8A0IspL-)** ([raw file](https://github.com/gluster/glusterd2/blob/5ad37fe52b72e9670675f41d6e6819d5771916a0/store/store.go#L57)):</sup>
When nothing has been stored. For example on a very fresh start the store is completely empty. Any GETs on the store will fail, this will in turn lead to failure of GET peers and GET volumes. GET peers and GET volumes in such a case should return empty lists, instead of failures.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/gluster/glusterd2/26)

commands/peers/addpeer.go
```diff
@@ -0,0 +1,53 @@
...
+	if req.Name == "" {
+		req.Name = req.Addresses[0]
+	}
+
+	//TODO: Do proper validation before initiating the add process
+
+	//FIXME: In the correct add process, the peer being probed would add it's details to the store once it's been validated. The code below is just a temporary stand-in to show how the API's would work
+
+	p := &peer.Peer{
+		ID:        uuid.NewRandom(),
```
atinmu: I asked this question earlier as well, what's the benefit of using NewRandom() over NewUUID(). We need to be uniform across the code. Currently we are using the later for generating volume's UUID.

commands/peers/commands.go
```diff
@@ -0,0 +1,40 @@
...
+type Command struct {
+}
+
+// Routes returns command routes. Required for the Command interface.
+func (c *Command) Routes() rest.Routes {
+	return rest.Routes{
+		rest.Route{
+			Name:        "GetPeer",
+			Method:      "GET",
+			Pattern:     "/peers/{peerid}",
```
atinmu: Shouldn't this be peer name instead of ID?

config/config.go
```diff
@@ -6,14 +6,28 @@ package config
...
+// Any configuration that require custom default values should be initalized manually.
 var (
 	LogLevel    = flag.String("loglevel", "debug", "Log messages upto this level")
 	RestAddress = flag.String("restaddress", ":24007", "Address to bind REST endpoint to")
+
+	LocalStateDir string
 )
 
+func initLocalStateDir() {
+	wd, _ := os.Getwd()
```
atinmu: Is it same to generate the UUID in the current working directory or we should have it in /var/lib/glusterd?

commands/peers/addpeer.go
kshlm: There are several different versions of UUID, v1 to v5. `NewUUID` generates v1 which is not truly random, it uses a systems MAC address to generate the UUID. `NewRandom` generates v4, which is truly random.

commands/peers/commands.go
kshlm: I'd like it to allow both name and uuid here. So don't think that 'peerid' just means the uuid.

config/config.go
kshlm: We'll change this later.

commands/peers/commands.go
atinmu: Currently we are using the ID as a key in the store to store/retrieve the peer details. We'd definitely need to support name as key as that's a 3.x semantics but is there any specific reason why you'd want to use both? I am not sure how we can do that in store though.

atinmu: Getting the peer entry from the store by a peer name is *possible* if we can retrieve all the entries using GetPeers and then matching the peer name from those records. However I'd still like to have the key as name and retrieving the peer based on IDs can follow the approach discussed here. Let me know your thoughts. in volume's store-utils key is name, not the id at this moment.

kshlm: How we implement the ID-to-Name translation (and vice-versa) is not important. But it is something we will be doing.

No one other than us is going to be accessing the information stored in the store directly. So we are free to use UUIDs as the key. Using UUIDs to be the key is nice as they are guaranteed not to clash.

kshlm: I will be important for some external consumers of our API that we provide access to peers/volumes using unique identifiers that remain unique throughout. We will be guaranteeing that names will be only unique at this point in time. A name can be removed and reused, or it could be changed.
For example, access via UUIDs will be beneficial for a system monitoring a volume. UUIDs will ensure that it is always monitoring the same volume, and not a new volume that could be recreated with the same name.

PR #28, opened 2015-11-06
Vendoring the dependent packages.

kshlm:
This makes it easier to do customized builds later.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ vendor-update:
 	@echo
 
 verify: check-reqs
-	@GO15VENDOREXPERIMENT=1 gometalinter --errors --deadline=1m $$(GO15VENDOREXPERIMENT=1 glide nv)
+	@GO15VENDOREXPERIMENT=1 gometalinter -D gotype --errors --deadline=1m $$(GO15VENDOREXPERIMENT=1 glide nv)
 
 test:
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -0,0 +1,36 @@
...
+vendor-update:
+	@echo Updating vendored packages
+	@GO15VENDOREXPERIMENT=1 glide -q up
+	@echo
+
+verify: check-reqs
+	@GO15VENDOREXPERIMENT=1 gometalinter -D gotype --errors --deadline=1m $$(GO15VENDOREXPERIMENT=1 glide nv)
+
+test:
+	@GO15VENDOREXPERIMENT=1 go test $$(GO15VENDOREXPERIMENT=1 glide nv)
```
atinmu: Wouldn't you need a '-v' option here for go test as otherwise it won't download the dependent packages?

kshlm: `-v` is the verbose flag. The `go get` command has a `-t` flag which causes it to get the packages needed for testing as well.

AFAIK, we are not using any external packages for tests now. If we were to, we should just add it to the glide configuration and glide would take care of getting the package.

atinmu: Ahh, I meant -t only. My bad!! IIRC, we were using heketi's tests package earlier which caused it to fail the tests with out the option. But with vendoring concept we can overcome that too.

Other changes look good to me.

kshlm:
Thanks for the review @atinmu

PR #38, opened 2015-11-23
Add API versioning support

### LINKED ISSUE #35: Refactor volume commands
Currently individual volume commands are written in separate packages which is not ideal. Follow peer commands code to refactor.

atinmu:
This patch adds the following:

1. Adds /version end point
2. Introduces API versioning support for all other ReST end points
3. Correct pattern names in volumes & peer commands

Closes #35


kshlm:
@atinmu you'd mentioned you took inspiration from another project for this, which was it?

context/context.go
```diff
@@ -15,25 +15,35 @@ import (
...
+	GlusterdVersion int
 )
 
 var (
 	initOnce sync.Once
 )
 
+func initOpVersion() {
+	//TODO : Need cluster awareness and then decide the op-version
+	GlusterdVersion = MaxOpVersion
```
kshlm: To be clear, GlusterD version is not the same as the op-version. Glusterd version is the version of the binary, op-version is the operating version for the features.

context/context.go
```diff
@@ -15,25 +15,35 @@ import (
...
 var (
-	MyUUID uuid.UUID
-	Rest   *rest.GDRest
-	TxnFw  *transaction.GDTxnFw
-	Store  *store.GDStore
+	MyUUID          uuid.UUID
+	Rest            *rest.GDRest
+	TxnFw           *transaction.GDTxnFw
+	Store           *store.GDStore
+	GlusterdVersion int
```
kshlm: This should be OpVersion. Not GlusterdVersion. The GlusterD binary version will be a constant and isn't a part of the context.

context/context.go
atinmu: Ahh I got confused from the ReST API wiki, even I was also wondering why would we need to show op-version through /version end point. I'll correct it.

context/context.go
atinmu: Will modify accordingly

atinmu:
@kshlm : Its skying

kshlm:
@atinmu Just one suggestion. Have better commit messages, instead of 'resolve comments'. Resolve comments only makes sense when looking at this pull-request as a whole. But if someone were to look at the single commit in the repo, they wouldn't understand it as much.

PR #40, opened 2015-11-24
Response structure change

atinmu:


utils/utils.go
```diff
@@ -253,3 +262,20 @@ func isBrickPathAlreadyInUse(brickPath string) bool {
 	}
 	return false
 }
+
+// SendHTTPResponse to send response back to the client
+func SendHTTPResponse(w http.ResponseWriter, statusCode int, rsp interface{}) {
```
kshlm: Stop dumping all utilities into the same file. Also, it's better to not use the global utilities package for functions which are used by or useful to only one other package. I suggest that you add these utilities into the `rest` package.

atinmu: Makes sense, will do

Issue #44, opened 2015-12-02
Get volume request should only display list of volume name and associated volume id of that volume


gaurav36:
Description of problem:
Currently, when you make a volume Get request then it will display list of whole volume info (volinfo) object. 

How reproducible:
Every time when you issue Get volume request.

Expected behaviour:
It should show list of volume name and associated volume id with it.

gaurav36:
https://github.com/gaurav36/glusterd2/commit/8b5c585614a75940506592c1f1db6198d3c344f6 pr got merged. closing this issue.

PR #45, opened 2015-12-02
Get volume request should show only list of volume name and respectiv…

gaurav36:
…e uuid

Currently, when you make a volume Get request then it will display list
of whole volume info (volinfo) object.

With this patch it will  show list of volume name and associated volume uuid
when user make get volume request.

Issue #44

### DIFF
--- a/volume/store-utils.go
+++ b/volume/store-utils.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/gluster/glusterd2/context"
 	"github.com/gluster/glusterd2/store"
+	"github.com/pborman/uuid"
 
 	log "github.com/Sirupsen/logrus"
 )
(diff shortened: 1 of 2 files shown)


volume/store-utils.go
```diff
@@ -56,13 +56,13 @@ func DeleteVolume(name string) error {
 
 //GetVolumes retrives the json objects from the store and converts them into
 //respective volinfo objects
-func GetVolumes() ([]Volinfo, error) {
+func GetVolumes() ([]Volume_list, error) {
```
atinmu: Almost all go projects follow lower camel cases. Could you change the variable name to something like volList?

atinmu:
Please ensure you compile the code with 'make install' before sending the PR. Looking at the CI failure its straight forward that the changes break compilation.

volume/struct.go
```diff
@@ -50,6 +50,13 @@ const (
 	DistDispStripe
 )
 
+// Volume list
+// Represent the list of volume during GetVol request
+type Volume_list struct {
```
kshlm: A simple map[string]string should suffice for this. No need to create a whole new type.

kshlm:
Please run `make verify` and `make build` before sending PRs.

volume/store-utils.go
gaurav36: solved in latest patch set.

volume/struct.go
```diff
@@ -50,6 +50,13 @@ const (
 	DistDispStripe
 )
 
+// Volume list
+// Represent the list of volume during GetVol request
+type VolumeList struct {
```
kshlm: Like I mentioned in my previous comment, there is no need for a separate type for this. A map is good enough.

atinmu:
@gaurav36 github model works with every changes as separate commits. I think you have followed a gerrit workflow where you have rebased your changes on the same commit which is unacceptable in this model. Now onwards ensure to send separate commits.

PR #49, opened 2015-12-07
The Transaction framework

kshlm:
Doesn't actually work, but should be helpful to understand how it would
work.

Still need to figure out how to do rollback for SimpleTxn.


transaction/transaction.go
```diff
@@ -5,3 +5,46 @@ package transaction
...
+	//Do the steps
+	for i, s := range t.Steps {
+		if s.Nodes[0] == All {
+			s.Nodes = t.Nodes
+		} else if s.Nodes[0] == Leader {
+			//s.Nodes[0] = LeaderName
+		}
+
+		if e := s.do(); e != nil {
+			t.undo(i)
```
atinmu: Post undo, I think you'd need to break here and should not continue for further steps

atinmu:
We'd need to look at how to address the SPOF in case a originator node goes down while performing a transaction

transaction/step.go
```diff
@@ -0,0 +1,48 @@
...
+
+// StepRet is what the Step returns
+type StepRet interface{}
+
+// StepFunc is the function that is supposed to be run during a transaction step
+type StepFunc func(StepArg) StepRet
+
+const (
+	//Leader is a constant string representing the leader node
+	Leader = "leader"
```
atinmu: Should we term it as leader or originator?

kshlm: Leader should be okay. It's the one leading the transaction.

transaction/transaction.go
kshlm: Yup.

context/context.go
```diff
@@ -1,60 +1,69 @@
...
+	return &Context{
+		parent: c,
+		data:   make(map[string]interface{}),
+		Log:    c.Log,
+	}
+}
 
-	initLocalStateDir()
+// NewLoggingContext returns a new context with the logger set to log given fields in addition to the parents logging fields
+func (c *Context) NewLoggingContext(fields log.Fields) *Context {
```
atinmu: NewLoggingContext () is already defined at line 24. Same function name is used here as well. What's the difference?

context/context.go
```diff
@@ -1,60 +1,69 @@
-// Package context is the runtime context of GlusterD
-//
-// Any package that needs access to the GlusterD runtime context just needs to
-// import this package.
 package context
```
atinmu: I'd be interested to see how the consumers are going to use these APIs

kshlm:
Works well enough and should be ready for merge now.

commands/volumes/volume-create.go
```diff
@@ -43,68 +47,117 @@ func createVolume(msg *volume.VolCreateRequest) (*volume.Volinfo, error) {
...
+		Commit:   generateVolfiles,
+		Store:    storeVolume,
+		Rollback: rollBackVolumeCreate,
 	}
-	httpStatusCode, e = validateVolumeCreate(msg, vol)
-	if e != nil {
-		rest.SendHTTPError(w, httpStatusCode, e.Error())
+
+	if e := txn.Do(); e != nil {
+		rest.SendHTTPError(w, http.StatusInternalServerError, e.Error())
```
atinmu: So this indicates we are reporting back a generic http status code in case of an error. Earlier we were using different status codes. Is it something we can't maintain the same?

commands/volumes/volume-create.go
```diff
@@ -43,68 +47,117 @@ func createVolume(msg *volume.VolCreateRequest) (*volume.Volinfo, error) {
...
+	c.Set("volinfo", vol)
+
+	return nil
 }
 
-func commitVolumeCreate(vol *volume.Volinfo) (int, error) {
+func generateVolfiles(c *context.Context) error {
+	vol, ok := c.Get("volinfo").(*volume.Volinfo)
+	if !ok {
+		return errors.New("failed to get volinfo from context")
```
atinmu: Can we define this error in gderrors as I can see this been used in multiple places?

transaction/simple.go
```diff
@@ -0,0 +1,72 @@
...
+		Nodes:    []string{All},
+	}
+	commitstep := &Step{
+		DoFunc:   commit,
+		UndoFunc: rollback,
+		Nodes:    []string{All},
+	}
+	storestep := &Step{
+		DoFunc:   store,
+		UndoFunc: nil,
```
atinmu: Doing nothing in Undo here is a problem given that store is post commit, ideally we should use rollback, isn't it?

atinmu: Ignore this as I now realize that undo function is going to call respective undoFuncs for all the steps.

commands/volumes/volume-create.go
kshlm: I will, but in a later PR. Dumping errors of different packages into a single `error` package doesn't seem really good to me. I'll do a refactoring of it, and will add this in it.

commands/volumes/volume-create.go
kshlm: We could. We could have the steps set a error value in the context, which can be returned from here.

atinmu: @kshlm : Would you mind doing it in this PR?

PR #53, opened 2015-12-10
migrate to etcd client api

### LINKED ISSUE #17: Decide on store and remove libkv
Though libkv is helpful, it brings in a lot of dependencies which are not actually useful. Ref [1][1]

It'll also be helpful if we decide on a particular store as we can make better use of its complete API.

[1]: https://gist.github.com/kshlm/150de99c8271d3030c43

atinmu:
We initially started with using libkv as an interface to the central stores like
consul/etcd. But later on we realized it brings in lot of dependencies and since
we've already choosen to use etcd its better to use etcd client APIs.

Closes #17

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -1,47 +1,68 @@
 parent: null
 package: github.com/gluster/glusterd2
 import:
+- package: github.com/coreos/etcd
+- package: github.com/shurcooL/sanitized_anchor_name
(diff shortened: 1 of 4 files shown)


kshlm:
I'm closing this. #54 switches to etcd.

atinmu:
I want to keep this open till we get to know the numbers w.r.t to dependencies of pulling in libkv/etcd vs etcd-client

PR #54, opened 2015-12-14
Switch to etcd and fix vendoring

kshlm:
glide itself has issues, so now we pull in a huge amount of dependencies. I've
opened issues on glide to help/solve this problems.


Makefile
```diff
@@ -26,7 +26,7 @@ install:
 
 vendor-update:
 	@echo Updating vendored packages
-	@GO15VENDOREXPERIMENT=1 glide -q up
+	@GO15VENDOREXPERIMENT=1 glide install
```
atinmu: If the package is already installed, does option 'install' updates it? What's the reason of removing 'up' ?

atinmu:
So do we have the data in terms of dependencies for using libkv/etcd vs etcd-client?

kshlm:
> So do we have the data in terms of dependencies for using libkv/etcd vs etcd-client?

Glide has its own problems so, we don't have a comparison yet.

Makefile
kshlm: With glide-0.8.0, a new `glide.lock` file is created. `glide install` will install/update/downgrade packages listed in the lockfile.

PR #58, opened 2015-12-21
Enable volume unit test

### LINKED ISSUE #57: Enable unit tests for volume package by mocking out few syscalls

atinmu:
This patch reenables the volume package's unit test by mocking out few
dependencies like etcd, xattr calls.

Closes #57


tests/patch.go
```diff
@@ -0,0 +1,39 @@
+// From https://gist.github.com/imosquera/6716490#sthash.O4z2aQQp.LUHz2Cbb.dpuf
```
lpabon: I'm seeing a lot of copying (not just this project).  Instead, please use github.com/heketi/tests.  I do not mind moving to another repo in the future, but in the mean time, it is better to pull from the same area.

tests/test-utils.go
```diff
@@ -0,0 +1,21 @@
+package tests
```
lpabon: You may want to call this `tests/mock-attr.go`, instead of the generic name of `test-utils.go`

utils/utils.go
```diff
@@ -17,9 +17,18 @@ import (
...
-const volumeIDXattr = "trusted.glusterfs.volume-id"
-const gfidXattr = "trusted.gfid"
+const (
+	testXattr     = "trusted.glusterfs.test"
+	volumeIDXattr = "trusted.glusterfs.volume-id"
+	gfidXattr     = "trusted.gfid"
+)
+
+var (
+	PathMax     = unix.PathMax
```
lpabon: Do you need to expose these functions to other packages?  If not, you probably want them using lower case letters.

utils/utils_test.go
```diff
@@ -85,7 +87,8 @@ func TestValidateBrickPathStats(t *testing.T) {
 }
 
 func TestValidateXattrSupport(t *testing.T) {
-	//TODO : xattr related calls need root permission and hence having valid
-	//tests will always fail in build system. Need to find a way to tackle
-	//it.
+	defer tests.Patch(&Setxattr, tests.MockSetxattr).Restore()
+	defer tests.Patch(&Getxattr, tests.MockGetxattr).Restore()
+	defer tests.Patch(&Removexattr, tests.MockRemovexattr).Restore()
+	tests.Assert(t, ValidateXattrSupport("/tmp/b1", "localhost", uuid.NewRandom(), true) == nil)
```
lpabon: Most of the tests are positive tests.  We should also add some negative tests to make sure that the code reacts in the correct way.  For example:

```go
// Do not set, returns nil the first time
var setxattr_err error
defer tests.Patch(&Setxattr, func (path string, attr string, data []byte, flags int) (err error) {
	return setxattr_err
}

// Some tests here

// Now check what happens when setxattr fails
baderror := errors.New("Bad")
setxattr_err = baderror

// Test again
tests.Assert(t, ValidateXattrSupport(...) == baderror)

```

volume/struct.go
```diff
@@ -26,6 +26,10 @@ const (
 	VolStopped
 )
 
+var (
+	MValidateBrickPathStats = utils.ValidateBrickPathStats
```
lpabon: Does MValidateBrickPathsStats need to be exposed?  Also, just curious, but why start with MV?

volume/volume-utils.go
```diff
@@ -9,6 +9,10 @@ import (
 	log "github.com/Sirupsen/logrus"
 )
 
+var (
+	MGetVolumes = GetVolumes
```
lpabon: Same here.

volume/volume_test.go
```diff
@@ -85,16 +76,31 @@ func TestNewBrickEntryFromRequestBricks(t *testing.T) {
 // TestNewVolumeEntryFromRequest tests whether the volume is created with a
 // valid request
 func TestNewVolumeEntryFromRequest(t *testing.T) {
```
lpabon: This is good.  Now you should also add negative tests.

tests/patch.go
atinmu: @lpabon : Actually for vendoring we are using glide and it has a known flaw of pulling the complete parent dependencies. For that we chose to reduce the dependency as much as possible. If glide guys fix it in sometime which they promised we'll definitely start using these utilities from the origin instead of forking. Does it make sense?

tests/test-utils.go
atinmu: Makes sense. Will do

utils/utils.go
atinmu: They are exposed in volume package, volume_test.go uses it.

utils/utils_test.go
atinmu: Thanks for this pseudo code, will cover the negative tests as well.

volume/struct.go
atinmu: Not really, will change it along with a better name :)

volume/volume-utils.go
atinmu: Will do

volume/volume_test.go
atinmu: Sure, will revisit the section and add negative tests.

atinmu:
Thanks @lpabon for taking your time to review the changes.

tests/patch.go
lpabon: I use [glock](https://github.com/robfig/glock) or you can use [gdeps](https://github.com/tools/godep)

utils/utils.go
lpabon: If they are in the same package, all variables are accessible.  Think of the same package as being in the same file.  Only export if you want to make the variables accessible from other packages.

atinmu: This is a utils package and we need to export it to volumes, so these should be exported.

tests/patch.go
kshlm: heketi/tests is a separate package from heketi/heketi and it doesn't import anything outside the Go stdlib. It is okay to import it here, instead of copying.

Makefile
```diff
@@ -26,7 +26,7 @@ install:
 
 vendor-update:
 	@echo Updating vendored packages
-	@GO15VENDOREXPERIMENT=1 glide install
+	@GO15VENDOREXPERIMENT=1 glide up
```
kshlm: This has been a bug with glide-0.8.2. Refer https://github.com/Masterminds/glide/issues/185 . This will be fixed in the next release.
`glide up` is a costly operation, which needs to be done only when adding new dependencies or updating dependencies to newer versions.

PR #63, opened 2015-12-30
Cross GlusterD interaction using pbrpc

atinmu:


rpc/server.go
```diff
@@ -0,0 +1,53 @@
...
+	"net/rpc"
+
+	"github.com/gluster/glusterd2/context"
+	"github.com/gluster/glusterd2/peer"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/kshlm/pbrpc/pbcodec"
+)
+
+func (r *Connection) PeerAddRPCSvc(args *peer.PeerAddRequest, reply *RPCPeerAddResp) error {
```
kshlm: The error is here. You should be the protobuf generated type RPCPeerAddReq. You changed it in all other places, except this.

Because of this, ReadRequestBody in pbcodec isn't able to unmarshal the request body, and returns an error. The `net/rpc` framework silently ignores this and doesn't log or pass up any errors.

kshlm:
This is not the structure I expected for RPC. The GD2 RPC package will be a general RPC package, which will be mostly a wrapper around `net/rpc` using `pbcodec` and the GD2 connection information. It would also provide methods for registering services. This package would be analogous to the `/rpc/` directory in GlusterFS. The actual services would be implemented else-where, again as in GlusterFS.

I would suggest the following directory/package structuring.

```
rpc/
    server/ -> would contain the code necessary to create and manage a listener and methods to register services
    client/ -> would contain code necessary to establish and manage client connections and methods to send requests
    services/ -> would contain the actual services. The services and the handler functions would be defined here
```

rpc/server.go
atinmu: Ahh, silly! I corrected it. Now that its hitting the server handler but still there is a deadlock, client is not get back the response :(

rpc/server.go
```diff
@@ -0,0 +1,53 @@
...
+	log.Debug("In PeerAdd")
+	if context.MaxOpVersion < 40000 {
+		*reply.OpRet = -1
```
kshlm: `reply.OpRet` and `reply.OpError` are nil pointers, pointing to nothing. There is actually no location for the data you are setting to be stored. This is a side effect of using protobuf. Protobuf generated structs use only pointers as members. `net/rpc` package will allocate the struct and initialize the pointers to nil. The pointers will not be pointing to any allocated memory. You have to set these pointers to data you've allocated.

```go
var (
  opRet int32
  opError string
)

// Update opRet and opError instead of reply.OpRet and reply.OpError

// before returning
reply.OpRet = &opRet
reply.OpError = &opError

return nil
}

atinmu: Yes, that did the trick! Many thanks @kshlm

kshlm:
Looks good now.

You have vim swap file committed, please remove it.

Also, it'd be nice if the listener port were configurable. At least add a TODO for this.

Issue #64, opened 2016-01-05
Start etcd daemon upon starting of glusterd


gaurav36:
Currently upon starting of glusterd daemon, etcd is not starting. For etcd bootstrapping first requirement is to run etcd daemon on every node.

gaurav36:
PR https://github.com/gluster/glusterd2/pull/65 got merged. closing this issue.

PR #65, opened 2016-01-05
Start etcd daemon upon starting of glusterd daemon

gaurav36:
Currently with starting of glusterd daemon etcd is not starting.
For bootstrapping of etcd first basic requirement is to run etcd
daemon on every node.

With this patch it will start etcd daemon upon starting of
glusterd daemon.

Issue #64


main.go
```diff
@@ -41,4 +42,12 @@ func main() {
 	if err != nil {
 		log.Fatal("Could not start GlusterD Rest Server. Aborting.")
 	}
+
+	// Starting etcd daemon upon starting of GlusterD
+	etcd_start := exec.Command("/bin/etcd")
```
atinmu: Have you tested these changes? I don't think etcd instance will come up here. OTOH, we'd need to pass some additional parameters to start etcd. Refer to etcd documentation.

gaurav36: I have tested this change only in sample programme. this is not my final change. will add some parameter in new PR.

context/etcd-mgmt.go
```diff
@@ -0,0 +1,36 @@
+package context
```
atinmu: Why this needs to be in `context` package? You could have a separate package called etcd-mgmt and then have bootstrap.go.

context/etcd-mgmt.go
```diff
@@ -0,0 +1,36 @@
...
+
+func StartEtcd() {
+
+	log.Info("Starting etcd")
+	hostname, err := os.Hostname()
+	if err != nil {
+		log.Fatal("Could not able to get hostname")
+	}
+
+	ip_2379 := strings.Join([]string{"http://", ":2379"}, hostname)
```
atinmu: ip_2379 - golang doesn't entertain this style, use cameling style instead

context/etcd-mgmt.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"os"
+	"os/exec"
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+func StartEtcd() {
```
atinmu: Shouldn't you need to handle error from this function?

context/etcd-mgmt.go
```diff
@@ -0,0 +1,36 @@
...
+	hostname, err := os.Hostname()
+	if err != nil {
+		log.Fatal("Could not able to get hostname")
+	}
+
+	ip_2379 := strings.Join([]string{"http://", ":2379"}, hostname)
+
+	ip_2380 := strings.Join([]string{"http://", ":2380"}, hostname)
+
+	default_etcd := strings.Join([]string{"default=", ""}, ip_2380)
```
atinmu: Why do you need to construct default_etcd like this? You can still pass the default parameters like what you have done for other arguments.

atinmu:
We'd also need to stop the etcd daemon on glusterd shutdown.

atinmu:
I'd also like to see unit test coverage for this.

etcdmgmt/etcd-mgmt.go
```diff
@@ -0,0 +1,60 @@
...
+func StartEtcd() error {
+
+	log.Info("Starting etcd")
+	hostname, err := os.Hostname()
+	if err != nil {
+		log.Fatal("Could not able to get hostname")
+		return err
+	}
+
+	HostPort2379 := "http://" + hostname + ":2379"
```
atinmu: We should not start names with caps unnecessarily until and unless they are exposed to other packages. Also I'd like to see a better name here. There is no harm of assigning same value to multiple variables. listenClientUrils, advClientUrls would be better for the sake of better code readability

etcdmgmt/etcd-mgmt.go
```diff
@@ -0,0 +1,60 @@
...
+		"-initial-advertise-peer-urls", HostPort2380,
+		"--initial-cluster", "default="+HostPort2380)
+
+	err = EtcdStart.Start()
+	if err != nil {
+		log.Fatal("Could not start etcd daemon.")
+		return err
+	}
+
+	//Checking health of etcd cluster
```
atinmu: Can we really wait for the cluster to be healthy for indefinite time? I think we should error out after a good amount of retries.

atinmu: Also have a more explanatory comment why are we checking the health here.

etcdmgmt/etcd-mgmt.go
```diff
@@ -0,0 +1,60 @@
...
+	for {
+		result := struct{ Health string }{}
+
+		resp, err := http.Get(HostPort2379 + "/health")
+		if err != nil {
+			continue
+		}
+
+		Body, err := ioutil.ReadAll(resp.Body)
```
atinmu: s/Body/body

main.go
```diff
@@ -13,6 +14,12 @@ import (
 func main() {
 	log.Info("GlusterD starting")
 
+	// Starting etcd daemon upon starting of GlusterD
+	err := etcdmgmt.StartEtcd()
+	if err != nil {
+		log.Fatal("Could not able to start etcd")
```
atinmu: log the error as well with log.WithField

atinmu:
You'd also need to merge the master and repush the changes as this branch indicates a conflict with the master

atinmu:
@gaurav36 : Also add test for etcdmgmt package

etcdmgmt/etcd-mgmt.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		log.Fatal("Could not able to get hostname")
+		return err
+	}
+
+	listenClientUrls := "http://" + hostname + ":2379"
+
+	advClientUrls := "http://" + hostname + ":2380"
+
+	etcdStart := exec.Command("/bin/etcd",
```
atinmu: s/etcdStart/etcdCommand

etcdmgmt/etcd-mgmt.go
```diff
@@ -0,0 +1,73 @@
...
+		"-listen-client-urls", listenClientUrls,
+		"-advertise-client-urls", listenClientUrls,
+		"-listen-peer-urls", advClientUrls,
+		"-initial-advertise-peer-urls", advClientUrls,
+		"--initial-cluster", "default="+advClientUrls)
+
```
atinmu: I had asked you to use different variable names for all these parameters, it doesn't harm us if the values are duplicate since the readability will improve.

etcdmgmt/etcd-mgmt.go
```diff
@@ -0,0 +1,73 @@
...
+	err = etcdStart.Start()
+	if err != nil {
+		log.Fatal("Could not start etcd daemon.")
```
atinmu: Again the comment was ignored here. We'd need to log the reason of why etcd daemon couldn't be started. log.WithField("error":err.Error()).Fatal(.....)

etcdmgmt/etcd-mgmt.go
```diff
@@ -0,0 +1,73 @@
...
+	// means etcd have initialized properly before using any etcd command
+	for {
+
+		// Waiting for 15 second. Within 15 second health of etcd should
+		// be true otherwise it should through an error
+		timer := time.NewTimer(time.Second * 15)
+		go func() {
+			<-timer.C
+			if result.Health != "true" {
+				log.Fatal("Health of etcd is not proper. Please check your etcd configuration.")
```
atinmu: Remove 'Please'
s/check your/check

etcdmgmt/etcd-mgmt.go
```diff
@@ -0,0 +1,73 @@
...
+		return err
+	}
+
+	result := struct{ Health string }{}
+	// Checking health of etcd. Health of the etcd should be true,
+	// means etcd have initialized properly before using any etcd command
+	for {
+
+		// Waiting for 15 second. Within 15 second health of etcd should
+		// be true otherwise it should through an error
```
atinmu: s/through/throw

etcdmgmt/etcd-mgmt.go
gaurav36: we are starting etcd it self. If we are executing any etcd command then we should have etcdCommand here.

atinmu:
@gaurav36 : rest of the things look good to me.
@kshlm : Do you want to have a look at it?

etcdmgmt/etcd-mgmt.go
gaurav36: for shake of avoiding multiple variable for having same value i used generic name. i dont have problem using separate variable . will do in next ptach

etcdmgmt/etcd-mgmt.go
atinmu: @gaurav36 : etcdCommand interprets what command string you have formed for the etcd instance to bring up. If you look at the signature of exec.Command it returns a command structure.  Check other projects if you wish, mostly they use 'cmd' for the return value.

etcdmgmt/etcd-mgmt.go
```diff
@@ -0,0 +1,77 @@
...
+
+	listenClientUrls := "http://" + hostname + ":2379"
+
+	advClientUrls := "http://" + hostname + ":2379"
+
+	listenPeerUrls := "http://" + hostname + ":2380"
+
+	initialAdvPeerUrls := "http://" + hostname + ":2380"
+
+	etcdStart := exec.Command("/bin/etcd",
```
atinmu: This would fail in systems where etcd is not installed. This has to be taken care in install-req.sh

gaurav36: I have created separate issue https://github.com/gluster/glusterd2/issues/70 for this. PR https://github.com/gluster/glusterd2/pull/71 will address this comment.

atinmu: Please do not use '/bin/etcd'. use etcd instead as the the executable should be picked up from $GOPATH

gaurav36: Done

PR #67, opened 2016-01-13
Set GOPATH variable if empty, else gometalinter installation fails.

sac:


kshlm:
@sac, we mention that GOPATH needs to be setup in the [workspace setup](https://github.com/gluster/glusterd2/wiki/Development-Guide#workspace-setup) section of the development guide.

The suggested way to clone GD2 right now, is to run `go get -d`. This requires GOPATH to be set up. So, ideally, one would not end up trying to install gometalinter, which is a development time tool, without having their GOPATH set up.

We could make this information more visible by having it directly in the README.

Do you any particular reason, which would make the above requirements invalid and would require automatic setting up of GOPATH?

sac:
@kshlm no particular reason, actually instead of `go get' (which apparently is the way to go) I forked and cloned the repository. And hit this road block, and thought `/usr' would be an ideal default. I will remove the check.

And also, is there any reason we don't do `root' user check?

kshlm:
We shouldn't need to be root to build the project. We would ideally need to run as root to operate as glusterd, but that will be a check within GD2.

I'm closing this issue, as we don't need it.

Issue #68, opened 2016-01-14
Could not able to create volume if brick is having in hierarchical directory stracture


gaurav36:
Currently volume creation fails if while volume creation brick are falling in hierarchical directory and that hierarchical directory does not exist.

gaurav36:
Patch got merged. Closing this issue.

PR #69, opened 2016-01-14
Supporting hierarchical directory structure for bricks

gaurav36:
While creating volume using glusterd2 it should support creation of
hirearchical directory stracture, if brick present in hierarchical
directory.

Issue #68

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -4,13 +4,14 @@ package utils
 import "C"
 
 import (
-	"golang.org/x/sys/unix"
 	"net"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
 	"syscall"
+
+	"golang.org/x/sys/unix"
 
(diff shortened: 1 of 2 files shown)


atinmu:
@gaurav36 : Please add a test in utils/utils_test.go covering this change.

utils/utils.go
```diff
@@ -12,6 +11,8 @@ import (
 	"strings"
 	"syscall"
 
+	"golang.org/x/sys/unix"
```
atinmu: Why do you need this package to be imported?

gaurav36: I have not include this package this package included automatically while saving file in vim-go. for more information about this package you can follow https://godoc.org/golang.org/x/sys/unix

gaurav36: or you can see the use case on    line number 28..31 in this file (utils/utils.go)

Issue #70, opened 2016-01-19
Install etcd as a package requirement for GlusterD 2.0


gaurav36:
Install etcd while executing install-reqs.sh (all necessary package for GlusterD 2.0).

gaurav36:
PR https://github.com/gaurav36/glusterd2/commit/0f938b570a299fdffd5f114b92fe6a4348ee685a got merged. Closing this issue.

PR #71, opened 2016-01-19
Install etcd as a package requirement for GlusterD 2.0

gaurav36:
Install etcd while executing install-reqs.sh (all necessary package for
GlusterD 2.0).

Issue #70

### DIFF
--- a/scripts/check-reqs.sh
+++ b/scripts/check-reqs.sh
@@ -2,7 +2,7 @@
 
 # Checks if the tools required for properly building, verifying and testing GlusterD are installed.
 
-TOOLS=(glide gometalinter)
+TOOLS=(glide gometalinter etcd)
 
 MISSING=0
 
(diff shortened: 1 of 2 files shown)


scripts/install-reqs.sh
```diff
@@ -44,5 +44,29 @@ install_gometalinter() {
   gometalinter --install --update || failed_install linters
 }
 
+install_etcd() {
+        ETCDVERSION="v2.2.0"
```
atinmu: Why aren't we using the latest package 2.2.4?

scripts/check-reqs.sh
```diff
@@ -2,17 +2,17 @@
...
-TOOLS=(glide gometalinter)
+TOOLS=(glide gometalinter etcd)
 
 MISSING=0
 
 for tool in ${TOOLS[@]}; do
   type $tool >/dev/null 2>&1
   if [ $? -ne 0 ]; then
-    echo "$tool is missing on the system"
+    echo "$tool package is missing on the system"
```
atinmu: it should be tool/package

gaurav36: it should not be tool package. because suppose etcd is missing then value of $tool will be etcd, and according to your opinion if i change it to too/package then it will become  etcd/package missing.

atinmu:
@gaurav36 : Apologies for misreading it.

Issue #72, opened 2016-01-21
etcd bootstrapping in GlusterD 2.0


gaurav36:
Currently Glusterd2 is not doing etcd bootstrapping automatically. It should do member add/delete and bootstrap cluster accordingly upon execution of peer probe/peer detach command.

kshlm:
This was done. But we will be redoing it using embedded etcd.

kshlm:
#147 is tracking this now.

PR #73, opened 2016-01-21
etcd bootstrapping in glusterd 2.0

gaurav36:
Adding member upon peer probe in the etcd cluster.

With this change it will add remote host url during peer probe of remote host
in the cluster.

Issue #72

### DIFF
--- a/commands/peers/addpeer.go
+++ b/commands/peers/addpeer.go
@@ -1,13 +1,19 @@
 package peercommands
 
 import (
+	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/gluster/glusterd2/errors"
 	"github.com/gluster/glusterd2/peer"
 	"github.com/gluster/glusterd2/rest"
 	"github.com/gluster/glusterd2/rpc/client"
 	"github.com/gluster/glusterd2/utils"
+
+	log "github.com/Sirupsen/logrus"
+	etcdclient "github.com/coreos/etcd/client"
+	etcdcontext "golang.org/x/net/context"
 
 	"github.com/pborman/uuid"
 )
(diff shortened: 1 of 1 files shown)


commands/peers/addpeer.go
```diff
@@ -43,6 +49,44 @@ func addPeerHandler(w http.ResponseWriter, r *http.Request) {
 		rest.SendHTTPError(w, http.StatusInternalServerError, *rsp.OpError)
 		return
 	}
+
+	// Add member to etcd server
```
kshlm: We'll need to do a lot of etcd actions in multiple other places. So it will be good if you get the etcd client initialized in a central location, and use that everywhere, instead of creating a new client every time.

commands/peers/addpeer.go
```diff
@@ -43,6 +49,44 @@ func addPeerHandler(w http.ResponseWriter, r *http.Request) {
...
+	if e != nil {
+		//rest.SendHTTPError(w, http.StatusInternalServerError, strings.e)
+		rest.SendHTTPError(w, http.StatusInternalServerError, e.Error())
+		return
+	}
+
+	newID := member.ID
+	newName := member.Name
+
+	log.Info("New member named %s with ID %s added to cluster", newName, newID)
```
kshlm: Please follow structured logging. We will not have logs which have format strings.

commands/peers/addpeer.go
```diff
@@ -43,6 +49,44 @@ func addPeerHandler(w http.ResponseWriter, r *http.Request) {
...
+		for _, u := range memb.PeerURLs {
+			n := memb.Name
+			if memb.ID == newID {
+				n = newName
+			}
+			conf = append(conf, fmt.Sprintf("%s=%s", n, u))
+		}
+	}
+
+	log.Info("ETCD_NAME=%q", newName)
```
kshlm: Same here. You should be using structured logging. Look at other logs for examples.

commands/peers/addpeer.go
```diff
@@ -43,6 +49,44 @@ func addPeerHandler(w http.ResponseWriter, r *http.Request) {
...
+	log.Info("ETCD_INITIAL_CLUSTER=%q", strings.Join(conf, ","))
+	log.Info("ETCD_INITIAL_CLUSTER_STATE=\"existing\"")
+
 	if e = peer.AddOrUpdatePeer(p); e != nil {
```
kshlm: You should be sending the above details to the peer you are adding. The new peer will need these details to add it's etcd to the cluster. Just logging it here doesn't help.

gaurav36:
closing this PR as updated PR available https://github.com/gluster/glusterd2/pull/77

Issue #66, opened 2016-01-12
Daemonize etcd


atinmu:
etcd instance can't be brought down when glusterd goes down since the other consumers like NSR would still be able to access it. This task is to have etcd daemonized such that on glusterd shutdown etcd instance doesn't go down.

PR #74, opened 2016-01-21
Daemonize etcd in GlusterD 2.0

gaurav36:
Currently when etcd daemon start upon starting of glusterd2 then upon
performing ^C (SIGINT) its killing etcd daemon.

With this patch it will set child process id to different from main
process id group and, etcd daemon will not be kill upon performing
SIGINT (^C) of main process.

Issue #66

### DIFF
--- a/etcdmgmt/etcd-mgmt.go
+++ b/etcdmgmt/etcd-mgmt.go
@@ -2,6 +2,7 @@ package etcdmgmt
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"os"
@@ -12,49 +13,14 @@ import (
 	log "github.com/Sirupsen/logrus"
 )
 
-func StartEtcd() error {
-
-	log.Info("Starting etcd")
-	hostname, err := os.Hostname()
-	if err != nil {
-		log.Fatal("Could not able to get hostname")
-		return err
-	}
-
-	listenClientUrls := "http://" + hostname + ":2379"
-
-	advClientUrls := "http://" + hostname + ":2379"
-
-	listenPeerUrls := "http://" + hostname + ":2380"
-
-	initialAdvPeerUrls := "http://" + hostname + ":2380"
-
-	etcdStart := exec.Command("etcd",
-		"-listen-client-urls", listenClientUrls,
-		"-advertise-client-urls", advClientUrls,
-		"-listen-peer-urls", listenPeerUrls,
-		"-initial-advertise-peer-urls", initialAdvPeerUrls,
-		"--initial-cluster", "default="+listenPeerUrls)
-
-	// Don't kill chlid process (etcd) upon ^C (SIGINT) of main glusterd process
-	etcdStart.SysProcAttr = &syscall.SysProcAttr{
-		Setpgid: true,
-	}
-
-	err = etcdStart.Start()
-	if err != nil {
-		log.WithField("error", err.Error()).Fatal("Could not start etcd daemon.")
-		return err
-	}
-
+func checkHealth(val time.Duration, listenClientUrls string) bool {
 	result := struct{ Health string }{}
 	// Checking health of etcd. Health of the etcd should be true,
 	// means etcd have initialized properly before using any etcd command
+	timer := time.NewTimer(time.Second * val)
 	for {
-
 		// Waiting for 15 second. Within 15 second health of etcd should
 		// be true otherwise it should throw an error
-		timer := time.NewTimer(time.Second * 15)
 		go func() {
 			<-timer.C
 			if result.Health != "true" {
@@ -77,6 +43,61 @@ func StartEtcd() error {
 			timer.Stop()
 			break
 		}
+	}
+	return true
+}
+
+func StartEtcd() error {
+	result := struct{ Health string }{}
+	log.Info("Starting etcd")
+	hostname, err := os.Hostname()
+	if err != nil {
+		log.Fatal("Could not able to get hostname")
+		return err
+	}
+
+	listenClientUrls := "http://" + hostname + ":2379"
+
+	advClientUrls := "http://" + hostname + ":2379"
+
+	listenPeerUrls := "http://" + hostname + ":2380"
+
+	initialAdvPeerUrls := "http://" + hostname + ":2380"
+
+	// Checking health of etcd cluster. If etcd daemon already running then
+	// no need to start etcd daemon again.
+	resp, err := http.Get(listenClientUrls + "/health")
+	if err == nil {
+		body, err := ioutil.ReadAll(resp.Body)
+
+		err = json.Unmarshal([]byte(body), &result)
+		if err == nil && result.Health == "true" {
+			log.Info("etcd daemon is already running")
+			return nil
+		}
+	}
+
+	etcdStart := exec.Command("etcd",
+		"-listen-client-urls", listenClientUrls,
+		"-advertise-client-urls", advClientUrls,
+		"-listen-peer-urls", listenPeerUrls,
+		"-initial-advertise-peer-urls", initialAdvPeerUrls,
+		"--initial-cluster", "default="+listenPeerUrls)
+
+	// Don't kill chlid process (etcd) upon ^C (SIGINT) of main glusterd process
+	etcdStart.SysProcAttr = &syscall.SysProcAttr{
+		Setpgid: true,
+	}
+
+	err = etcdStart.Start()
+	if err != nil {
+		log.WithField("error", err.Error()).Fatal("Could not start etcd daemon.")
+		return err
+	}
+
+	if check := checkHealth(15, listenClientUrls); check != true {
+		fmt.Println("check is: ", check)
+		log.Fatal("Health of etcd is not proper. Check etcd configuration.")
 	}
 
 	return nil
--- a/etcdmgmt/etcd-mgmt.go
+++ b/etcdmgmt/etcd-mgmt.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"syscall"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -34,6 +35,11 @@ func StartEtcd() error {
 		"-listen-peer-urls", listenPeerUrls,
 		"-initial-advertise-peer-urls", initialAdvPeerUrls,
 		"--initial-cluster", "default="+listenPeerUrls)
+
+	// Don't kill chlid process (etcd) upon ^C (SIGINT) of main glusterd process
+	etcdStart.SysProcAttr = &syscall.SysProcAttr{
+		Setpgid: true,
+	}
 
 	err = etcdStart.Start()
 	if err != nil {


etcdmgmt/etcd-mgmt.go
```diff
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"syscall"
```
kshlm: Don't use `syscall`. Use `golang.org/x/sys/unix` instead. Refer the NOTE in https://golang.org/pkg/syscall/

atinmu:
Commit message needs some correction. This change is not limited to SIGINT, be it abrupt or graceful shutdown of parent process which is glusterd in this case, we'd need to ensure that etcd doesn't go down. I'd suggest to change the commit message like following:

"Ensure etcd doesn't go down on a glusterd shutdown"

atinmu:
Have you tested Multiple glusterd restarts with this change? What happens when etcd instance is already running and glusterd comes back?

etcdmgmt/etcd-mgmt.go
```diff
@@ -35,6 +36,11 @@ func StartEtcd() error {
 		"-initial-advertise-peer-urls", initialAdvPeerUrls,
 		"--initial-cluster", "default="+listenPeerUrls)
 
+	// Don't kill chlid process (etcd) upon ^C (SIGINT) of main glusterd process
```
atinmu: As I mentioned in my earlier comment this is not limited to ^C. Please update the comment accordingly

etcdmgmt/etcd-mgmt.go
gaurav36: @kshlm  will change in next PR

etcdmgmt/etcd-mgmt.go
gaurav36: Could you tell me what other signal apart from SIGINT can result into child process shutdown in this case, IMO currently only SIGINT is causing child process shut down.

gaurav36: >> Have you tested Multiple glusterd restarts with this change? What happens when etcd instance is already running and glusterd comes back?
@atinmu  yes i have tested above case. It will work normally.

atinmu: How about SIGKILL, SIGTERM? Aren't you aware that these signals are for shutting down a process?

Regarding running etcd instance from the code its straight forward that there is no check of whether a running etcd instance is available and if so etcd start shoild be skipped. When I pulled your changes I could easily see a defunct etcd process on a glusterd restart. I am not sure about what you have tested.

Moreover, this change will solve the issue of not letting the etcd instance go down on a glusterd restart but this doesn't cover the definition of daemonizing a process which should be a framework to manage multiple daemons. However I feel we can go ahead with it for now and later can come up with a generic framework.

gaurav36: @atinmu 
>> How about SIGKILL, SIGTERM? Aren't you aware that these signals are for shutting down a process?

I am aware very well about these all the signal. It seems to miscommunication in my last comment. I asked that how are you getting child process killed without ^C. 
suppose my glusterd2 is running and glusterd2 have started etcd child process and if you issue kill (SIGKILL/SIGTERM) from terminal then its not killing child process (etcd daemon). It will only kill etcd daemon when you issue ^C from terminal. typing ^C in unix send signal to every process group. I asked how did you killed etcd daemon from SIGKILL and SIGTERM.

>> Regarding running etcd instance from the code its straight forward that there is no check of whether a running etcd instance is available and if so etcd start shoild be skipped. When I pulled your changes I could easily see a defunct etcd process on a glusterd restart. I am not sure about what you have tested.

you should have described this in the bug itself. doing this is not a big deal.

 >> Moreover, this change will solve the issue of not letting the etcd instance go down on a glusterd restart but this doesn't cover the definition of daemonizing a process which should be a framework to manage multiple daemons.


While creating issue you did not talked about its framework so i was not aware about it. i was mainly focusing on "etcd instance can't be brought down when glusterd goes down since the other consumers like NSR would still be able to access it" part as you have created in issue #66

atinmu: I am honestly failing to understand your arguments here.  

Point 1 : Currently glusterd2 daemon itself is not daemonized and hence once you execute it it doesnt return back to the prompt and with that you bring in interrupts, however this wouldn't be true as we have to daemonize it (this is how existing glusterd also behave). So if your observation is that etcd doesnt go down in SIGKILL or SIGTERM we dont need this change at all as we can run the current glusterd2 instance in background to avoid the command prompt to get stuck.

Point 2 : Being a programmer I expect these basic logics to be thought out before sending the changes. A github issue will highlight what is the problem we are tryung to solve but it will *never* say how are you going to achieve it. Starting etcd instance without checking whether an already etcd process is running or not is a *mistake*, no matter whatever reason you provide and I dont want to get into that further.

Point 3: Word 'daemonize' has some significant meaning. Daemonizing a process doesn't only mean changing the a single attribute of a process.  Again to repeat daemonize itself explains what needs to be done.

atinmu: You were partially right that if a SIGKILL/SIGTERM is sent to the process, child process remains as it is but thats only true for a foreground process. If glusterd2 is run in background its not correct.

gaurav36: @atinmu 

>> Point 1 : Currently glusterd2 daemon itself is not daemonized and hence once you execute it it doesnt return back to the prompt and with that you bring in interrupts, however this wouldn't be true as we have to daemonize it (this is how existing glusterd also behave). So if your observation is that etcd doesnt go down in SIGKILL or SIGTERM we dont need this change at all as we can run the current glusterd2 instance in background to avoid the command prompt to get stuck.

Yes, that's true. When glusterd2 daemon itself will daemonized then we don't need to execute ^C and without this change it will work. but its a safe to don't include etcd daemon id in parent process group id.

 >> Point 2 : Being a programmer I expect these basic logics to be thought out before sending the changes. A github issue will highlight what is the problem we are tryung to solve but it will never say how are you going to achieve it. Starting etcd instance without checking whether an already etcd process is running or not is a mistake, no matter whatever reason you provide and I dont want to get into that further.

That i will take care in my next PR

>> Point 3: Word 'daemonize' has some significant meaning. Daemonizing a process doesn't only mean changing the a single attribute of a process. Again to repeat daemonize itself explains what needs to be done.

I didn't expect that we are expecting daemonizing framework by end of this month. If i remember correctly by this EOM we just need to avoid ectd killing upon ^C. 

>>    You were partially right that if a SIGKILL/SIGTERM is sent to the process, child process remains as it is but thats only true for a foreground process. If glusterd2 is run in background its not correct.

Again i will ask you did your etcd daemon killed when you run glusterd2 process in background and if yes i would like to know how did it happen.

atinmu: >> Point 1 : Currently glusterd2 daemon itself is not daemonized and hence once you execute it it doesnt return back to the prompt and with that you bring in interrupts, however this wouldn't be true as we have to daemonize it (this is how existing glusterd also behave). So if your observation is that etcd doesnt go down in SIGKILL or SIGTERM we dont need this change at all as we can run the current glusterd2 instance in background to avoid the command prompt to get stuck.

>>>Yes, that's true. When glusterd2 daemon itself will daemonized then we don't need to execute ^C and without this change it will work. but its a safe to don't include etcd daemon id in parent process group id.

I am just wondering why SIGINT is behaving differently here. Do you have any background?

 >> Point 2 : Being a programmer I expect these basic logics to be thought out before sending the changes. A github issue will highlight what is the problem we are tryung to solve but it will never say how are you going to achieve it. Starting etcd instance without checking whether an already etcd process is running or not is a mistake, no matter whatever reason you provide and I dont want to get into that further.

>>>That i will take care in my next PR

Good that we have an agreement here

>> Point 3: Word 'daemonize' has some significant meaning. Daemonizing a process doesn't only mean changing the a single attribute of a process. Again to repeat daemonize itself explains what needs to be done.

>>> I didn't expect that we are expecting daemonizing framework by end of this month. If i remember correctly by this EOM we just need to avoid ectd killing upon ^C. 

If that's the case use a different issue number or atleast don't use the term 'daemonize'

>>    You were partially right that if a SIGKILL/SIGTERM is sent to the process, child process remains as it is but thats only true for a foreground process. If glusterd2 is run in background its not correct.

>>>Again i will ask you that did your etcd daemon killed when you run glusterd2 process in background and if yes i would like to know the process of killing because in my system it didn't killed.

This is a mistake from end. It doesn't happen like that. You are correct.

So to summarize following changes/actions are required:

Add a check of whether etcd is running or not before starting etcd process to avoid any defunct instances
Address @kshlm's comments
Remove the word 'daemonize' from the commit message and change the heading of the issue

etcdmgmt/etcd-mgmt.go
```diff
@@ -28,6 +64,19 @@ func StartEtcd() error {
 
 	initialAdvPeerUrls := "http://" + hostname + ":2380"
 
+	// Checking health of etcd cluster. If etcd daemon already running then
```
atinmu: Is there any specific reason for not to go for a pid check? I can understand that the same can go for a toss if an etcd instance is run independently but that should never happen IMO.

etcdmgmt/etcd-mgmt.go
atinmu: This is still not addressed.

etcdmgmt/etcd-mgmt.go
atinmu: Also what happens if etcd health is actually bad and etcd is already running?

gaurav36: @atinmu  By checking health of the cluster it seems to be problem here. We need to come up with different way like storing pidof etcd it self in etcd. @kshlm  you can also give your opinion in this case.

kshlm: I'd already given you my opinion on this. Attempting to store information about an etcd instance in etcd is not right. Just to make it clear to you, when you need to decide if something must be stored in etcd, ask the question 'Is this information relevant to the whole cluster or just to this one peer?'. Only store information relevant to the cluster in etcd. Everything else, use local storage, possibly using `/var/lib` or `/var/run`.

I'd also told about my preferred approach to check etcd liveness on GD2 startup. In simple terms,
- Check if you can connect to etcd and check for cluster health (this is the same as what you are doing to check if etcd is operational after you start it)
- Start etcd
- Wait for etcd to become operational

gaurav36: @kshlm   but if i have go with health checkup approach then, suppose we are having 5 node clustersetup and one node went down then health of etcd will not be good, and upon stopping glusterd etcd is continuously running. upon starting glusterd it will check whether health is good or not if health is not good then it will start it again.
@kshlm  second concern is that if we go through health check approach and etcd is already running but health is not good then upon restarting glusterd will create defaunt process of etcd

storing pid of etcd in seperate file seems to be good approach.

gaurav36:
closing this PR as updated duplicate PR available https://github.com/gluster/glusterd2/pull/76

PR #76, opened 2016-01-29
Daemonize etcd

gaurav36:
Do not kill etcd daemon upon issuing ^c (SIGINT) in glusterd daemon
    
    Currently when etcd daemon start upon starting of glusterd2 then upon
    performing ^C (SIGINT) its killing etcd daemon.
    
    With this patch it will set child process id to different from main
    process id group and, etcd daemon will not be kill upon performing
    SIGINT (^C) of main process.
    
 
    
    Issue #66
    

utils/utils.go
```diff
@@ -266,3 +269,62 @@ func isBrickPathAlreadyInUse(brickPath string) bool {
...
+			"path": dir,
+		}).Fatal("directory path is not a writable")
+	}
+	// defer happens in LIFO
+	defer syscall.Unlink(t.Name())
+	defer t.Close()
+}
+
+// Function to check whether the process with given pid exist or not in the system
+func CheckPidExist(pid int) bool {
```
atinmu: Isn't it better to name it as CheckProcessExist?

etcdmgmt/etcd-mgmt.go
```diff
@@ -39,43 +88,70 @@ func StartEtcd() (*exec.Cmd, error) {
 		"-initial-advertise-peer-urls", initialAdvPeerUrls,
 		"--initial-cluster", "default="+listenPeerUrls)
 
+	// Don't kill chlid process (etcd) upon ^C (SIGINT) of main glusterd process
+	etcdCmd.SysProcAttr = &syscall.SysProcAttr{
```
atinmu: Please add a TODO to remove syscall usage

context/context.go
```diff
@@ -29,7 +31,7 @@ var (
 	Rest      *rest.GDRest
 	TxnFw     *transaction.GDTxnFw
 	OpVersion int
-	EtcdCtx   *exec.Cmd
+	EtcdCtx   *os.Process
```
atinmu: Can we rename this to ETCDProcessCtx?

etcdmgmt/etcd-mgmt.go
```diff
@@ -6,16 +6,65 @@ import (
...
+		if result.Health == "true" {
+			timer.Stop()
+			break
+		}
+	}
+	return true
+}
+
+var (
+	etcdPidDir  = "/var/run/gluster/"
```
kshlm: This cannot be hardcoded. If you are going to store the etcdPid in GD2's workdir, use config.WorkDir. Don't hardcode paths like this.

PR #83, opened 2016-02-26
Delete peer

atinmu:


commands/peers/addpeer.go
```diff
@@ -39,37 +39,37 @@ func addPeerHandler(w http.ResponseWriter, r *http.Request) {
...
 		log.WithFields(log.Fields{
-			"error":  e,
-			"member": p.Name,
+			"error": e,
+			"peer":  req.Name,
 		}).Error("Failed to add member into etcd cluster")
 		rest.SendHTTPError(w, http.StatusInternalServerError, e.Error())
 		return
 	}
+	log.Info("rsp UUID", *rsp.UUID)
```
gaurav36: logging message can be improve here.

atinmu: This was added for debugging and I forgot to clean it up, will remove it

context/context.go
```diff
@@ -63,7 +63,7 @@ func doInit() {
 
 	utils.InitDir(config.LocalStateDir)
 
-	initMyUUID()
+	MyUUID = initMyUUID()
```
gaurav36: where are you using this MyUUID ?. if you are not using then you can ignore return value of initMyUUID()

main.go
```diff
@@ -28,6 +29,9 @@ func main() {
 		context.Rest.SetRoutes(c.Routes())
 	}
 
+	// Store self information in the store
+	peer.AddSelfDetails()
```
gaurav36: s/AddSelfDetails/AddSelfPeerDetails

peer/store-utils.go
```diff
@@ -70,10 +70,18 @@ func GetPeers() ([]Peer, error) {
...
 			continue
 		}
+		if p.ID.String() == context.MyUUID.String() {
+			// Skip self's information
+			continue
+		}
 		peers[i] = p
+		i = i + 1
+	}
+	if i == 0 {
```
gaurav36: do we really need i variable here ? you can make it without using i variable.

gaurav36:
you can add test case for this PR if you like.

context/context.go
atinmu: MyUUID is a global variable in context and the same been referred at `ValidateAdd` in `rpc/services/peer-rpc-svc.go`

main.go
atinmu: Self can't be a peer :)
Do you have any other better name(s) in your mind?

peer/store-utils.go
atinmu: This is intended. If we use i defined in the for loop then it gets automatically incremented after every iteration and even a `continue` statement would increment it which we don't want since we want to return the peer structure populated with all the peers but the self.

main.go
gaurav36: AddSelf(peer/node)details()

peer/store-utils.go
gaurav36: i don't mean to say using i in for loop. i mean to say before for loop you are making empty slice of peers. so you can simply "return peers, nil". if there is peer in the cluster then peers slice will not have anything so it can be done by simply using "return peers, nil"  instead of using extra i variable.

main.go
gaurav36: AddSelfDetails is no more meaningful because Self can be volume, or might be other information. for making it more clear you can use AddSelf(peer/node)details()

commands/peers/deletepeer.go
```diff
@@ -3,29 +3,63 @@ package peercommands
...
+	log "github.com/Sirupsen/logrus"
+	etcdclient "github.com/coreos/etcd/client"
+	etcdcontext "golang.org/x/net/context"
 
 	"github.com/gorilla/mux"
 )
 
 func deletePeerHandler(w http.ResponseWriter, r *http.Request) {
-	p := mux.Vars(r)
+	peerReq := mux.Vars(r)
```
gaurav36: write comment about use case of this function.

peer/store-utils.go
atinmu: Yes you are right, I will do that change!

commands/peers/deletepeer.go
atinmu: What do you mean by use case here?

main.go
atinmu: Can you elaborate when Self can mean a volume?

gaurav36: well, later in some use case like (AddSelfVolumeDetails or something else). by making it peer or node its making more meaningful.

commands/peers/deletepeer.go
gaurav36: i mean to say what this function is doing. we should write comment just above function

main.go
atinmu: Self always indicates the node and volume is something which is shared and not owned by a single node in gluster terminology. I can't buy in this comment.

commands/peers/deletepeer.go
atinmu: Comments are must when the functions are exposed and this function is certainly not since it starts with a small letter. These comments actually get integrated in go doc where you can view the functions accessible from the package and that's why comments explaining what the function does is really needed.

commands/peers/addpeer.go
```diff
@@ -39,37 +39,36 @@ func addPeerHandler(w http.ResponseWriter, r *http.Request) {
 
 	//FIXME: In the correct add process, the peer being probed would add it's details to the store once it's been validated. The code below is just a temporary stand-in to show how the API's would work
```
gaurav36: imo above comment can be remove now. "//FIXME: In the correct add process, the peer being probed would add it's details to the store once it's been validated. The code below is just a temporary stand-in to show how the API's would work "

atinmu: Yes, will remove it.

peer/store-utils.go
```diff
@@ -58,9 +58,9 @@ func GetPeers() ([]Peer, error) {
 		return nil, err
 	}
 
-	peers := make([]Peer, len(pairs))
-
-	for i, pair := range pairs {
+	peers := make([]Peer, len(pairs)-1)
+	i := 0
```
gaurav36: could you remove i variable from here.

peer/store-utils.go
```diff
@@ -70,7 +70,12 @@ func GetPeers() ([]Peer, error) {
 			}).Error("Failed to unmarshal peer")
 			continue
 		}
+		if p.ID.String() == context.MyUUID.String() {
+			// Skip self's information
+			continue
+		}
 		peers[i] = p
+		i = i + 1
```
gaurav36: no use of i variable.

gaurav36:
LGTM

PR #87, opened 2016-03-01
etcd proxy configuration support

### LINKED ISSUE #84: etcd proxy (client) support
Given a add peer request with proxy = true etcd should act as a client on that node, default shoud be to add etcd node as server

atinmu:
Adds the code flow for adding a proxy etcd node

Closes #84

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,16 +5,21 @@ import (
 	"os/signal"
 
 	"github.com/gluster/glusterd2/commands"
+	"github.com/gluster/glusterd2/config"
 	"github.com/gluster/glusterd2/context"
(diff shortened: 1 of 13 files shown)


main.go
```diff
@@ -16,6 +18,9 @@ import (
 func main() {
 	log.Info("GlusterD starting")
 
+	utils.InitDir(config.LocalStateDir)
+	context.MyUUID = context.InitMyUUID()
```
gaurav36: can we assign value in context before starting etcd ?

atinmu: Yes, why can't we? This change was done to ensure that we initialize MyUUID such that `peer.AddSelfDetails()` can refer to it and store its own information into the store.

PR #90, opened 2016-06-17
Extending transaction framework to multiple nodes

kshlm:
This is still a WIP.


prashanthpai:
On our conversation of having response headers, one good example is returning `X-Transaction-Id` header on every response. This transaction id can be looked up in the logs for further information.

kshlm:
ok to test

kshlm:
This is basically in a workable state now. The only problem is, using a `map[string]interface{}` as a dict isn't working out as I expected. This cannot be reliably encoded, sent over RPC and decoded, when storing non golang native types. This causes transaction using multiple-nodes to fail right now. But this failures also show that the rollback mechanism is working.

@atinmu @prashanthpai could you guys review this?

commands/command.go
```diff
@@ -12,8 +12,8 @@ import (
 type Command interface {
```
prashanthpai: The name `Command` seems to be a misnomer and a little misleading. Something like `Operation` would be better. Don't bother now - something to think about at later stage.

peer/store-utils.go
```diff
@@ -11,12 +11,21 @@ import (
...
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/pborman/uuid"
 )
 
 const (
 	peerPrefix string = store.GlusterPrefix + "peers/"
 )
 
+var (
```
prashanthpai: Is this a workaround for something ? Please add a comment on why these new names and mapping has been done.

kshlm: The comments were actually in the commit that introduced this. https://github.com/gluster/glusterd2/pull/90/commits/74190809a3c1525ca5164ddeb8940d4d9a86aa88

This was  done to allow these functions to be mocked during unit testing.

prashanthpai: Ah ok. Although it seems like a bad practice to change actual code to fix unit tests.

atinmu: Can't we name them like 'MockFunctionName' with naming convention, at least I used it in my earlier commits.

transaction/simple.go
```diff
@@ -2,29 +2,35 @@ package transaction
...
+	// Stage is the registered name of the staging StepFunc
 	// Stage function verifies if the node can perform an operation.
-	Stage StepFunc
+	Stage string
+	// Commit is the registered name of the commit StepFunc
 	// Commit performs the operation on the a node
-	Commit StepFunc
+	Commit string
+	// Store is the registered name of the store StepFunc
 	// Store stores the results of an operation. This will only be run on the leader
```
prashanthpai: In Store, is the "leader" the node that initiates transaction ?

transaction/simple.go
```diff
@@ -2,29 +2,35 @@ package transaction
...
-	Store StepFunc
+	Store string
+	// Rollback is the registered name of the rollback StepFunc
 	// Rollback rollsback any changes done by Commit
-	Rollback StepFunc
+	Rollback string
```
prashanthpai: May be name it as `RollbackCommit`.

transaction/simple.go
```diff
@@ -2,29 +2,35 @@ package transaction
...
-	Stage StepFunc
+	Stage string
+	// Commit is the registered name of the commit StepFunc
 	// Commit performs the operation on the a node
-	Commit StepFunc
+	Commit string
+	// Store is the registered name of the store StepFunc
 	// Store stores the results of an operation. This will only be run on the leader
-	Store StepFunc
+	Store string
```
prashanthpai: On failure store, what will be the action ? Would commit rollback be initiated ?

transaction/simple.go
```diff
@@ -2,29 +2,35 @@ package transaction
 
 import (
 	"github.com/gluster/glusterd2/context"
+
+	"github.com/pborman/uuid"
 )
 
 // SimpleTxn is transaction with fixed stage, commit and store steps
 type SimpleTxn struct {
```
prashanthpai: Taking an operation such as volume create as example, I'd prefer to see some detailed comment as to how this mechanism will adhere (resilient to TOCTOU and network disruptions) to some well known protocols such as 2PC or 3PC/E3PC which has been widely implemented and has formal proofs. If I missed this in design docs or specs, please point me there.

transaction/simple.go
```diff
@@ -2,29 +2,35 @@ package transaction
...
 	// Ctx is the transaction context
 	Ctx *context.Context
 	// Nodes are the nodes where the stage and commit functions are performed
-	Nodes []string
+	Nodes []uuid.UUID
 	// LockKey is the key to be locked
 	LockKey string
```
prashanthpai: Where does the locking begin and end ? IOW, at what step is the lock taken and then relinquished ?

commands/volumes/volume-create.go
```diff
@@ -120,6 +121,39 @@ func rollBackVolumeCreate(c *context.Context) error {
...
 
+func registerVolCreateStepFuncs() {
+	var sfs = []struct {
+		name string
+		sf   transaction.StepFunc
+	}{
+		{"vol-create.Stage", validateVolumeCreate},
+		{"vol-create.Commit", generateVolfiles},
+		{"vol-create.Store", storeVolume},
+		{"vol-create.Rollback", rollBackVolumeCreate},
```
prashanthpai: This function only removes brick paths. Should gluster do it ?

Besides, I don't see any code to remove volfiles generated in generateVolfiles.

prashanthpai: On a broader note, shouldn't volfiles be written into etcd store instead of storing local filesystem ? That way any glusterd instance can serve volfile of any volume.

peer/store-utils-mock.go
```diff
@@ -0,0 +1,19 @@
+package peer
+
+// This file contains mock functions to be used during unit tests.
```
prashanthpai: Then this probably doesn't belong here in package `peer` and should be in package `tests` ?

transaction/simple.go
kshlm: The SimpleTxn resembles the existing transaction mechanism in GD-1, which is a two-phase commit. It brings in a much needed rollback which isn't present GD-1.

The stage step is the commit-request phase, and the commit step is the commit-phase. In addition to these We have lock and unlock steps to handle concurrent actions on a resource.

In the stage step, the leader (or co-ordinator) requests each follower (or cohort) to validate that they can perform the commit. The commit is done once everyone replies with a yes. If the commit fails on anyone, any changes done are rolled back.

We do have a problem here, as is common in a 2-phase commit, that we don't have a timeout. This is something we'll need to address.

transaction/simple.go
kshlm: You need to refer to `NewSimpleTxn()` which builds the `Txn` object that will be used.
Lock is the first step in the transaction, and unlock the last.

commands/volumes/volume-create.go
kshlm: I've just reused the existing functions here. But, you're correct. We will need to cleanup volfiles if any were generated.

We do need to remove the brick paths, as the volume create failed, and they are no longer useful. This allows them to be reused easily. We don't remove brick-paths when we're doing a volume-delete as it might still contain data.

peer/store-utils-mock.go
kshlm: The mock functions provided here are for functions that are exported by this package. They are for other use in unit tests of other packages that make use of this package and functions exported by it.
Instead of dumping these mock functions into a separate test package containing all other mocked functions, it stay in this package itself close to the functions being mocked.

kshlm:
I'm gonna merge this is now, so that I can start converting the other commands to the framework. It's been working well. Any thing that needs to be addressed will be fixed as it's found.

PR #93, opened 2016-06-23
Set /var/lib/glusterd as default local state directory

### LINKED ISSUE #33: local state directory default path should be changed

prashanthpai:
This closes #33

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,8 +6,6 @@ package config
 
 import (
 	"flag"
-	"os"
-	"path"
 )
 
 // All configuration values which should be available for use by other packages need to be defined here as global variables.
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -34,9 +32,8 @@ var (
 )
 
 func initLocalStateDir() {
-	wd, _ := os.Getwd()
 
-	flag.StringVar(&LocalStateDir, "localstatedir", path.Join(wd, "glusterd"), "Directory to store local state information.")
+	flag.StringVar(&LocalStateDir, "localstatedir", "/var/lib/glusterd", "Directory to store local state information.")
```
kshlm: This is something I'm against. The default shouldn't be a directory that would possibly used in a production environment. Doing so can cause frustrations during development.

A default based on the specific system can be set when GD2 is packaged.

PR #94, opened 2016-06-24
Configure etcd -data-dir location

prashanthpai:

### DIFF
--- a/etcdmgmt/etcd-mgmt.go
+++ b/etcdmgmt/etcd-mgmt.go
@@ -33,6 +33,9 @@ var (
 	ExecName = "etcd"
 	// Configuration directory for storing etcd configuration
 	ETCDConfDir = "/var/lib/glusterd"
+	// Etcd -data-dir location. This contains WAL files and log snapshots
+	// https://coreos.com/etcd/docs/latest/admin_guide.html#data-directory
+	ETCDDataDir = "/var/lib/etcd"
 	ETCDEnvFile = ETCDConfDir + "/etcdenv.conf"
 	// If this file is touched on ETCDConfDir that indicates that etcd
 	// instance need to come with proxy mode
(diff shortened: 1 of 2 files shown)


etcdmgmt/etcd-mgmt.go
```diff
@@ -33,6 +33,9 @@ var (
 	ExecName = "etcd"
 	// Configuration directory for storing etcd configuration
 	ETCDConfDir = "/var/lib/glusterd"
+	// Etcd -data-dir location. This contains WAL files and log snapshots
+	// https://coreos.com/etcd/docs/latest/admin_guide.html#data-directory
+	ETCDDataDir = "/var/lib/etcd"
```
kshlm: This is something that I've been meaning to change from a long time. I don't like that these paths have been hardcoded, and that too paths that are accessible only to root.

If you can, please change these paths to follow LocalStateDir. If no etcd paths are provided, the etcd conf and data should be inside LocalStateDir. If the user requires they can provide an alternative path.

When GD2 packages are built later, the correct paths for the respective environment will be set during the build process.

prashanthpai: If it is guaranteed that lifecycle of etcd daemons/processes are exclusively controlled by GD2 only, this should be straight-forward and not confusing. If not (user invokes systemctl or configures etcd to spawn at startup), I need some more clarification on this. Made the default set to `/var/lib/etcd` based on conventions used in etcd packaging as illustrated in this [systemd service file](https://github.com/coreos/etcd/blob/master/contrib/systemd/etcd.service)

kshlm: We don't yet support an externally managed etcd yet, but it is on the TODO list.
The externally managed etcd, if it has a server running on a GD2 host, will have it's conf/store as per the conventions in the etcd package.

Currently we only support etcd managed by ourselves. For this sort of an etcd, it's my opinion that, its conf and data should be under the GD2 paths. This managed etcd should ideally not conflict with any other etcd daemons running on the system. We would be clashing ports with other etcds, and that is something that we should change.

prashanthpai:
Abandoning this in favor of PR #96 . Will send another PR on top of PR #96 if need be.

PR #95, opened 2016-06-24
Allow binding to specific network interface

prashanthpai:
Currently, both the RPC and REST services bind to their respective
ports on all network interfaces. This may not always be desirable.
For example, when a machine is part of multiple networks, the
admin may want to make the REST service available only on one
network interface for external consumption.

This change adds command line options to specify both IP and port
these services should bind to. The default behaviour is not changed
i.e bind on all available interfaces.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -12,13 +12,21 @@ import (
 // Any configuration that doesn't require custom default values can have their flags initialised right here.
 // Any configuration that require custom default values should be initalized manually.
 var (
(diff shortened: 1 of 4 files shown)


kshlm:
You could have just set RestAddress or RPCAddress as 'ip:port', for eg. '192.0.2.1:24007'. No need to split the option into an address and a port.

kshlm:
:+1: Looks good to me.

PR #96, opened 2016-06-24
Add support for config files

kshlm:
GD2 can be configured using config files. The very good spf13/viper
package has been used to provide this facility.

GD2 will search for a file named "glusterdconf.(toml|yaml|json)" in
"/etc/glusterd" and the current working directory by default. A specific
config file can also be given using the `--config` flag on the command
line.

All of the command line flags, can also be configured in the config
file. The preference for values will be in the order
  Flags > Config file > Coded defaults

The glusterd/config package has been removed. Any GD2 package requiring
access to the config needs to now import the spf13/viper package and use
the relevant Get* function to get the configuration.

For eg.
```go
import config "github.com/spf13/viper"

func someFunc() {
  rpcAddress := config.GetString("rpcaddress")
}
```


kshlm:
@atinmu @prashanthpai What do you guys think of this?

config.go
```diff
@@ -0,0 +1,94 @@
...
+		"/etc/glusterd",
+		".",
+	}
+	defaultConfName = "glusterdconf"
+)
+
+// parseFlags sets up the flags and parses them, this needs to be called before any other operation
+func parseFlags() {
+	flag.StringVar(&confFile, "config", "", "Configuration file for GlusterD. By default looks for glusterdconf.(yaml|toml|json) in /etc/glusterd and current working directory.")
+	flag.StringVar(&localStateDir, "localstatedir", "", "Directory to store local state information. Defaults to current working directory.")
```
prashanthpai: Can you use `os.Getwd()` here and lose `setDefaults()` function altogether, makes things simpler.

config.go
```diff
@@ -0,0 +1,94 @@
...
+	rpcAddress    string
+	localStateDir string
+)
+
+var (
+	defaultConfPaths = []string{
+		"/etc/glusterd",
+		".",
+	}
+	defaultConfName = "glusterdconf"
```
prashanthpai: Purely a personal preference: drop the 'conf' at the end. IOW, files can simply be `glusterd.yaml` or `glusterd.json`

prashanthpai:
spf13/viper seems like a good solution. Does it makes sense to have a wrapper subpackage (retain config) around viper to get around having to import viper. What would happen if global vars are accessed directly and not via viper ?

config.go
kshlm: It was a tossup between having conf in the name or not. In the end I felt that the name should give an indication that this was the configuration for GD2, so I kept the conf. But I agree that having it just `glusterd.{yaml|toml|json}` is better. I'll change this.

config.go
kshlm: This was a carry forward from what was there before. Also, as I've added in the comment for `setDefaults` it can be used to setup config options which are not available as flags. I don't expect all config options to be available as flags, even though right now, they are all flags.

kshlm:
> spf13/viper seems like a good solution. Does it make sense to have a wrapper subpackage (retain config ?) around viper to get around having to import viper. What would happen if global vars such as restAddress are accessed directly and not via viper ?

While converting I did try to see if I could retain the earlier config package. But it would just have been an extra level of indirection, that wasn't required IMO.

The variables just hold values passed by the command line flags. These are not the final values of the config options. The Viper package sets the final values for all the options by collating all the values from the set defaults, the config files, the environment variables (we don't use this) and the command line flags. If a option is set in multiple locations, it selects the final value following the precedence Flags > Env > File > Defaults.

prashanthpai:
LGTM :+1:

kshlm:
Thanks for the review @prashanthpai. Merging this.

atinmu:
wait @kshlm, I am reviewing it.

kshlm:
> wait @kshlm, I am reviewing it.
Um, I've just merged it. Just add any comments you have, I'll answer them here or address them in a new PR if required.

atinmu:
@kshlm : I tried to create a glusterd.yaml file in '/etc/glusterd' with following content:
``localstatedir = "/var/lib/glusterd"                                                  
  rpcaddress = ":24008"                                                                
  restaddress = ":24007"
``
While bringing up glusterd2 instance I see the following log:
DEBU[0000] failed to read any config files, continuing with defaults  config=glusterd.(toml|yaml|json) error=While parsing config: yaml: unmarshal errors:
  line 1: cannot unmarshal !!str `localst...` into map[string]interface {} paths=[/etc/glusterd .]
DEBU[0000] running with configuration

kshlm:
> @kshlm : I tried to create a glusterd.yaml file in '/etc/glusterd' with following content:
> localstatedir = "/var/lib/glusterd" rpcaddress = ":24008" restaddress = ":24007" 
> While bringing up glusterd2 instance I see the following log:
> DEBU[0000] failed to read any config files, continuing with defaults config=glusterd.(toml|yaml|json) error=While parsing config: yaml: unmarshal errors:
> line 1: cannot unmarshal !!str localst... into map[string]interface {} paths=[/etc/glusterd .]
> DEBU[0000] running with configuration

You need to create a proper yaml file. Yaml doesn't use '=' to assign values. The relevant yaml file would be,

```yml
localstatedir: "/var/lib/glusterd"
rpcaddress: ":24007"
restaddress: ":24008"
```

What you've created is a [TOML](https://github.com/toml-lang/toml) file, but gave the extension yaml.

glusterd.toml.example
```diff
@@ -0,0 +1,3 @@
+localstatedir = "/var/lib/glusterd"
+rpcaddress = ":24008"
+restaddress = ":24007"
```
atinmu: Can't we also have log level as an option here?

kshlm: Yes you can. All command line flags will be available in config files. The config files can have extra options that aren't available as command line flags.

This file is just an example, and doesn't document all the options.

PR #103, opened 2016-07-15
Attach peer fixup

prashanthpai:


main.go
```diff
@@ -44,11 +44,7 @@ func main() {
 		context.Rest.SetRoutes(c.Routes())
 	}
 
-	// Store self information in the store if GlusterD is coming up for
-	// first time
-	if context.Restart == false {
-		peer.AddSelfDetails()
-	}
+	peer.AddSelfDetails()
```
atinmu: Have you tested this when glusterd restarts?

prashanthpai: Yes. I can see the node details in etcd store soon after glusterd start. One small caveat is that on every restart of glusterd, the etcd entry attributes createdIndex and modifiedIndex numeric value is bumped up.

peer/store-utils.go
```diff
@@ -79,12 +79,12 @@ func GetPeers() ([]Peer, error) {
 	if err != nil || pairs == nil {
 		return nil, err
 	}
-	// If there is only one entry in the store this indicates its a stand
-	// alone cluster, no need to return back it's own information
-	if len(pairs) == 1 {
+	if len(pairs) == 0 {
```
atinmu: Can this ever happen given that self details is added into etcd store?

atinmu: @prashanthpai : As discussed, initMyUUID () is been called twice and that has caused this issue. I don't see that change reflected here. With this change, are you able to see both the peer details in get peers request? I doubt so since context.Restart is never set true which means self details are not added into the etcd store?

prashanthpai: Never in the happy path. I can't imagine other error scenarios. Do you prefer removing this check ?

prashanthpai: I have removed the second call to `InitMyUUID()` in `context/global.go`. I can see the peer details on startup.

atinmu: My bad! I could see that now. I am used to gerrit interface :(

prashanthpai:
@atinmu Removed the check for number of peers.

PR #108, opened 2016-07-19
Fix crash caused on peer attaching local node

### LINKED ISSUE #101: Adding current node as a peer fails with etcd crash
If a node tries to add itself as a peer, it fails with `error=client: etcd cluster is unavailable or misconfigured member=gd2-2` and etcd becomes defunct.

prashanthpai:
This closes #101

### DIFF
--- a/commands/peers/addpeer.go
+++ b/commands/peers/addpeer.go
@@ -35,6 +35,20 @@ func addPeerHandler(w http.ResponseWriter, r *http.Request) {
 		req.Name = req.Addresses[0]
 	}
 
+	local_node := false
+	for _, addr := range req.Addresses {
+		local, err := utils.IsLocalAddress(addr)
+		if err != nil || local == true {
+			local_node = true
+			break
+		}
+	}
+
+	if local_node == true {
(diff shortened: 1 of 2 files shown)


commands/peers/addpeer.go
```diff
@@ -35,6 +35,20 @@ func addPeerHandler(w http.ResponseWriter, r *http.Request) {
 		req.Name = req.Addresses[0]
 	}
 
+	local_node := false
+	for _, addr := range req.Addresses {
+		local, err := utils.IsLocalAddress(addr)
+		if err != nil || local == true {
```
atinmu: if err != nil then you should error out with a different code than ErrPeerLocalNode.

prashanthpai: This was deliberate. This err can be returned by call to `net.LookupIP` (internally done in `IsLocalAddress`) which can fail on any one of the multiple addresses which is OK because the attach peer API allows specifying multiple addresses. The goal here is to identify if the peer is local node. If DNS resolution fails one one of the addresses, that should be okay at this point in time. If the host is not reachable, it'll fail later during the RPC call right. What do you think ?

Issue #110, opened 2016-07-25
myUUIDFile path initialised using local state dir from config which is unavailable during compile time


prashanthpai:
The following will never get initialised properly and will be set to string `uuid`

~~~
myUUIDFile = path.Join(config.GetString("localstatedir"), "uuid")
~~~

Simple reproducer example:

```
package main

import (
	"fmt"
	config "github.com/spf13/viper"
)

var (
	etcdVar = config.GetString("etcd")
)

func main() {
	config.SetDefault("etcd", "12345")
	fmt.Println(config.GetString("etcd"))
	fmt.Println(etcdVar)
}
```

kshlm:
#111 is merged now. Closing the issue.

PR #111, opened 2016-07-25
Fix initialization of myUUIDFile

prashanthpai:
This closes issue #110

### DIFF
--- a/context/myuuid.go
+++ b/context/myuuid.go
@@ -11,13 +11,14 @@ import (
 )
 
 var (
-	myUUIDFile = path.Join(config.GetString("localstatedir"), "uuid")
+	myUUIDFile string
 )
 
 // InitMyUUID initializes MyUUID by reading the `<config.LocalStateDir>/uuid` file.
 // If the file is not present it generates a new UUID and saves it to the file.
 // If the file is not accessible. initMyUUID panics.
 func InitMyUUID() uuid.UUID {
+	myUUIDFile = path.Join(config.GetString("localstatedir"), "uuid")
 	ubytes, err := ioutil.ReadFile(myUUIDFile)
 	if err != nil {
 		switch {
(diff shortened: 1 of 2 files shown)


context/myuuid.go
```diff
@@ -11,13 +11,14 @@ import (
...
 var (
-	myUUIDFile = path.Join(config.GetString("localstatedir"), "uuid")
+	myUUIDFile string
 )
 
 // InitMyUUID initializes MyUUID by reading the `<config.LocalStateDir>/uuid` file.
 // If the file is not present it generates a new UUID and saves it to the file.
 // If the file is not accessible. initMyUUID panics.
 func InitMyUUID() uuid.UUID {
+	myUUIDFile = path.Join(config.GetString("localstatedir"), "uuid")
```
kshlm: I guess just this should be enough to fix the issue.

Global variables get initialized during package initialization when a Go binary is loaded. This happens before any functions are called. The previous config package we had initialized LocalStateDir during package initialization, as the `flag` package parses the command line on init (be default).

With the change to Viper, the config variables are now only ready for use after `initConfig()` is called. `InitMyUUID` gets called after `initConfig()` in `main()`. So just having this one change should be enough to the UUID file working properly.

prashanthpai: Correct. Other changes are cosmetic changes: One is to rename method 'dumpConfig', other is usage of `SetDefault` which takes effect only if no command line arg or conf is provided and no default is provided. But we can provide default value during `parseFlags` itself which only called during runtime.

kshlm: About `SetDefaults`, I had a comment explaining why I had it. Not all configuration options might have a flag exposed. `SetDefault` could be used for setting the defaults for these config options. But I agree it's useless right now.

kshlm:
ok to test

PR #116, opened 2016-08-03
makefile: Reduce gometalinter parallel jobs to 4

kshlm:
The default of 12 is too much and can hang your system.


kshlm:
ok to test

PR #119, opened 2016-08-08
config: Update default configs ...

kshlm:
and introduce workdir ...
and reintroduce rpcport.

### DIFF
--- a/config.go
+++ b/config.go
@@ -27,7 +27,10 @@ var (
 
 // parseFlags sets up the flags and parses them, this needs to be called before any other operation
 func parseFlags() {
-	flag.String("localstatedir", "", "Directory to store local state information. Defaults to current working directory.")
+	flag.String("workdir", "", "Working directory for GlusterD. (default: current directory)")
+	flag.String("localstatedir", "", "Directory to store local state information. (default: workdir)")
+	flag.String("rundir", "", "Directory to store runtime data. (default: workdir/run)")
+	flag.String("logdir", "", "Directory to store logs. (default: workdir/log)")
 	flag.String("config", "", "Configuration file for GlusterD. By default looks for glusterd.(yaml|toml|json) in /etc/glusterd and current working directory.")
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -27,7 +28,10 @@ var (
 
 // parseFlags sets up the flags and parses them, this needs to be called before any other operation
 func parseFlags() {
-	flag.String("localstatedir", "", "Directory to store local state information. Defaults to current working directory.")
+	flag.String("workdir", "", "Working directory for GlusterD. (default: current directory)")
```
prashanthpai: What is this distinction between workdir and localstatedir ?

config.go
```diff
@@ -13,6 +13,7 @@ const (
 	defaultLogLevel    = "debug"
 	defaultRestAddress = ":24007"
 	defaultRpcAddress  = ":24008"
+	defaultRpcPort     = 24008
```
prashanthpai: You could derive this from `defaultRpcAddress` or the other way too.

config.go
kshlm: `workdir` is the default directory for the process. It will be the directory from which relative paths are rooted. Setting a `workdir` helps avoid clutter in the current directory caused by (possible) artifacts that get generated.
For eg. the way we launch etcd right now, we don't set a `data-dir` for etcd. etcd creates it's data-dir in its working directory. As a child of GD2, etcds working directory is GD2 working directory. This leads to a `default.etcd` directory being created in whichever directory you start GD2 from.

`localstatedir` is the place where information managed and generated by GD2 is stored. This is the `/var/lib/glusterd` directory.

prashanthpai: Ok

config.go
kshlm: For now, I want to keep this. We can figure out how exactly for fix this later.

kshlm:
@atinmu @ppai I'll merge if you have don't have any problems with this.

prashanthpai:
@kshlm  I'm ok with merging this.

kshlm:
Thanks @prashanthpai.

PR #123, opened 2016-09-19
Convert remaining volume commands to transaction framework

kshlm:


atinmu:
@kshlm changes look good, shall I go ahead and merge it?

PR #124, opened 2016-09-20
Convert volume start and stop to transaction framework

kshlm:


commands/volumes/volume-stop.go
```diff
@@ -4,13 +4,61 @@ import (
...
 	"github.com/gluster/glusterd2/rest"
+	"github.com/gluster/glusterd2/transaction"
 	"github.com/gluster/glusterd2/volume"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/gorilla/mux"
+	"github.com/pborman/uuid"
 )
 
+func stopBricks(c transaction.TxnCtx) error {
```
prashanthpai: This is very similar to undoStartBricks. This can be refactored later.

commands/volumes/volume-start.go
```diff
@@ -4,13 +4,94 @@ import (
...
+	}
+	return nil
+}
+
+func registerVolStartStepFuncs() {
+	transaction.RegisterStepFunc(startBricks, "vol-start.Commit")
+	transaction.RegisterStepFunc(undoStartBricks, "vol-start.Undo")
+}
+
+func nodesForVolStart(vol *volume.Volinfo) []uuid.UUID {
```
prashanthpai: This can be generalized across multiple operations.

commands/volumes/volume-stop.go
```diff
@@ -4,13 +4,61 @@ import (
...
+				"brick":  brick.Hostname + ":" + brick.Path,
+			}).Info("would stop brick")
+			//TODO: Stop running brick processes once the daemon management package is ready
+		}
+	}
+	return nil
+}
+
+func registerVolStopStepFuncs() {
+	transaction.RegisterStepFunc(stopBricks, "vol-stop.Commit")
```
prashanthpai: This has no rollback. I think it make sense. But If one of the nodes in unreachable ?

kshlm: Nodes will be checked if they are online before the transaction steps are executed. If a node does go offline during the transaction, it should be okay when it come back online later, as the status would be changed in the store.

commands/volumes/volume-start.go
kshlm: Yup. On it.

PR #128, opened 2016-09-21
Switch to gRPC as the RPC framework of choice.

kshlm:
This is a WIP. Don't merge yet.


atinmu:
Looks good to me, @prashanthpai - do you want to go for one round of review here?

kshlm:
Don't merge this! Still a WIP.

kshlm:
@prashanthpai Please review this when you have time.

scripts/check-protoc.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
prashanthpai: I pulled this PR locally and ran `make` which complained about unavailability of protoc 

~~~
./scripts/check-protoc.sh: line 25: protoc: command not found
Protobuf compiler (protoc)  or later is missing on this system.
Install protoc  using the preferred method for your system.
Refer to https://developers.google.com/protocol-buffers/ if Protobuf  is not available in the system repositories.
Makefile:17: recipe for target 'check-protoc' failed
make: *** [check-protoc] Error 1
~~~

Did `sudo dnf install /usr/bin/protoc` and got the packages. But the above error didn't go away :(

~~~
[ppai@gd2-1 glusterd2]$ which protoc
/usr/bin/protoc
~~~

commands/peers/peer-rpc-clnt.go
```diff
@@ -0,0 +1,120 @@
...
+	args := &PeerAddReq{Name: p.Name, Addresses: p.Addresses}
+
+	rsp := new(PeerAddResp)
+	remoteAddress := fmt.Sprintf("%s:%s", p.Name, config.GetString("rpcport"))
+	rpcConn, e := grpc.Dial(remoteAddress, grpc.WithInsecure())
+	if e != nil {
+		log.WithField("error", e).Error("net.Dial() call failed")
+		opRet = -1
+		opError = e.Error()
+		rsp.OpRet = opRet
```
kshlm: Yup. I need to fix this. This file was moved from the previous location in `rpc/client` with just changes required to compile.

commands/peers/peer-rpc-svc.go
```diff
@@ -7,63 +7,75 @@ import (
...
 )
 
 type PeerService int
 
+func init() {
+	server.Register(new(PeerService))
+}
+
+func (p *PeerService) RegisterService(s *grpc.Server) {
+	RegisterPeerServiceServer(s, p)
```
kshlm: Yes. This expected with gRPC.

scripts/check-protoc.sh
kshlm: You need proto3. The error message should have explained it better, but because I didn't use the proper version variable you got a useless error.

Proto3 is currently unavailable on Fedora. I've been trying to get a [copr repository][1] for it, but I'm yet to get it to build correctly. You'll need to install proto3 manually for the time being.

Proto3 is only required to get the `protoc` compiler, which is only required to generate the relevant Go source from the `.proto`s. It's not actually required for building GD2. I'll modify the `Makefile` to reflect this.

You can run `make install` or `make glusterd2` to just build, without checking.

[1]: https://copr.fedorainfracloud.org/coprs/kshlm/protobuf/

kshlm:
@prashanthpai Thanks. I'll merge now.

PR #136, opened 2016-09-26
Daemon framework

prashanthpai:
@kshlm Posted for some early feedback.

### DIFF
--- a/brick/brick.go
+++ b/brick/brick.go
@@ -10,7 +10,7 @@ import (
 
 	"github.com/gluster/glusterd2/gdctx"
 	"github.com/gluster/glusterd2/utils"
-	"github.com/gluster/glusterd2/volume"
+	"github.com/pborman/uuid"
 
 	config "github.com/spf13/viper"
 )
@@ -20,6 +20,14 @@ const (
 	glusterfsd = "/usr/sbin/glusterfsd"
 )
 
+// Brickinfo represents the information of a brick
+// TODO: Move this into Brick struct ?
+type Brickinfo struct {
+	Hostname string
+	Path     string
+	ID       uuid.UUID
+}
+
 type Brick struct {
 	// Externally consumable using methods of Daemon interface
 	binarypath     string
@@ -28,8 +36,8 @@ type Brick struct {
 	pidfilepath    string
 
 	// For internal use
-	brickinfo volume.Brickinfo
-	volinfo   *volume.Volinfo
+	brickinfo Brickinfo
+	volName   string // Introduce this in Brickinfo itself ?
 	port      int
 }
 
@@ -48,7 +56,7 @@ func (b *Brick) Args() string {
 
 	brickPathWithoutSlashes := strings.Trim(strings.Replace(b.brickinfo.Path, "/", "-", -1), "-")
 	logFile := path.Join(config.GetString("logdir"), "glusterfs", "bricks", fmt.Sprintf("%s.log", brickPathWithoutSlashes))
-	volFileId := fmt.Sprintf("%s.%s.%s", b.volinfo.Name, b.brickinfo.Hostname, brickPathWithoutSlashes)
+	volFileId := fmt.Sprintf("%s.%s.%s", b.volName, b.brickinfo.Hostname, brickPathWithoutSlashes)
 	//TODO: For now, getting next available port. Use portmap ?
 	brickPort := strconv.Itoa(GetNextAvailableFreePort())
 
@@ -61,7 +69,7 @@ func (b *Brick) Args() string {
 	buffer.WriteString(fmt.Sprintf(" --brick-port %s", brickPort))
 	buffer.WriteString(fmt.Sprintf(" -l %s", logFile))
 	buffer.WriteString(fmt.Sprintf(" --xlator-option *-posix.glusterd-uuid=%s", gdctx.MyUUID))
-	buffer.WriteString(fmt.Sprintf(" --xlator-option %s-server.listen-port=%s", b.volinfo.Name, brickPort))
+	buffer.WriteString(fmt.Sprintf(" --xlator-option %s-server.listen-port=%s", b.volName, brickPort))
 
 	b.args = buffer.String()
 	return b.args
@@ -79,7 +87,7 @@ func (b *Brick) SocketFile() string {
 	// Example: /var/lib/glusterd/vols/<vol-name>/run/<host-name>-<brick-path>
 	brickPathWithoutSlashes := strings.Trim(strings.Replace(b.brickinfo.Path, "/", "-", -1), "-")
 	fakeSockFileName := fmt.Sprintf("%s-%s", b.brickinfo.Hostname, brickPathWithoutSlashes)
-	volumedir := utils.GetVolumeDir(b.volinfo.Name)
+	volumedir := utils.GetVolumeDir(b.volName)
 	fakeSockFilePath := path.Join(volumedir, "run", fakeSockFileName)
 
 	// Then md5sum of the above path shall be the name of socket file.
@@ -106,7 +114,7 @@ func (b *Brick) PidFile() string {
 }
 
 // Returns a new instance of Brick type which implements the Daemon interface
-func NewDaemon(vinfo *volume.Volinfo, binfo volume.Brickinfo) (*Brick, error) {
-	brickObject := &Brick{binarypath: glusterfsd, brickinfo: binfo, volinfo: vinfo}
+func NewDaemon(volName string, binfo Brickinfo) (*Brick, error) {
+	brickObject := &Brick{binarypath: glusterfsd, brickinfo: binfo, volName: volName}
 	return brickObject, nil
 }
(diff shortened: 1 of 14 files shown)


brick/brick.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (b *Brick) PidFile() string {
+
+	if b.pidfilepath != "" {
+		return b.pidfilepath
+	}
+
+	// TODO: A more "standard location" to place pidfiles in is /var/run/
+	// For now, following whatever glusterd1 does
+	volumedir := utils.GetVolumeDir(b.volinfo.Name)
```
kshlm: Use `config.GetString("rundir")`. This will be the standard location.

brick/brick.go
```diff
@@ -0,0 +1,88 @@
...
+	brickPathWithoutSlashes := strings.Replace(b.brickinfo.Path, "/", "-", -1)
+	pidfilename := fmt.Sprintf("%s-%s.pid", b.brickinfo.Hostname, brickPathWithoutSlashes)
+	b.pidfilepath = path.Join(volumedir, "run", pidfilename)
+
+	// It is assumed that the responsibility of this function is to only
+	// generate and return the pid file path and not to create it ?
```
kshlm: Yup. The process launched will create the pidfile at this given path.

brick/brick.go
```diff
@@ -0,0 +1,88 @@
...
+
+type Brick struct {
+	// Externally consumable using methods of Daemon interface
+	binarypath     string
+	args           string
+	socketfilepath string
+	pidfilepath    string
+
+	// For internal use
+	brickinfo volume.Brickinfo
```
kshlm: You could also just implement these interface functions directly on a `volume.Brickinfo`. If you have a particular reason to do it this way, let me know.

daemon/daemon.go
```diff
@@ -0,0 +1,89 @@
...
+	if err == nil {
+		// Check if process is running
+		_, err := utils.GetProcess(pid)
+		if err == nil {
+			return errors.ErrProcessAlreadyRunning
+		}
+	}
+
+	// TODO: Modify pidfile under some sort of lock
+	err = utils.WritePidToFile(cmd.Process.Pid, d.PidFile())
```
kshlm: I don't think this is required. The process that is started should be writing/updating the pidfile. Unless we plan to track processes ourselves, this isn't needed.

daemon/daemon.go
```diff
@@ -0,0 +1,89 @@
+package daemon
```
kshlm: As a whole this package looks good. 👍

utils/utils.go
```diff
@@ -24,9 +25,10 @@ import (
 )
 
 const (
-	testXattr     = "trusted.glusterfs.test"
-	volumeIDXattr = "trusted.glusterfs.volume-id"
-	gfidXattr     = "trusted.gfid"
+	testXattr       = "trusted.glusterfs.test"
+	volumeIDXattr   = "trusted.glusterfs.volume-id"
+	gfidXattr       = "trusted.gfid"
+	GlusterdSockDir = "/var/run/gluster"
```
kshlm: Don't hard code paths. For the rundir use `config.GetString("rundir")`.

utils/utils.go
```diff
@@ -367,3 +369,56 @@ func GetLocalIP() (string, error) {
 func GetFuncName(fn interface{}) string {
 	return runtime.FuncForPC(reflect.ValueOf(fn).Pointer()).Name()
 }
+
+func WritePidToFile(pid int, path string) error {
```
kshlm: I'd like to keep utils within the packages that use them, until they become usefull elsewhere. Then they can be moved into the common utils package. To better organize code, add these in a `utils.go` file in the `daemon` package.

volume/struct.go
```diff
@@ -220,6 +220,10 @@ func (v *Volinfo) Nodes() []uuid.UUID {
 	var nodes []uuid.UUID
 
 	for _, b := range v.Bricks {
+		// This is a BUG. nodes[] should be a set (like in python)
+		// to avoid duplicates. Although not in production, there
+		// can be multiple bricks on the same machine and this will
+		// result nodes being added multiple times to the list.
```
kshlm: I didn't get this. Could you explain it a little more? A reply to this comment should be sufficient.

utils/volumeutils.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"fmt"
+	"path"
+	"strings"
+
+	config "github.com/spf13/viper"
+)
+
+func GetVolumeDir(volumeName string) string {
```
kshlm: I'm wondering if these two functions could be associated with the `Volinfo` and `Brickinfo` types themselves.

utils/volumeutils.go
```diff
@@ -0,0 +1,19 @@
...
+	return path.Join(config.GetString("localstatedir"), "vols", volumeName)
+}
+
+func GetBrickVolFilePath(volumeName string, brickHostName string, brickPath string) string {
+	volumeDir := GetVolumeDir(volumeName)
+	brickPathWithoutSlashes := strings.Replace(brickPath, "/", "-", -1)
+	return fmt.Sprintf("%s/%s.%s.%s.vol", volumeDir, volumeName, brickHostName, brickPathWithoutSlashes)
```
kshlm: Use `path.Join`

kshlm:
add to whitelist

kshlm:
ok to test

volume/struct.go
prashanthpai: For example: For a plain distribute or plain replica(2) volume, if my bricks are `192.168.56.1:/export/brick1` and `192.168.56.1:/export/brick2`, the same node UUID will appear twice in the `nodes` list as `b.ID` will be the same for both bricks.

This results in the `startBricks` being run twice on the same node. Each run of `startBricks` will loop over list of bricks and start them.

Is my understanding correct ?

brick/brick.go
prashanthpai: I intended to have this separation as `volume.Brickinfo` is part of volume package and I'd prefer each daemon to be in it's own package - hence the `brick` package. We could move contents of `BrickInfo` here and update wherever it's used but again, `Brickinfo` has only three fields and I don't see a reason for `volume` package to import `brick` package in it's code. I'd prefer them to be more modular with less dependency.

IMO, a general rule of thumb good to follow is any daemon package should only import "generic" packages like utils, context, config and errors. I'd prefer to remove the import and use of `volume` package here in future.

brick/brick.go
prashanthpai: Will do

daemon/daemon.go
prashanthpai: Will remove

utils/utils.go
prashanthpai: I did see some etcd code using similar code for process management. So, decided to move it utils so that any package can use it.

utils/volumeutils.go
prashanthpai: Will do

utils/volumeutils.go
prashanthpai: Good idea. On the long term, I'd prefer `brick` package to not depend on `volume` package.

utils/utils.go
kshlm: Ideally the `etcdmdmt` package will also the daemon package to manage the etcd daemon. But we are planning to move to embedding etcd. In any case, this will not provide extra utility being here.

volume/struct.go
kshlm: You are right. Open a issue for this, better way to track it.

brick/brick.go
kshlm: I'd prefer not to have separate daemon package implementing a type satisfying the `daemon.Daemon` interface for every daemon. That just adds more packages to manage.

As a guideline to follow, every daemon should have its own package, which implements a whole lot more functionality related to the daemon, as well as implementing the interface. This keeps information related to the daemon in one place, not spread around, making it easier to maintain.

If implementing the interface within the `volume` package doesn't sound right, then keep this package and move `Brickinfo` here. This would adhere to the above guideline, as a brick is a standalone daemon unit.

daemon/utils.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	// From https://golang.org/pkg/os/#FindProcess:
+	// On Unix systems, FindProcess always succeeds and returns a Process
+	// for the given pid, regardless of whether the process exists.
+	//
+	// Sending signal 0 can be used to check for the existence of a process ID
+	// Refer `man 2 kill`
+	err = process.Signal(syscall.Signal(0))
+	if err == nil {
```
kshlm: Just to be consistent with other error checks, have this as `if err != nil`

prashanthpai: Done.

prashanthpai:
@kshlm Yes, I'll do that.

Please note that, in it's current form, the framework only brings up the process. It doesn't set up pipes for communication, set process mask etc among other things.

kshlm:
@prashanthpai That's fine. Document whatever exists. When new stuff comes in you document it then.
I mainly want the external interfaces documented.

PR #138, opened 2016-09-29
Document daemon interface

prashanthpai:

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -10,9 +10,12 @@ import (
 	log "github.com/Sirupsen/logrus"
 )
 
+// Daemon interface should be implemented by individual daemons which wants
+// glusterd to manage the lifecycle of the daemon.
 type Daemon interface {
 
-	// Name of the daemon. This will be primarily used for logging.
+	// Name should return the name of the daemon. This will be primarily
+	// used for logging.
 	Name() string
 
 	// Absolute path to the binary/executable of the daemon.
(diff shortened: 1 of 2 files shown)


daemon/daemon.go
```diff
@@ -11,14 +11,44 @@ import (
 )
 
 type Daemon interface {
```
kshlm: `Daemon` the type needs documentation as well.

daemon/daemon.go
```diff
@@ -11,14 +11,44 @@ import (
...
-	Name() string       // Name of daemon
-	Path() string       // absolute path to the binary
-	Args() string       // args to pass to binary during start
-	SocketFile() string // path to socket file to connect to
-	PidFile() string    // path to pidfile
+
+	// Name of the daemon. This will be primarily used for logging.
```
kshlm: Godoc documentation generally follows a '<Identifier> does something' format. So this would be documented as `// Name should return the name of the daemon. This will be primarily used for logging.` Change all the other comments to follow this as well.

`golint` should have caught this, but I've just realised that since it's being run as a part of `gometalinter` with warnings silent, you are not being warned. I'll fix that later. For now, you can run `golint  <file>` to check if the file is good.

prashanthpai: Thanks. Done.

daemon/daemon.go
```diff
@@ -10,9 +10,12 @@ import (
...
+// Daemon interface should be implemented by individual daemons which wants
+// glusterd to manage the lifecycle of the daemon.
 type Daemon interface {
 
-	// Name of the daemon. This will be primarily used for logging.
+	// Name should return the name of the daemon. This will be primarily
+	// used for logging.
 	Name() string
 
 	// Absolute path to the binary/executable of the daemon.
```
kshlm: You've missed the rest of the interface functions.

prashanthpai: Oops! I thought golint catches these too. Will fix.

PR #143, opened 2016-09-30
Don't let child process become zombie

prashanthpai:
glusterfsd, when spawned will exit because:
* Directories in logfile path and pidfile path don't exist
* It tries to use portmap and communicate with glusterd to get volfile.
  This isn't implemented yet.

When glusterfsd dies immediately after spawn, it'll become zombie if
exit status of the process isn't read by parent.

This change also moves the code that reads pid file to place where it
really should be. Also checks if process is really running before it
returns success to the caller.

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -77,10 +77,8 @@ func Start(d Daemon, wait bool) error {
 	} else {
 		// If the process exits at some point later, do read it's
 		// exit status. This should not let it be a zombie.
-		done := make(chan error, 1)
 		go func() {
-			done <- cmd.Wait()
-			err := <-done
+			err := cmd.Wait()
 			log.WithFields(log.Fields{
 				"name":   d.Name(),
 				"pid":    cmd.Process.Pid,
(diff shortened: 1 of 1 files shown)


daemon/daemon.go
```diff
@@ -64,16 +74,27 @@ func Start(d Daemon, wait bool) error {
 		// Wait for the process to exit
 		err = cmd.Wait()
 		return err
+	} else {
+		// If the process exits at some point later, do read it's
+		// exit status. This should not let it be a zombie.
+		done := make(chan error, 1)
+		go func() {
+			done <- cmd.Wait()
```
kshlm: You don't need a channel here. You can just directly assign to `err`. As a thumb rule, you should be using channels when communicating across channels most of the time.

daemon/daemon.go
```diff
@@ -64,16 +74,27 @@ func Start(d Daemon, wait bool) error {
...
-	if err == nil {
-		// Check if process is running
-		_, err := GetProcess(pid)
-		if err == nil {
-			return errors.ErrProcessAlreadyRunning
-		}
+	// Check if the daemon is running
+	// TODO: Need some form of waiting period or timeout here before
+	// returning success. The process may die shortly after spawn.
+	_, err = GetProcess(pid)
```
kshlm: `pid` would be uninitialized here. I think you should be using  `cmd.Process.Pid`

daemon/daemon.go
prashanthpai: Okay.

prashanthpai:
@kshlm Have addressed review comments and updated the PR.

PR #154, opened 2016-10-27
Simplify prefix

prashanthpai:


store/store.go
```diff
@@ -77,27 +53,17 @@ func (s *GDStore) InitPrefix(p string) error {
...
-
 			if e := s.Put(p, nil, &store.WriteOptions{IsDir: true}); e != nil {
-				log.WithFields(log.Fields{
-					"preifx": p,
-					"error":  e,
-				}).Error("error initing prefix")
-
 				return e
+			} else {
+				log.WithField("prefix", p).Debug("Prefix not found. Created prefix.")
```
kshlm: Split this log. Log "Prefix not found" when you enter the case, and "Created prefix" only on success.

gdctx/global.go
```diff
@@ -71,7 +71,12 @@ func doInit() {
 
 	Rest = rest.New()
 
-	initStore()
+	if Restart == true {
```
kshlm: Couldn't this just be `InitStore(!Restart)`. Add a comment clarifying what's being done above this, and everyones happy.

gdctx/store.go
```diff
@@ -15,14 +19,25 @@ func RegisterStorePrefix(prefix string) {
 }
 
 // InitStore is to initialize the store
-func initStore() {
-	Store = store.New(Restart)
+func InitStore(initPrefix bool) {
+	Store = store.New()
+
+	if initPrefix == true {
```
kshlm: The linter should have complained about this. You don't need to check for `boolval == true` for an `if` statement. Keep it `if initPrefix {`.

kshlm: On why this wasn't caught, `make verify` uses gometalinter, which I've set to ignore warnings for everything including golint. I'll probably change it some time later.

gdctx/store.go
```diff
@@ -15,14 +19,25 @@ func RegisterStorePrefix(prefix string) {
 }
 
 // InitStore is to initialize the store
-func initStore() {
-	Store = store.New(Restart)
+func InitStore(initPrefix bool) {
+	Store = store.New()
```
kshlm: Begin variable names with lowercase, unless you are exporting the variable. Infact, capitalize names of variables, types, functions etc. only if you want to export them. If you are only using these within this package, keep them all in lowercase and unexported.

prashanthpai: In this particular case, `Store` variable is exported and consumed by other packages.

kshlm: Cool! It's fine then.

PR #157, opened 2016-10-27
Fix golint warnings

prashanthpai:
Barring transaction subpackage which contains mainly generated code,
other subpackages have had their golint warnings fixed.

We would want to ensure that all new code adheres to golint.

Signed-off-by: Prashanth Pai ppai@redhat.com

### DIFF
--- a/brick/brick.go
+++ b/brick/brick.go
@@ -28,6 +28,7 @@ type Brickinfo struct {
 	ID       uuid.UUID
 }
 
+// Brick type represents information about the brick daemon
 type Brick struct {
 	// Externally consumable using methods of Daemon interface
 	binarypath     string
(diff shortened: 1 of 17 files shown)


purpleidea:
> We would want to ensure that all new code adheres to golint.

You probably want to make sure that all code adheres to `go vet`. Golint is *usually* right, but not always!

kshlm:
@purpleidea, we're using gometalinter, which runs a bunch of stuff, including golint and go vet.

kshlm:
@purpleidea, we're using gometalinter, which runs a bunch of stuff, including golint and go vet.

kshlm:
@purpleidea, we're using gometalinter, which runs a bunch of stuff, including golint and go vet.

prashanthpai:
@kshlm Rebased to master. Consider merging this.

PR #148, opened 2016-10-18
Embed etcd

prashanthpai:

PR #159, opened 2016-11-02
Etcd client refactor

prashanthpai:


prashanthpai:
Was initially part of PR #148. But this really has no dependency on that, so keeping things separate here.

prashanthpai:
Getting this merged will allow me to start working on moving our internal client to v3.

etcdmgmt/client.go
```diff
@@ -0,0 +1,35 @@
...
+	etcd "github.com/coreos/etcd/client"
+)
+
+var (
+	etcdClient etcd.Client
+)
+
+// InitEtcdClient will initialize etcd client. This instance of the client
+// should only be used to maintain/modify cluster membership. For storing
+// key-values in etcd store, one should use libkv instead.
```
kshlm: This is because libkv doesn't work with restarts right?

PR #166, opened 2016-11-16
Store etcd configs across glusterd restarts

prashanthpai:

### DIFF
--- a/commands/peers/peer-rpc-svc.go
+++ b/commands/peers/peer-rpc-svc.go
@@ -80,7 +80,7 @@ func (p *PeerService) ExportAndStoreETCDConfig(nc netctx.Context, c *EtcdConfigR
 	var opRet int32
 	var opError string
 
-	newEtcdConfig, err := etcdmgmt.GetNewEtcdConfig()
+	newEtcdConfig, err := etcdmgmt.GetEtcdConfig(false)
 	if err != nil {
 		opRet = -1
 		opError = fmt.Sprintf("Could not fetch etcd configuration.")
@@ -145,7 +145,15 @@ func (p *PeerService) ExportAndStoreETCDConfig(nc netctx.Context, c *EtcdConfigR
 		peer.AddSelfDetails()
 	} else {
 		// Store the etcd config in a file for use during restarts.
-		err = etcdmgmt.StoreEtcdConfig(newEtcdConfig)
+
+		emcfg := &etcdmgmt.EtcdMinimalConfig{
+			InitialCluster: newEtcdConfig.InitialCluster,
+			ClusterState:   newEtcdConfig.ClusterState,
+			Name:           newEtcdConfig.Name,
+			Dir:            newEtcdConfig.Dir,
+		}
+
+		err = etcdmgmt.StoreEtcdConfig(emcfg)
 		if err != nil {
 			opRet = -1
 			opError = fmt.Sprintf("Error storing etcd configuration.")
--- a/commands/peers/peer-rpc-svc.go
+++ b/commands/peers/peer-rpc-svc.go
@@ -80,7 +80,7 @@ func (p *PeerService) ExportAndStoreETCDConfig(nc netctx.Context, c *EtcdConfigR
 	var opRet int32
 	var opError string
 
-	newEtcdConfig, err := etcdmgmt.GetNewEtcdConfig(false)
+	newEtcdConfig, err := etcdmgmt.GetNewEtcdConfig()
 	if err != nil {
 		opRet = -1
 		opError = fmt.Sprintf("Could not fetch etcd configuration.")
--- a/commands/peers/peer-rpc-svc.go
+++ b/commands/peers/peer-rpc-svc.go
@@ -80,7 +80,7 @@ func (p *PeerService) ExportAndStoreETCDConfig(nc netctx.Context, c *EtcdConfigR
 	var opRet int32
 	var opError string
 
-	newEtcdConfig, err := etcdmgmt.GetNewEtcdConfig()
+	newEtcdConfig, err := etcdmgmt.GetNewEtcdConfig(false)
 	if err != nil {
 		opRet = -1
 		opError = fmt.Sprintf("Could not fetch etcd configuration.")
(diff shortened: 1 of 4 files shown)


commands/peers/peer-rpc-svc.go
```diff
@@ -143,6 +143,14 @@ func (p *PeerService) ExportAndStoreETCDConfig(nc netctx.Context, c *EtcdConfigR
 		// standalone cluster.
 		gdctx.InitStore(true)
```
kshlm: Shouldn't you be cleaning up the store config here?

etcdmgmt/server.go
```diff
@@ -20,9 +20,7 @@ var etcdInstance = struct {
 
 // GetNewEtcdConfig will return reference to embed.Config object. This
 // is to be passed to embed.StartEtcd() function.
-func GetNewEtcdConfig() (*embed.Config, error) {
-
-	//TODO: Read local stored config to handle glusterd2 restarts
+func GetNewEtcdConfig(readConf bool) (*embed.Config, error) {
```
kshlm: Let `GetNewEtcdConfig` do just that, get a new etcd config. Let main handle reading or generating a new config.

Also move this to `utils.go`.

main.go
```diff
@@ -49,7 +49,7 @@ func main() {
 	gdctx.SetLocalHostIP()
 
 	// Start embedded etcd server
-	etcdConfig, err := etcdmgmt.GetNewEtcdConfig()
+	etcdConfig, err := etcdmgmt.GetNewEtcdConfig(true)
```
kshlm: You should do the reading of the etcd config here. You could do,
```go
config, err := ReadConfig()
if err == ENOENT {
  config, err := NewConfig()
  //handle error
} else {
 panic()
}

File: etcdmgmt/utils.go [hunk 4]
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/coreos/etcd/embed"
+	"github.com/ghodss/yaml"
+	config "github.com/spf13/viper"
+)
+
+// EtcdConfigFile is path to etcd config file
+var EtcdConfigFile string
+
+// EtcdMinimalConfig represents essential etcd parameters.
+type EtcdMinimalConfig struct {
```
kshlm: Is there a reason to leave out things like client-urls and peer-urls here?

etcdmgmt/server.go
prashanthpai: Right, that'll be better. Will do.

etcdmgmt/utils.go
prashanthpai: They are already set in `GetNewEtcdConfig` and also inside `embed.NewConfig()`. Also, the things left out remain same for standalone mode and also when glusterd2 is part of the cluster. Only initial cluster, cluster state and name is required to reconfigure etcd to make it join a cluster.

`EtcdMinimalConfig` is a subset of `EtcdConfigReq`

prashanthpai:
@kshlm Have addressed your comments.

etcdmgmt/utils.go
```diff
@@ -0,0 +1,104 @@
...
+	err = ioutil.WriteFile(EtcdConfigFile, y, 0644)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// ReadEtcdConfig reads etcd configuration info stored in file
+func ReadEtcdConfig() (*EtcdMinimalConfig, error) {
```
kshlm: Didn't notice this the last time. You should have this return `embed.Config` as well. To properly complement `GetNewEtcdConfig` and `StoreEtcdConfig`.

main.go
```diff
@@ -53,6 +53,14 @@ func main() {
 	if err != nil {
 		log.WithField("Error", err).Fatal("Could not fetch config options for etcd.")
 	}
+	oldCfg, err := etcdmgmt.ReadEtcdConfig()
```
kshlm: With changes done to `ReadEtcdConfig`, this becomes simpler `if cannot read create new`.

prashanthpai: Oh no. I think you missed the point. I was avoiding code duplication. With whatever you're suggesting, I'll have to be calling `GetNewEtcdConfig` inside `ReadEtcdConfig` and then again in main().

I think it'll be better off if I just rename `GetNewEtcdConfig` to `GetEtcdConfig`. And let the function figure out whether to read it from file, or create new ?

etcdmgmt/utils.go
kshlm: `EtcdMinimalConfig` is just an internal representation just to make it easier for the package to store/restore config. An external user shouldn't need to know about it to make use of `(Get,Store,Read)EtcdConfig`.

If in the future you needed to store more information, and made changes to `EtcdMinimalConfig`, doing it this way would require the user of `StoreEtcdConfig` or `ReadEtcdConfig` to make changes. Why should they do it, when you could just do it for them within these functions, and give/accept their EtcdConfigs?

prashanthpai: At this point in time, `ReadEtcdConfig` is to be only internally consumed by `GetEtcdConfig`. So I should really make it private. So there won't be any external consumers of that.

I'll make StoreEtcdConfig take in `embed.Config`.

To summarize:  only `GetEtcdConfig` and `StoreEtcdConfig` will be externally consumable and they will return and accept `embed.Config` respectively. This should be reasonable, what do you think ?

etcdmgmt/utils.go
```diff
@@ -0,0 +1,115 @@
...
+	log "github.com/Sirupsen/logrus"
+	"github.com/coreos/etcd/embed"
+	"github.com/ghodss/yaml"
+	config "github.com/spf13/viper"
+)
+
+// EtcdConfigFile is path to etcd config file
+var EtcdConfigFile string
+
+type etcdMinimalConfig struct {
```
prashanthpai: This is no longer exposed.

prashanthpai: Also the method `readEtcdConfig` which deals with unexported structure `etcdMinimalConfig` is also no longer public function.

etcdmgmt/utils.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &cfg, nil
+}
+
+func init() {
+	if EtcdConfigFile == "" {
+		EtcdConfigFile = path.Join(config.GetString("localstatedir"), "etcd.yaml")
```
kshlm: Just saw this. 

Does this work? `init` functions are called when packages are loaded, before `main()` is run. But for us `config` is only truly complete after `initConfig` is called from `main`. So this will always only set the defaults, ignoring any custom values set from config files.

(Maybe we could try having an `init` function for the `main` package. I wonder if it would work)

PR #173, opened 2016-11-18
Update glide.lock to point to right libkv commit

prashanthpai:


prashanthpai:
@kshlm This will fix failing builds.

PR #174, opened 2016-11-18
Remove etcd binary as an external dependency

prashanthpai:


kshlm:
Could you also remove etcd in `extras/test-env/setup.sh`?

PR #175, opened 2016-11-18
Set GO15VENDOREXPERIMENT to enable native vendor support

rafikc30:


kshlm:
ok to test

PR #180, opened 2016-11-24
Implement volume status as example of transaction result aggregation

prashanthpai:
**Implement volume status endpoint:**

Querying for the status of volume will result in network calls to only
those nodes that contain bricks that make up the volume.
The transaction initiator node aggregates the results from various
nodes and sends it back to client.

This also serves as an example that the transaction framework need
not do the result aggregation part. The nodes can use transaction
context to store results. The initiator node then can do it's own
implementation (this is largely op-specific) to collate results stored
in transaction context and send the aggregated result to client.

### DIFF
--- a/commands/volumes/volume-status.go
+++ b/commands/volumes/volume-status.go
@@ -60,8 +60,8 @@ func checkStatus(ctx transaction.TxnCtx) error {
 		brickStatuses = append(brickStatuses, fakeStatus)
 	}
 
-	// Store the results in transaction context. This will consumed by the
-	// node that initiated the transaction.
+	// Store the results in transaction context. This will be consumed by
+	// the node that initiated the transaction.
 	ctx.SetNodeResult(gdctx.MyUUID, brickStatusTxnKey, brickStatuses)
 
 	return nil
--- a/commands/volumes/volume-status.go
+++ b/commands/volumes/volume-status.go
@@ -16,6 +16,10 @@ import (
 	"github.com/pborman/uuid"
 )
 
+const (
+	brickStatusTxnKey string = "brickstatuses"
+)
+
 func checkStatus(ctx transaction.TxnCtx) error {
 	var volname string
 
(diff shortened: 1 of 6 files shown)


prashanthpai:
The CI failure is unrelated to this PR.

prashanthpai:
Sample volume status output:

```
[ppai@gd2-1 ~]$ curl -s -X GET http://127.0.0.1:24007/v1/volumes/test/status | python -m json.tool
{
    "Brickstatuses": [
        {
            "Hostname": "192.168.56.25",
            "ID": "e80e78b6-56b0-483a-be20-96c4e299feb2",
            "Online": false,
            "Path": "/export/brick1/data",
            "Pid": 1234
        },
        {
            "Hostname": "192.168.56.25",
            "ID": "e80e78b6-56b0-483a-be20-96c4e299feb2",
            "Online": false,
            "Path": "/export/brick3/data",
            "Pid": 1234
        },
        {
            "Hostname": "192.168.56.26",
            "ID": "312974aa-e8d8-4d2c-a2b5-fd2008e46560",
            "Online": false,
            "Path": "/export/brick2/data",
            "Pid": 1234
        },
        {
            "Hostname": "192.168.56.26",
            "ID": "312974aa-e8d8-4d2c-a2b5-fd2008e46560",
            "Online": false,
            "Path": "/export/brick4/data",
            "Pid": 1234
        }
    ]
}
```

commands/volumes/volume-status.go
```diff
@@ -0,0 +1,142 @@
...
+			ID:       binfo.ID,
+			Online:   false,
+			Pid:      1234,
+		}
+		brickStatuses = append(brickStatuses, fakeStatus)
+	}
+
+	// Store the results in transaction context. This will consumed by the
+	// node that initiated the transaction.
+	key := gdctx.MyUUID.String() + "/" + "brickstatuses"
```
kshlm: It'll be better if we can provide a properly defined method for this. We wouldn't want users to be creating arbitrary keys here. Some thing like `TxnCtx.SetMyResult(key, value)` which would set the value with the key "MyUUID/<key>". And a complementary `TxnCtx.GetResult(uuid, key)`.

prashanthpai:
> Right now, you've done aggregation at the end of the transaction. But if we wanted to aggregate at every step, this wouldn't work would it.

Right. At this point I haven't seen the need for it, yet. However, whenever needed, the results of each step can still be stored in transaction context and have the aggregation function as a step that runs only on the initiator node.

PR #183, opened 2016-11-29
Consolidate peer and etcd membership relationship

### LINKED ISSUE #182: Consolidate peer and etcd membership relationship better
- [x] Use hostname as peer name whenever user hasn't specified a name
- [x] Use peer UUID as etcd member name
- [x] Update grpc and protobuf to `grpc.SupportPackageIsVersion4`
- [x] Regenerate `*pb.go` files
- [x] Directly use generated struct `EtcdConfigReq`
- [x] Make addresses a single field (currently a list/array)
- [x] Remove client field from peer and request object

prashanthpai:
This does the following changes:

* Use hostname as peer name whenever user hasn't specified a name
* Use peer UUID as etcd member name
* Update grpc and protobuf to `grpc.SupportPackageIsVersion4`
* Regenerate `*pb.go` files
* Directly use generated struct `EtcdConfigReq`
* Allow only single address to be set as peer address
* Remove client field from Peer object

Closes #182


prashanthpai:
@kshlm This is ready for review now. Fixes all things mentioned in checklist of issue #182

peer/peer.go
```diff
@@ -7,11 +7,10 @@ import (
...
 // Peer reperesents a GlusterD
 type Peer struct {
-	ID        uuid.UUID `json:"id"`
-	Name      string    `json:"name"`
-	Addresses []string  `json:"addresses"`
-	Client    bool      `json:"client"`
-	MemberID  uint64    `json:"memberID"`
+	ID       uuid.UUID `json:"id"`
+	Name     string    `json:"name"`
+	Address  string    `json:"address"`
```
kshlm: It's okay if the peer add ReST API had just one initial address. But a peer could have multiple addresses. While we don't yet have a way to add newer addresses, we can atleast be ready here for when we have it.

commands/peers/addpeer.go
```diff
@@ -26,112 +26,92 @@ func addPeerHandler(w http.ResponseWriter, r *http.Request) {
...
-				n := memb.Name
-				if memb.ID == p.MemberID {
-					n = newName
-				}
-				conf = append(conf, fmt.Sprintf("%s=%s", n, u))
+	log.WithField("member-id", newMember.ID).Info("Added new member to etcd cluster")
+
+	mlist, e := etcdmgmt.EtcdMemberList()
+	if e != nil {
+		log.WithField("err", e).Error("Failed to list members in etcd cluster")
```
kshlm: The field key "err" is different from the one used above on line 64. Keep it consistent.
Also, when you have an error to be added to the log context, you could use `log.WithError`.

peer/peer.go
prashanthpai: > But a peer could have multiple addresses

True. I was wondering if there was any real use case on how the multiple peer addresses will be used. I could not think of any. Despite backed by central store, the peers are never stateless (contain bricks), so peer addresses cannot be used for failover or load balancing.

prashanthpai: I see. Okay

prashanthpai:
@kshlm Have restored storing multiple addresses per peer.

commands/peers/peer-rpc.proto
```diff
@@ -24,11 +24,9 @@ message PeerDeleteReq {
 }
 
 message EtcdConfigReq {
- string PeerName = 1;
- string Name = 2;
+ string EtcdName = 2;
```
kshlm: We haven't yet finalized these protobuf messaages. So we can still change the tags of message members. Start from 1 again. If we were changing after we had finalized, we'd have to give new tags to any new member, like `EtcdName` here, and retire the old member and tag.

Refer https://developers.google.com/protocol-buffers/docs/proto3#assigning-tags for more info.

PR #190, opened 2016-12-13
Simplify brick volfile generation using template

### LINKED ISSUE #146: Generated volfiles are incomplete
The volfiles that are being generated are not complete. Though the structure being generated is correct, it misses a lot of options that need to be present. This is preventing the bricks from starting.

To fix this for now, we should just concentrate on a simple brick graph with just server, io-threads, locks and posix xlators.

prashanthpai:
Closes #146


kshlm:
Just noting here that this volgen is just a temporary thing.

PR #195, opened 2016-12-15
Fully support addresses without ports

prashanthpai:
* Allow addresses without ports during add peer operation
* Normalize peer address before comparison

### DIFF
--- a/utils/peerutils.go
+++ b/utils/peerutils.go
@@ -3,6 +3,7 @@ package utils
 import (
 	"errors"
 	"net"
+	"strings"
 
 	config "github.com/spf13/viper"
 )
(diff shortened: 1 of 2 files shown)


prashanthpai:
@kshlm This fixes the glitch that occurred during the demo earlier for add peer. Need to fix the absence of port elsewhere too (like in address specified during volume create)

utils/peerutils.go
```diff
@@ -3,27 +3,38 @@ package utils
...
 // return an address of the form <ip:port>
 func FormRemotePeerAddress(peeraddress string) (string, error) {
-	var remotePeerAddress string
 
 	host, port, err := net.SplitHostPort(peeraddress)
-	if err != nil || host == "" {
-		return "", errors.New("Invalid peer address")
+	if err != nil {
+		// net.SplitHostPort() returns an error if port is missing.
+		if strings.HasPrefix(err.Error(), "missing port in address") {
```
kshlm: Did you try to find an alternate way to check for error here? There should be a better way to check for this error other than comparing strings.

prashanthpai: I did this after looking at the [source code](https://golang.org/src/net/ipsock.go?s=3288:3354#L99) of `SplitHostPort`. The alternative was to use `strings.Split`

kshlm: Okay then. I'll merge. But this just feels not right.

PR #196, opened 2016-12-16
Avoid race in brick port allocation and use

### LINKED ISSUE #193: Brick port allocation should handle races
Right now, when a volume is comprised of multiple bricks on same node/instance, there's a TOCTOU bug between port allocation and actual biding to the port by the brick process.

prashanthpai:
This fixes the TOCTOU race for brick port by:

* Starting bricks sequentially (per instance).
* Detect when glusterfsd process exits with EADDRINUSE

Closes #193 

### DIFF
--- a/commands/volumes/volume-start.go
+++ b/commands/volumes/volume-start.go
@@ -41,14 +41,14 @@ func startBricks(c transaction.TxnCtx) error {
 			c.Logger().WithFields(log.Fields{
 				"volume": volname,
 				"brick":  b.Hostname + ":" + b.Path,
-			}).Info("would start brick")
+			}).Info("Starting brick")
 
 			brickDaemon, err := brick.NewDaemon(vol.Name, b)
 			if err != nil {
 				return err
 			}
 
-			err = daemon.Start(brickDaemon, false)
+			err = daemon.Start(brickDaemon, true)
 			if err != nil {
 				return err
 			}
--- a/commands/volumes/volume-start.go
+++ b/commands/volumes/volume-start.go
@@ -2,6 +2,9 @@ package volumecommands
 
 import (
 	"net/http"
+	"os/exec"
+	"syscall"
+	"time"
 
 	"github.com/gluster/glusterd2/brick"
 	"github.com/gluster/glusterd2/daemon"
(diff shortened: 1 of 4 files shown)


prashanthpai:
This partially addresses issue #193 but for single instances per node which is how real deployments will be.

prashanthpai:
This also relies on http://review.gluster.org/#/c/16200/ to ensure that glusterfsd process consistently exits with `EADDRINUSE` whenever the port specified in arg is already in use.

commands/volumes/volume-start.go
```diff
@@ -41,15 +44,36 @@ func startBricks(c transaction.TxnCtx) error {
...
 				return err
 			}
 
-			err = daemon.Start(brickDaemon, false)
+			retries := 0
+		RETRY:
+			err = daemon.Start(brickDaemon, true)
 			if err != nil {
+				// Retry iff brick failed to start because of port being in use.
+				if exiterr, ok := err.(*exec.ExitError); ok {
```
kshlm: Way too much indentation here. Makes it hard to read.

commands/volumes/volume-start.go
```diff
@@ -41,15 +44,36 @@ func startBricks(c transaction.TxnCtx) error {
...
+					if status, ok := exiterr.Sys().(syscall.WaitStatus); ok {
```
kshlm: Whenever possible avoid using the `syscall` package. Use `golang.org/x/sys/unix` instead.

commands/volumes/volume-start.go
```diff
@@ -41,15 +44,36 @@ func startBricks(c transaction.TxnCtx) error {
...
+						if status.ExitStatus() == int(syscall.EADDRINUSE) {
+							retries++
+							if retries <= 3 {
+								c.Logger().Info("Brick port unavailable. Retrying...")
```
kshlm: `3` is a magic number here. Better declare it as a `const` someplace. Even better if it is configurable.

commands/volumes/volume-start.go
```diff
@@ -41,15 +44,36 @@ func startBricks(c transaction.TxnCtx) error {
...
+			}).Info("Starting brick")
 
 			brickDaemon, err := brick.NewDaemon(vol.Name, b)
 			if err != nil {
 				return err
 			}
 
-			err = daemon.Start(brickDaemon, false)
+			retries := 0
+		RETRY:
```
kshlm: You can have the same retry behavior using a loop instead of a jump.  Do everything below in a loop and break when you've successfully completed.

commands/volumes/volume-start.go
prashanthpai: Will do. I'm curious to know why though. Quoting go doc:

> Package unix contains an interface to the low-level operating system primitives....
> The primary use of this package is inside other packages that provide a more portable interface to the system, such as "os", "time" and "net". Use those packages rather than this one if you can...

prashanthpai: I tried changing things, and it doesn't work well during type conversions. For example, this conversion fails for me - `status, ok := exiterr.Sys().(unix.WaitStatus)`. The [documentation](https://golang.org/pkg/os/#ProcessState.Sys) states to use syscall.WaitStatus

Only errno constants be can interchangably used between `syscall` and `unix` packages. This is because the definitions of these constants in `unix` package calls `syscall` ones.

commands/volumes/volume-start.go
prashanthpai: Done.

commands/volumes/volume-start.go
prashanthpai: Done.

commands/volumes/volume-start.go
kshlm: Okay. Thanks for looking it up.

kshlm:
retest this please

prashanthpai:
retest

PR #197, opened 2016-12-19
XXX: libmgmt: Rough first hack

purpleidea:
This doesn't compile yet, but it's meant as a rough guide to get the
ball rolling here. The idea is to throw this back and forth between me
and ppai/kshlm. Note that I had to copy the vendor/* contents from mgmt
into vendor/ here.

Don't merge! Patch on top or rebase, and eventually we'll squash this
all down.


purpleidea:
@centos-ci shh...

As discussed...
@kshlm @prashanthpai WIP POC. DO NOT MERGE :)

kshlm:
@purpleidea Getting started on this. I wanted clarify the version of mgmt to use. I guess I should be using the current git HEAD. But you also have a 0.0.7 version available. So which one should I be using?

purpleidea:
@kshlm I'd use git master. I'm online on IRC if you want to poke me or share a hangout to back and forth hack.

purpleidea:
@kshlm any update on this WIP hacking?

kshlm:
@purpleidea I just got to a state where the code compiled. I didn't proceed further as I'm not convinced that we should be using libmgmt as a whole (and because I was away). For now, I don't want to proceed here further. But I still want to get the etcd bits into GD2.

purpleidea:
On Wed, Jan 4, 2017 at 1:18 AM, Kaushal M <notifications@github.com> wrote:
>
> @purpleidea I just got to a state where the code compiled. I didn't proceed further as I'm not convinced that we should be using libmgmt as a whole (and because I was away). For now, I don't want to proceed here further. But I still want to get the etcd bits into GD2.


Using this "libmgmt" bit is what brings in the etcd bits. If you don't
want to use the rest of mgmt you don't have to. Can you post your
branch that compiles?

Cheers

prashanthpai:
I'd prefer to see this co-exist with existing etcd management code and preferably be configurable to choose between `plain old manual etcd` and `libmgmt`. That should give us a good comparison of how the user workflow would change w.r.t peer operations. It would also show how less/more intrusive it would be to change our peer management code, should etcd come up with their own elastic management in future OR if we move to a different distributed clustering system. Currently, for central store purposes, `libkv` gives us that abstraction. But for clustering, we're very much strongly coupled with etcd alone.

purpleidea:
On Fri, Jan 6, 2017 at 8:08 AM, Prashanth Pai <notifications@github.com> wrote:
>
> should etcd come up with their own elastic management in future

Their devs have said that this isn't a goal for etcd core.

kshlm:
I've opened #247 which updates this.

PR #198, opened 2016-12-20
Selectively generate brick volfiles and also include auth credentials

prashanthpai:
* Generate brick volfiles only on nodes that contain those bricks
* Generate trusted username and password in brick volfiles

### DIFF
--- a/volgen/core.go
+++ b/volgen/core.go
@@ -5,6 +5,7 @@ package volgen
 import (
 	"fmt"
 	"os"
+	"path"
 	"strings"
 
 	"github.com/gluster/glusterd2/brick"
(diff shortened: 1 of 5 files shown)


volgen/core.go
```diff
@@ -71,32 +73,43 @@ func GenerateVolfile(vinfo *volume.Volinfo) error {
...
-	// Create volume directory (/var/lib/glusterd/vols/<VOLNAME>)
-	err := os.MkdirAll(vdir, os.ModeDir|os.ModePerm)
-	if err != nil {
-		return err
-	}
-
-	*path = fmt.Sprintf("%s/trusted-%s.tcp-fuse.vol", vdir, vinfo.Name)
-	return err
+func getClientVolFilePath(vinfo *volume.Volinfo) string {
+	return fmt.Sprintf("%s/trusted-%s.tcp-fuse.vol", utils.GetVolumeDir(vinfo.Name), vinfo.Name)
```
kshlm: When generating paths, it's better to use the `path` package. Generate the filename, and the generate the full path by doing a `path.Join(dir,filename)`.

volume/struct.go
```diff
@@ -86,6 +86,12 @@ type Volinfo struct {
 	Bricks   []brick.Brickinfo
 }
 
+// VolAuth represents username and password used by trusted/internal clients
```
prashanthpai: I started out with intent to make this part of `Volinfo` structure itself and then realised that `Volinfo` struct is unmarshalled and sent "as is" to client as json response for `GET` request on the volume. We certainly wouldn't want these auth details to be exposed in client request responses

PR #199, opened 2016-12-23
Generate correct client volfile from templates

prashanthpai:
This PR does the following:

* Generate correct client volfile from templates
* Use int type in Volume structures
* Move brick volfile template to volgen package
* Retire old volgen code


prashanthpai:
With these volfiles, adding `remote-port` entry to client volfile will allow `glusterfs` process to successfully mount.

PR #221, opened 2017-01-16
Name configurable addresses to be more meaningful

prashanthpai:
The name of the configurable options should represent the services
provided or consumer of the port and not the protocol used.

* Renamed 'rpcaddress' to 'peeraddress': Denotes the address used
  for peer to peer communication.
* Renamed 'restaddress' to 'clientaddress': Denotes the address
  used by clients to communicate with glusterd2.

Coincidentally, these are analogous to those provided by etcd.

### DIFF
--- a/config.go
+++ b/config.go
@@ -14,8 +14,8 @@ import (
 
 const (
 	defaultLogLevel          = "debug"
-	defaultRestAddress       = ":24007"
-	defaultRPCAddress        = ":24008"
+	defaultClientAddress     = ":24007"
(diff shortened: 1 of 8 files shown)


prashanthpai:
> In a GlusterFS context a client address would imply something a GlusterFS client would connect to

That's exactly what this PR intends to do. Going forward, the port `24007` will be the port a GlusterFS client will connect to. Hence the name `clientaddress`. The same port will also serve REST requests from the clients

Issue #220, opened 2017-01-13
Use cmux to multiplex multiple protocol over same port


prashanthpai:
The glusterd port `24007` (default unless changed) is used for REST client requests. However, glusterfs client process expects to talk to glusterd on that port using Sun RPC. We could explore if connection multiplexer such as [cmux](https://github.com/soheilhy/cmux) can be used. If yes, then we can server multiple protocols over a single port.

prashanthpai:
A `cmux.Matcher` has been [implemented](https://github.com/prashanthpai/sunrpc/blob/master/cmux_matcher.go) to guess if the connection uses Sun RPC.

prashanthpai:
This issue is addressed in PR #222

PR #222, opened 2017-01-16
Multiplex different protocols over the same port

prashanthpai:
Use cmux, which is a connection multiplexer. cmux matches the
connection when it's accepted. This still doesn't support multiple
protocols over same connection, but that is not what we want.

Closes issue #220


prashanthpai:
This PR is dependent on PR #221

PR #225, opened 2017-01-20
Refactor all the servers/listeners

kshlm:
... and make use of github.com/thejerf/suture to manage them.

More information on `suture` can be found in this [blog post][1]

This PR, refactors the GD2 rest and grpc servers, to make them suture services.
A new `muxsrv` package has been created, creates a cmux server supervisor,
which currently manages the rest server.  A `servers` package implements a
supervisor which manages the above.

The `main` function creates and starts a global/main supervisor, and then adds
other services/supervisors to it.  Right now, this is just the servers
supervisor. The etcdmgmt service will be converted to be a suture service, and
be managed by the main supervisor.

[1]: http://www.jerf.org/iri/post/2930

PR #228, opened 2017-01-25
Add mount rpc programs

prashanthpai:
This PR refactors and simplifies the process of adding new RPC programs
and new procedures. Following RPC programs have been introduced
(procedures are not exhaustive yet) on top of it:

`Glusterfs Handshake`
`Gluster Dump`
`GlusterFS Portmap`

### DIFF
--- a/rpc/sunrpcserver/dump_prog.go
+++ b/rpc/sunrpcserver/dump_prog.go
@@ -1,0 +1,90 @@
+package sunrpcserver
+
+// GfDump is a type for GlusterFS Dump RPC program
+type GfDump genericProgram
+
+func newGfDump() *GfDump {
+	// rpc/rpc-lib/src/xdr-common.h
+	return &GfDump{
+		name:        "GF-DUMP",
+		progNum:     uint32(123451501),
+		progVersion: uint32(1),
+		procedures: []Procedure{
+			Procedure{uint32(1), "Dump"}, // GF_DUMP_DUMP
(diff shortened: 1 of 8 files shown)


rpc/sunrpcserver/dump_prog.go
```diff
@@ -0,0 +1,90 @@
+package sunrpcserver
+
+// GfDump is a type for GlusterFS Dump RPC program
+type GfDump genericProgram
+
+func newGfDump() *GfDump {
+	// rpc/rpc-lib/src/xdr-common.h
+	return &GfDump{
+		name:        "GF-DUMP",
+		progNum:     uint32(123451501),
```
kshlm: Using the random prognum directly will confuse people looking at this code. Please define the numbers as constants and use them instead. Name the constants using Go style camel case (if you had removed the consts to appease the linter)

rpc/sunrpcserver/dump_prog.go
```diff
@@ -0,0 +1,90 @@
...
+		progVersion: uint32(1),
+		procedures: []Procedure{
+			Procedure{uint32(1), "Dump"}, // GF_DUMP_DUMP
```
kshlm: Same with the procnums. Define consts and use them here instead of the number directly.

kshlm:
I'm approving and merging this change. I'll do the fixes for the comments I had on #225

PR #230, opened 2017-01-31
Use ServeCodec in sunrpc server

prashanthpai:
ServeRequest synchronously serves a single request. ServeCodec runs
the server on a single connection. ServeConn blocks, serving the
connection until the client hangs up.


prashanthpai:
While trying to mount a volume, I noticed that the sunrpc server served only the first RPC call sent by the client. With this change, glusterfs client processes should be able to communicate successfully with sunrpc server.

PR #239, opened 2017-02-03
Get notified on sunrpc clients' disconnection

prashanthpai:
sunrpc codec can now notify (via a channel) whenever a sunrpc client
connection is closed. The sunrpc server maintains a list of clients
which is updated on connection and disconnection.
    
But there is no provision yet to know if these sunrpc clients are
glusterfs mount processes or something else.


prashanthpai:
[Here](https://github.com/prashanthpai/sunrpc/commit/755218237e26dd638a5c19bfca85c43099d5a905#diff-212d3376a22bf275eb9770dc5e32ccd1) is the corresponding change in the sunrpc codec which adds the ability to get notified on close.

PR #241, opened 2017-02-07
plugins: Glusterd2 Plugins Infrastructure

### LINKED ISSUE #131: Pluggability
GD2 needs to have plugin system. This is a core requirement.

One of the goals of GD2 is to avoid non GlusterD developers having to modify GD2 directly to get there features working.

Pluggability is needed for,

- Xlators - to add new nodes in the volume graph and to add new volume options
- Commands - to be able to add new ReST endpoints, a new CLI commands and extend existing commands (particularly volume-create)
- Events - to add and send new events

aravindavk:
This is initial code which enables creation of plugins for glusterd2.
Interface design is still in initial stage, do not start working on
any real plugin Yet!

Plugin interface is

    type Gd2Plugin interface{
        SunRpcProcedures() program.Program
        RestRoutes() route.Routes
        RegisterStepFuncs()
    }

Plugin can register sunrpc procedure, REST route and can register step
functions required for Transaction(Transaction APIs are not stable
yet)

Modified `servers/sunrpc/dump_prog.go` as plugin.
`plugins/dump/{init,sunrpc}.go` files shows how to add sunrpc
program into glusterd2's sunrpc server.

Included `hello` plugin, which plugs following REST routes into
glusterd2 REST server.

    GET  /v1/hello
    POST /v1/hello

TODO:
- Store APIs need to be finalized for Plugins
- Daemon management APIs for plugins(Enable/Disable/Start/Stop a
  daemon)
- Simplify the Plugin Interface for ease of use, reduce number of
  imports required inside plugin etc.

Closes #131

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,7 @@ _projects/
 # Ignore build and release directories
 build/
 releases/
+# Generated file
+plugins/plugins.go
+*.etcd
+uuid
--- a/plugins/common.go
+++ b/plugins/common.go
@@ -1,0 +1,12 @@
+package plugins
+
+import (
+	"github.com/gluster/glusterd2/servers/rest/route"
+	"github.com/gluster/glusterd2/servers/sunrpc/program"
+)
+
+type Gd2Plugin interface{
+	SunRpcProcedures() program.Program
+	RestRoutes() route.Routes
+	RegisterStepFuncs()
+}
--- a/scripts/detect-plugins.sh
+++ b/scripts/detect-plugins.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+# This script fetches the list of plugins avalable in plugins directory
+
+IMPORT_PREFIX=github.com/gluster/glusterd2/plugins
+echo "package plugins"
+echo
+
+echo "import ("
+cd "plugins"
+for p in `ls -d */`
+do
+    # ${p%/} is to remove trailing slash
+    echo -e "\t\"${IMPORT_PREFIX}/${p%/}\""
+done
+echo ")"
+
+echo
+echo "var PluginsList = []Gd2Plugin{"
+
+for p in `ls -d */`
+do
+    p=${p%/}
+    # Like &helloplugin.HelloPlugin{}
+    echo -e "\t&${p}plugin.${p^}Plugin{},"
+done
+
+echo "}"
--- a/servers/sunrpc/handshake_prog.go
+++ b/servers/sunrpc/handshake_prog.go
@@ -6,7 +6,8 @@ import (
 	"path"
 
 	"github.com/gluster/glusterd2/utils"
-
+	"github.com/gluster/glusterd2/servers/sunrpc/program"
+	"github.com/gluster/glusterd2/servers/sunrpc/serialize"
 	log "github.com/Sirupsen/logrus"
 )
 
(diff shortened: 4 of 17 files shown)


aravindavk:
@kshlm @prashanthpai @atinmu Please review.

servers/sunrpc/handshake_prog.go
```diff
@@ -6,7 +6,8 @@ import (
 	"path"
 
 	"github.com/gluster/glusterd2/utils"
-
+	"github.com/gluster/glusterd2/servers/sunrpc/program"
```
prashanthpai: Usually, I'd prefer the plugins/rpc programs to not import anything from  `servers/sunrpc`. This will cleanly avoid cyclic imports without having to create subpackages. As a coincidence I [moved](https://github.com/prashanthpai/sunrpc/commit/628e7a95e2aa0d3b8c8723cdd2f3297518b9d341#diff-4518f35e76c810981e52e8c8f90c4fcb) the `Procedure` struct to `sunrpc` package today. Hope this helps. Let me know what you think.

servers/sunrpc/handshake_prog.go
```diff
@@ -6,7 +6,8 @@ import (
...
+	"github.com/gluster/glusterd2/servers/sunrpc/serialize"
```
prashanthpai: I wanted to do this, move serialise out. As mentioned above, serialise can be moved to a package in source's root dir - something like package `dict`.

plugins/common.go
```diff
@@ -0,0 +1,12 @@
+package plugins
+
+import (
+	"github.com/gluster/glusterd2/servers/rest/route"
+	"github.com/gluster/glusterd2/servers/sunrpc/program"
+)
+
+type Gd2Plugin interface{
+	SunRpcProcedures() program.Program
```
aravindavk: Done. I will change this name to SunRpcProgram.

scripts/detect-plugins.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
aravindavk: Valid point, will change this.

.gitignore
```diff
@@ -11,3 +11,7 @@ _projects/
 # Ignore build and release directories
 build/
 releases/
+# Generated file
+plugins/plugins.go
+*.etcd
```
aravindavk: After make, I ran ./build/glusterd2 it generated above files in that directory itself. will remove these ignores

servers/sunrpc/handshake_prog.go
aravindavk: Nice, I will rebase once that change merges here.

servers/sunrpc/handshake_prog.go
aravindavk: Ok, is it not specific to sunrpc or is it xdr specific?

plugins/common.go
```diff
@@ -0,0 +1,12 @@
+package plugins
+
+import (
+	"github.com/gluster/glusterd2/servers/rest/route"
+	"github.com/gluster/glusterd2/servers/sunrpc/program"
+)
+
+type Gd2Plugin interface{
```
aravindavk: Ok, I will copy that instead of move. and change name as something else for example purpose.

aravindavk:
Thanks for the quick comments.

servers/sunrpc/handshake_prog.go
prashanthpai: The dict serialisation is a gluster specific only. The entire dict is serialised before passing it to xdr. What xdr gets is just a opaque data blob. So technically, the dict implementation is not specific to sunrpc or xdr although that's the only place we use it today.

aravindavk:
I couldn't do proper rebase and changes, ended up sending new pull request #294

Closing this issue, Please review #294

PR #244, opened 2017-02-22
Refactor brick package

prashanthpai:
This PR changes/fixes the following:

    * Rename structure 'Brick' (implements Daemon interface) to 'Glusterfsd'
      to explicitly denote that the type only refers to process lifecycle
      of the bricks and not the brick themselves.
    * Separate brick info and brick status by the type of information that
      they contain i.e static or real-time.
    * Fix a TODO. Rename ID field in Brickinfo to NodeID. It was confusing
    * Remove dead code in utils_misc.go
    * Use Node UUID in brick volfile path


prashanthpai:
A PR that implements pmap will follow after this.

PR #246, opened 2017-02-22
Implement internal portmapper registry/service

prashanthpai:
This should allow bricks and mount process to talk to
glusterd2 about brick ports.


prashanthpai:
This is complete enough to be reviewed but I intend to test this further. Please do not merge.

pmap/pmap.go
```diff
@@ -0,0 +1,272 @@
...
+func isPortFree(port int) bool {
+	conn, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
+	if err != nil {
+		return false
+	}
+	conn.Close()
+	return true
+}
+
+func registrySearchByPort(port int) string {
```
samikshan: Yeah. I am not sure if this function is required. (even on GD1 side). If you still feel this is required, maybe returning a string[] instead would be better.

pmap/pmap.go
```diff
@@ -0,0 +1,272 @@
...
+
+// NOTE: Unlike glusterd1's implementation, the search here is not overloaded
+// with delete operation. This is intentionally kept simple
+func registrySearch(brickname string, ptype PortType) int {
+	registry.RLock()
+	defer registry.RUnlock()
+
+	for p := registry.LastAlloc; p >= registry.BasePort; p-- {
+
+		if registry.Ports[p].Type != GfPmapPortFree {
```
kshlm: Did you mean to do something here?

pmap/pmap.go
```diff
@@ -0,0 +1,272 @@
...
+		port = registrySearch(brickname, ptype)
+		if port != 0 {
+			goto REMOVE
+		}
+	}
+
+	if xprt != nil {
+		port = registrySearchByXprt(xprt, ptype)
+		if port != 0 {
+			goto REMOVE
```
kshlm: Instead of using jumps, just make the REMOVE block a function and call it instead. Makes it cleaner.

PR #247, opened 2017-02-27
Import libmgmt into GD2

kshlm:


kshlm:
This just builds now, doesn't run properly yet. Requires more fixes.

kshlm:
:angry: the dependencies are not co-operating ATM. I fix one, and break another. This will take a little while more.

kshlm:
Now we compile and start. But GD2 doesn't work yet. Need to fix etcd endpoints GD2 uses to make this work.

kshlm:
Nothing seems to be working :rage: . I'll check again tomorrow.

kshlm:
I've opened an [issue](https://github.com/purpleidea/mgmt/issues/170) on mgmt, about making libmgmt leaner. Right now libmgmt pulls in everything mgmt, including stuff like prometheus, libvirt, ssh etc., which we have no use for. If these were not pulled in, this PR would have been a lot smaller, and a have passed CI.

purpleidea:
WHY: https://github.com/gluster/glusterd2/blob/master/glide.lock ???

purpleidea:
>  Vendor github.com/purpleidea/mgmt 

Don't do this! Use git master. We put things in there that we think are good! There are a few exceptions, but it should be perfect for development...

scripts/build.sh
```diff
@@ -14,12 +14,10 @@ VERSION=$($(dirname $0)/pkg-version --full)
...
 
-## This so that mgmt builds without the libvirt and augeas resrouces on centos-ci
-## TODO Remove this once mgmt can be built engine only without resources
-export GOTAGS='novirt noaugeas'
-
 echo "Building $BIN $VERSION"
-go build -ldflags "${LDFLAGS}" -o $OUTDIR/$BIN || exit 1
+## Adding GOTAGS so that mgmt builds without the libvirt and augeas resrouces on centos-ci
+## TODO Remove this once mgmt can be built engine only without resources
+GOTAGS='novirt noaugeas' go build -ldflags "${LDFLAGS}" -o $OUTDIR/$BIN || exit 1
```
roidelapluie: Nope. use:
`go build -ldflags "${LDFLAGS}" -o $OUTDIR/$BIN -tags "novirt noaugeas"`

GOTAGS are transformed in our

scripts/build.sh
```diff
@@ -17,7 +17,7 @@ BIN=$(basename $(go list -f '{{.ImportPath}}'))
 echo "Building $BIN $VERSION"
 ## Adding GOTAGS so that mgmt builds without the libvirt and augeas resrouces on centos-ci
 ## TODO Remove this once mgmt can be built engine only without resources
-GOTAGS='novirt noaugeas' go build -ldflags "${LDFLAGS}" -o $OUTDIR/$BIN || exit 1
+go build -ldflags "${LDFLAGS}" -o $OUTDIR/$BIN -tags "novirt noaugeas" || exit 1
```
roidelapluie: you might want to use build -i https://ttboj.wordpress.com/2017/02/26/faster-golang-builds/

kshlm:
> Please don't do this! Use git master. We put things in there that we think are good! There are a few exceptions, but it should be perfect for development...

It is actually git master that is being vendored. But if you prefer that we not vendor mgmt right now, I will unvendor it.

purpleidea:
@kshlm 

> But if you prefer that we not vendor mgmt right now, I will unvendor it.

I make the recommendation in order to help you make your project successful. You're the glusterd2 maintainer, not me, so it's really up to you. I recommend the same for the other deps as well. If you look in the mgmt vendor/ dir you'll see the minimum that I have found required to vendor etcd within mgmt. HTH

kshlm:
GD2+mgmt works in a single node, so I'm merging this now. @purpleidea you can now begin with your POC.

PR #284, opened 2017-03-21
Switch to etcd/clientv3 for store and remove libkv

### LINKED ISSUE #260: Retire libkv
Our cluster membership APIs use etcd v3 API
Our store consumption/access happens over etcd v2 API (via libkv)

Now that we have zeroed in on etcd as our preferred choice, we can directly use it and move all etcd APIs eing used to v3.

kshlm:
Closes #260


commands/volumes/volume-create.go
```diff
@@ -200,29 +204,33 @@ func volumeCreateHandler(w http.ResponseWriter, r *http.Request) {
 		Store:    "vol-create.Store",
 		Rollback: "vol-create.Rollback",
 		LogFields: &log.Fields{
-			"reqid": uuid.NewRandom().String(),
+			"reqid": reqid,
```
kshlm: My original reason for having them was that a transaction could lead to new requests on other gd2s. But that would just make things more complicated to follow. A single ID the whole way through would be better.

Doing it this way is what [opentracing](http://opentracing.io) sort of does. It allocates an ID at the beginning of a request and the ID flows through the whole cluster as actions are performed. I was looking into it to see if we could make use of it. We should try to use this sort of an established standard to trace operations instead of trying to invent our own.

prashanthpai: > A single ID the whole way through would be better.

Exactly. Currently, we have two. I'll file an issue for this later.

kshlm:
> I'm trying this change out. I have two glusterd2 instances on the same node (using different ports). Volume create request just hangs for me. The CPU usage shoots up to close 100%

I've had it intermittently as well. Seems to be a problem with the locking mechanism. It resolved itself sometimes after sometime. Waiting a little while before doing any operations after probe seems to help. I've only seen this on fresh starts. On restarts, it just works fine.

kshlm:
> I've had it intermittently as well. Seems to be a problem with the locking mechanism. It resolved itself sometimes after sometime. Waiting a little while before doing any operations after probe seems to help. I've only seen this on fresh starts. On restarts, it just works fine.

Seems like the newly added peer is sort of hung, which is preventing etcd operations from happening. Not sure how my changes here is causing this. Need to investigate more.

kshlm:
@prashanthpai This now works properly. The problem was that the embedded etcd was destroyed without disconnecting the store first. This caused `store.Session.Close()` to hang later on when attempted to reconnect as that needs to do some cleanup before closing down.

The first peer was always working right. It used to hang waiting for the new peer to respond.

kshlm:
> Peer probe works fine now. But I see the same issue (CPU usage shoots up) during peer detach now.
> 

Huh, works fine for me. Are you testing with the single node multiple instances?

prashanthpai:
> Huh, works fine for me. Are you testing with the single node multiple instances?

Single node, multiple instance. I'll test this further.

prashanthpai:
It's still the same for me:

```
INFO[2017-04-07T17:31:26+05:30] Etcd embedded server is stopped.             
INFO[2017-04-07T17:31:26+05:30] Etcd data dir, WAL dir and config file removed 
ERRO[2017-04-07T17:31:26+05:30] Could not start embedded etcd server.         Error=listen tcp 192.168.56.25:2380: bind: address already in use
```

This is weird. Out of curiosity, I tried a peer listing on the peer being removed, and here's what I see:

```
[ppai@gd2-1 json]$ curl -s -X GET http://192.168.56.25:23007/v1/peers | python -m json.tool
{
    "Error": "grpc: the client connection is closing"
}
```

prashanthpai:
This is definitely a problem. This session close line is the one that takes a very long time and the CPU usage shoots up:

```go
        if e := s.Session.Close(); e != nil {
                log.WithError(e).Warn("failed to close etcd session")
        }
```

The port is not correctly closed (you check this manually) before calling `etcdmgmt.StartEmbeddedEtcd()`

kshlm:
@prashanthpai I don't seem to be able to hit your problem at all. I do get a slight pause when session.Close is happening, but it always succeeds and etcd restarts correctly.

Considering that this is only a problem during detach, and on the node being detached, I guess for now we can consider this a known issue, and merge this PR. We anyway plan to replace this current solution with elastic-etcd, which shouldn't have this problem.

PR #292, opened 2017-04-05
Introduce middlewares and do some cleanup

### LINKED ISSUE #287: Use one single ID for transaction logging
Here's what a debug log for a txn step currently (master) looks like:

```
DEBU[2017-03-22T14:32:29+05:30] RunStep request recieved                      reqid=9c47b856-f49c-4218-9783-ba38d3246569 stepfunc=vol-create.Stage txnid=4c9aa1e4-e6a6-4ff7-95a7-d861e3e85f73
```

We have to IDs - `reqid` and `txnid`. A single ID the whole way through would be better.

Some discussion about this is in the review comments here: https://github.com/gluster/glusterd2/pull/284

prashanthpai:
* Introduces two minimal middlewares:
    1. Request logger: logs all incoming REST requests
    2. Request ID Generator: Assigns a request ID to each request
* Middlewares are easily chained using alice.
  https://justinas.org/alice-painless-middleware-chaining-for-go/
* Removes a lot of old logging junk. Uses request scoped logger.

TODO: Use context and opentracing in future

Closes #287

### DIFF
--- a/commands/volumes/volume-create.go
+++ b/commands/volumes/volume-create.go
@@ -169,8 +169,7 @@ func nodesForVolCreate(req *volume.VolCreateRequest) ([]uuid.UUID, error) {
 
 func volumeCreateHandler(w http.ResponseWriter, r *http.Request) {
 	req := new(volume.VolCreateRequest)
-	reqID := r.Header.Get("X-Request-ID")
-	logger := log.WithField("reqid", reqID)
+	reqID, logger := restutils.GetReqIDandLogger(r)
 
 	httpStatus, e := unmarshalVolCreateRequest(req, r)
 	if e != nil {
--- a/commands/volumes/volume-create.go
+++ b/commands/volumes/volume-create.go
@@ -19,16 +19,13 @@ import (
 func unmarshalVolCreateRequest(msg *volume.VolCreateRequest, r *http.Request) (int, error) {
 	e := utils.GetJSONFromRequest(r, msg)
 	if e != nil {
-		log.WithField("error", e).Error("Failed to parse the JSON Request")
 		return 422, gderrors.ErrJSONParsingFailed
 	}
 
 	if msg.Name == "" {
-		log.Error("Volume name is empty")
 		return http.StatusBadRequest, gderrors.ErrEmptyVolName
 	}
 	if len(msg.Bricks) <= 0 {
-		log.WithField("volume", msg.Name).Error("Brick list is empty")
 		return http.StatusBadRequest, gderrors.ErrEmptyBrickList
 	}
 	return 0, nil
@@ -150,8 +147,6 @@ func nodesForVolCreate(req *volume.VolCreateRequest) ([]uuid.UUID, error) {
 		// <peer-uuid>:<brick-path>
 		// <ip>:<port>:<brick-path>
 		// <ip>:<brick-path>
-		// TODO: Peer names, as of today, aren't unique. Support it ?
-		// TODO: Change API to have host and path as separate fields
 
 		host, _, err := utils.ParseHostAndBrickPath(b)
 		if err != nil {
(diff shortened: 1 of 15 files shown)


prashanthpai:
I'd also prefer having this as a middleware. I'll work towards that.

transaction/transaction.go
```diff
@@ -20,33 +20,37 @@ const (
...
 
 	return t
 }
 
 // NewTxnWithLoggingContext creates a Txn with a Context with given logging fields
-func NewTxnWithLoggingContext(f log.Fields) *Txn {
-	t := new(Txn)
-	t.ID = uuid.NewRandom()
+func NewTxnWithLoggingContext(f log.Fields, id string) *Txn {
+	t := NewTxn(id)
```
aravindavk: rest of the statements looks duplicate. Can this be,

    t := NewTxn(id)
    t.WithLogFields(f)
    return t

transaction/transaction.go
```diff
@@ -20,33 +20,37 @@ const (
...
-func NewTxn() *Txn {
+func NewTxn(id string) *Txn {
 	t := new(Txn)
-	t.ID = uuid.NewRandom()
-	prefix := txnPrefix + t.ID.String()
 
+	if uuid.Parse(id) != nil {
+		t.ID = uuid.Parse(id)
+	} else {
+		t.ID = uuid.NewRandom()
```
aravindavk: When this can happen?  If reqid is corrupted for any reason, this code silently assigns new id and continues. Parse error can be logged here to catch the same.

transaction/transaction.go
prashanthpai: This `NewTxnWithLoggingContext` isn't used anywhere right now (can as well get rid of it). Retained it for future use if any. Adding a method like `WithLogFields` will change internal state of `t` silently.

prashanthpai:
retest

kshlm:
retest this please

transaction/transaction.go
```diff
@@ -20,33 +20,39 @@ const (
...
 
 // NewTxn returns an initialized Txn without any steps
-func NewTxn() *Txn {
+func NewTxn(id string) *Txn {
 	t := new(Txn)
-	t.ID = uuid.NewRandom()
-	prefix := txnPrefix + t.ID.String()
 
+	parsedID := uuid.Parse(id)
+	if parsedID != nil {
```
kshlm: You could simplify this. `t.ID` has a nil zero value. So you could just parse directly into it.
```
if t.ID = uuid.Parse(); t.ID == nil {
  t.ID = uuid.NewRandom()
}

File: middleware/request_id.go [hunk 4]
```diff
@@ -0,0 +1,31 @@
+package middleware
+
+import (
+	"net/http"
+
+	"github.com/pborman/uuid"
+)
+
+// TODO
+// In Go, the idiomatic and recommended way to attach any request scoped
```
kshlm: I've thought of doing this several times. But each time attempt it, I just get confused at how `context` is supposed to be actually used, and don't proceed forward. But this is definitely something we need to do.

commands/volumes/volume-create.go
```diff
@@ -174,9 +169,12 @@ func nodesForVolCreate(req *volume.VolCreateRequest) ([]uuid.UUID, error) {
 
 func volumeCreateHandler(w http.ResponseWriter, r *http.Request) {
 	req := new(volume.VolCreateRequest)
+	reqID := r.Header.Get("X-Request-ID")
+	logger := log.WithField("reqid", reqID)
```
kshlm: This pattern keeps repeating and should be made into a single function.

prashanthpai:
> We need to think about making use of the context package.

Yes. Using opentracing in Go also depends on this. I'll do that in a separate change later because I expect it to be quite intrusive change. To use `context`, we should pass a Context parameter as the first argument to **every function** on the call path between incoming and outgoing requests.

```go
func DoSomething(ctx context.Context, arg Arg) error {
	// ... use ctx ...
}
```

PR #294, opened 2017-04-06
plugins: Glusterd2 Plugins Infrastructure

### LINKED ISSUE #131: Pluggability

aravindavk:
This is initial code which enables creation of plugins for glusterd2.

Plugin interface is

type Gd2Plugin interface{
    SunRpcProgram() sunrpc.Program
    RestRoutes() route.Routes
    RegisterStepFuncs()
}

Plugin can register sunrpc procedure, REST route and can register step
functions required for Transaction(Transaction APIs are not stable
yet)

Included hello plugin, which plugs following REST routes into
glusterd2 REST server.

GET  /v1/hello
POST /v1/hello

To add new Plugin,

- Create a plugin package and add to import and plugin list in the
  file `$SRC/plugins/plugins.go`
- Create a struct and implement the above interfaces

Closes #131
Signed-off-by: Aravinda VK avishwan@redhat.com

### DIFF
--- a/plugins/common.go
+++ b/plugins/common.go
@@ -5,8 +5,11 @@ import (
 	"github.com/prashanthpai/sunrpc"
 )
 
-type Gd2Plugin interface {
-	SunRpcProgram() sunrpc.Program
+// GlusterdPlugin is an interface that every Glusterd plugin will
+// implement to add sunrpc program, REST routes and Transaction step
+// functions
+type GlusterdPlugin interface {
+	SunRPCProgram() sunrpc.Program
 	RestRoutes() route.Routes
 	RegisterStepFuncs()
 }
--- a/plugins/common.go
+++ b/plugins/common.go
@@ -1,0 +1,12 @@
+package plugins
+
+import (
+	"github.com/gluster/glusterd2/servers/rest/route"
+	"github.com/prashanthpai/sunrpc"
+)
+
+type Gd2Plugin interface {
+	SunRpcProgram() sunrpc.Program
+	RestRoutes() route.Routes
+	RegisterStepFuncs()
+}
--- a/plugins/common.go
+++ b/plugins/common.go
@@ -9,6 +9,7 @@ import (
 // implement to add sunrpc program, REST routes and Transaction step
 // functions
 type GlusterdPlugin interface {
+	Name() string
 	SunRPCProgram() sunrpc.Program
 	RestRoutes() route.Routes
 	RegisterStepFuncs()
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 GOPATH := $(shell go env GOPATH)
 GOBIN := '$(GOPATH)/bin'
+PLUGINS ?= yes
 
 .PHONY: all build check check-go check-reqs install vendor-update verify glusterd2 release check-protoc
 
@@ -22,11 +23,11 @@ check-reqs:
 	@echo
 
 glusterd2:
-	@./scripts/build.sh
+	@PLUGINS=$(PLUGINS) ./scripts/build.sh
 	@echo
 
 install: check-go check-reqs vendor-update
-	@./scripts/build.sh $(GOBIN)
+	@PLUGINS=$(PLUGINS) ./scripts/build.sh $(GOBIN)
 	@echo Setting CAP_SYS_ADMIN for glusterd2 \(requires sudo\)
 	sudo setcap cap_sys_admin+ep $(GOBIN)/glusterd2
 	@echo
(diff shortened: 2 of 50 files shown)


plugins/common.go
```diff
@@ -0,0 +1,12 @@
+package plugins
+
+import (
+	"github.com/gluster/glusterd2/servers/rest/route"
+	"github.com/prashanthpai/sunrpc"
+)
+
+type Gd2Plugin interface {
+	SunRpcProgram() sunrpc.Program
```
prashanthpai: This will mean one plugin can have only one RPC program, which is completely okay for our uses I guess 👍

plugins/common.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"github.com/gluster/glusterd2/servers/rest/route"
+	"github.com/prashanthpai/sunrpc"
+)
+
+// GlusterdPlugin is an interface that every Glusterd plugin will
+// implement to add sunrpc program, REST routes and Transaction step
+// functions
+type GlusterdPlugin interface {
+	SunRPCProgram() sunrpc.Program
```
kshlm: Any examples for why a plugin would like to implement a sunrpc Program? Just curious.

plugins/common.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"github.com/gluster/glusterd2/servers/rest/route"
+	"github.com/prashanthpai/sunrpc"
+)
+
+// GlusterdPlugin is an interface that every Glusterd plugin will
+// implement to add sunrpc program, REST routes and Transaction step
+// functions
+type GlusterdPlugin interface {
```
kshlm: Add a `Name()` method to the interface. This way we have a proper name to be used in logs instead of trying to reflect types.

plugins/common.go
aravindavk: - Quota plugin communicating with Quotad
- Bitrot plugin communicating with Bitd
- gf_event can be sunrpc program to which will be enabled with events plugin. (Once RPC available, other components can make rpc call instead of UDP call as now)

plugins/common.go
aravindavk: Nice suggestion.

purpleidea:
As requested, here's a review.

Conceptually, I think I'm missing more background on the general plugin idea. Are these plugins which are meant to perform any action in gd2? Are they plugins which only represent translators? Can these affect the UI or API parts of the code too?

My general comment wrt plugins is the following: First have a particular interface (call it foo) which is fulfilled by more than one struct which implements this interface. These should be core parts of the code, and should demonstrate useful functionality. Step two is to use the golang 1.8 plugin code to import all the methods for that interface. Is there some existing code which implements `GlusterdPlugin` already?

Lastly the idea of having something very specific in an interface eg: "sunrpc" if the design is meant for more general things, seems like a bit of code smell. So I'd either name this `GlusterdSunrpcPlugin` if it's going to always be sunrpc specific, or remove that method if it's not.

There's nothing wrong IMO about having more than one (or even many) different interfaces which are fulfilled with plugins, so there's no reason for a single catch all.

Here is a specific example. In `mgmt` we currently `don't` have the concept of plugins, but that's only because we haven't implemented it yet. There are a number of things that could very easily be in scope for being plugins, and these are already defined as interfaces. Eg;

1) Res interface (in mgmt/resources/resources.go)
2) GAPI interface (in mgmt/gapi/gapi.go)
3) Language facts (interface is defined, but hasn't hit git master yet)
... and so on.

Hope this was useful, and happy to discuss IRL tomorrow too.

A quick nitpick: Comments above functions are proper sentences. IOW, they should end with a period.
eg:

```
+// SunRPCProgram returns sunrpc program to register with Glusterd
```

prashanthpai:
LGTM 👍 

@kshlm Can you review the build flags added ?

aravindavk:
@kshlm please review.

plugins/common.go
```diff
@@ -0,0 +1,16 @@
...
+	Name() string
+	SunRPCProgram() sunrpc.Program
+	RestRoutes() route.Routes
+	RegisterStepFuncs()
```
kshlm: I don't know why I had not noticed this before.
`RestRoutes()` and `RegisterStepFuncs()` make up the `Command` interface. You could have just embedded it in the plugin interface.

```go
import "github.com/gluster/glusterd2/commands"

type GlusterPlugin interface {
  Name() string
  SunRPCProgram() sunrpc.Program
  commands.Command
}
```

Issue #266, opened 2017-02-27
Design and implement volume set functionality


prashanthpai:
And also provide flexible way to define and validate rules for volume set keys.

kshlm:
We also need to set rules on naming options to avoid situations where multiple options have the same key.

prashanthpai:
To avoid xlator authors having to modify glusterd2 code to let glusterd2 know about xlator options, we can read the symbols in xlator's shared object from go (with caveats). Here's a [sample program](https://github.com/prashanthpai/xlator-sym-go) that demonstrates that.

```sh
[ppai@gd2-1 xlator-sym-go]$ ./main /usr/local/lib/glusterfs/3.12dev/xlator | head -n 100

/usr/local/lib/glusterfs/3.12dev/xlator/cluster/afr.so
read-subvolume
read-subvolume-index
read-hash-mode
choose-local
favorite-child
background-self-heal-count
halo-shd-max-latency
halo-enabled
halo-nfsd-max-latency
halo-max-latency
halo-max-replicas
halo-min-replicas
heal-wait-queue-length
data-self-heal
data-self-heal-algorithm
data-self-heal-window-size
metadata-self-heal
entry-self-heal
data-change-log
metadata-change-log
entry-change-log
optimistic-change-log
inodelk-trace
entrylk-trace
pre-op-compat
eager-lock
self-heal-daemon
iam-self-heal-daemon
iam-nfs-daemon
quorum-type
quorum-count
quorum-reads
node-uuid
post-op-delay-secs
self-heal-readdir-size
ensure-durability
afr-dirty-xattr
afr-pending-xattr
metadata-splitbrain-forced-heal
heal-timeout
consistent-metadata
arbiter-count
shd-max-threads
shd-wait-qlength
locking-scheme
granular-entry-heal
favorite-child-policy
consistent-io
use-compound-fops

/usr/local/lib/glusterfs/3.12dev/xlator/cluster/dht.so
lookup-unhashed
lookup-optimize
min-free-disk
min-free-inodes
unhashed-sticky-bit
use-readdirp
assert-no-child-down
directory-layout-spread
decommissioned-bricks
rebalance-cmd
commit-hash
node-uuid
rebalance-stats
readdir-optimize
rsync-hash-regex
extra-hash-regex
rebalance-filter
xattr-name
weighted-rebalance
local-volume-name
tier-pause
tier-promote-frequency
tier-demote-frequency
write-freq-threshold
read-freq-threshold
[...]

PR #297, opened 2017-04-18
options: Framework to manage Cluster and Volume options

aravindavk:
This framework provides facility to store the Cluster options and volume
options in etcd store.(`volume set` command in glusterd 1)

During Get request, fetches the value from etcd store, if not found fetches
from the defaults list. Returns error if not found in both places.

Validation facility added while adding/updating options in store.

Following APIs exposed under `options` subpackage

    import "github.com/gluster/glusterd2/options"

    options.GetClusterOption(key string) (interface{}, error)
    options.GetVolumeOption(volname string, key string) (interface{}, error)

    options.SetClusterOption(key string, value interface{}) error
    options.SetVolumeOption(volname string, key string,
                            value interface{}) error

    options.ResetClusterOption(key string) error
    options.ResetVolumeOption(volname string, key string) error

TODO:
- Implement new APIs `GetAllClusterOptions` and `GetAllVolumeOptions`
- Add all existing default options, values and validation functions

This patch adds the framework required for Issue #266 and volgen.

Signed-off-by: Aravinda VK <mail@aravindavk.in>

### DIFF
--- a/options/common.go
+++ b/options/common.go
@@ -39,12 +39,16 @@ func GetClusterOption(key string) (interface{}, error) {
 	}
 
 	// Else return error
-	return "", errors.New("Invalid Option")
+	return nil, errors.New("Invalid Option")
 }
 
 // SetClusterOption validates and saves the cluster configuration in
 // etcd store.
 func SetClusterOption(key string, value interface{}) error {
+	if _, ok := clusterDefaultOptions[key]; !ok {
+		return errors.New("Invalid Key")
+	}
+
 	clusterKey := clusterOptsPrefix + key
 	if clusterOptValidate(key, value) {
 		d, err := json.Marshal(value)
@@ -59,6 +63,10 @@ func SetClusterOption(key string, value interface{}) error {
 
 // ResetClusterOption deletes cluster configurations stored in etcd store.
 func ResetClusterOption(key string) error {
+	if _, ok := clusterDefaultOptions[key]; !ok {
+		return errors.New("Invalid Key")
+	}
+
 	clusterKey := clusterOptsPrefix + key
 	_, e := gdctx.Store.Delete(context.TODO(), clusterKey)
 	return e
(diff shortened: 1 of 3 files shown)


prashanthpai:
retest this please

options/common.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	// If not available get value from defaults
+	dv, dvOk := clusterDefaultOptions[key]
+	if dvOk {
+		return dv.defaultValue, nil
+	}
+
+	// Else return error
+	return "", errors.New("Invalid Option")
```
prashanthpai: Please use `nil` instead of empty string as return type is `interface{}`. Also in `GetVolumeOption()`

options/validations.go
```diff
@@ -0,0 +1,37 @@
+package options
+
+func clusterOptValidate(key string, value interface{}) bool {
```
prashanthpai: This function name is a little misleading as you are only validating the value here. What if `clusterDefaultOptions[key]` doesn't exist ? Won't this return true ?

options/options.go
```diff
@@ -0,0 +1,23 @@
+package options
+
+type validationFunc func(val interface{}) bool
+
+type optionDetails struct {
+	defaultValue    interface{}
+	validationFuncs []validationFunc
+	help            string
+}
+
```
prashanthpai: @kshlm wanted to avoid having duplication of xlator options. I can't think of any other way though, especially the options that changes the volfile graph seems needed here.

options/common.go
aravindavk: ok

options/options.go
aravindavk: Unless Translator exports defaults in a consumable way, avoiding this is not possible. We can change this later if Translator itself provides defaults.

options/validations.go
aravindavk: Good catch. I will add the validation for key name.

options/options.go
kshlm: We intend to have xlators export their option table in a consumable manner. I'd rather have us not hardcode any sort of option table in GD2 from the beginning itself. But I'm okay with this for now.

aravindavk:
This patch not required now.

PR #303, opened 2017-05-04
Import libmgmt into GD2

kshlm:


kshlm:
While this builds right now, we cannot run GD2. On start GD2 immediately panics as it inits all the packages with,
```
panic: http: multiple registrations for /debug/requests

goroutine 1 [running]:
net/http.(*ServeMux).Handle(0x2113c40, 0x167f80b, 0xf, 0x20b9260, 0x16c3d88)
        /usr/lib/go/src/net/http/server.go:2254 +0x610
net/http.(*ServeMux).HandleFunc(0x2113c40, 0x167f80b, 0xf, 0x16c3d88)
        /usr/lib/go/src/net/http/server.go:2286 +0x55
net/http.HandleFunc(0x167f80b, 0xf, 0x16c3d88)
        /usr/lib/go/src/net/http/server.go:2298 +0x4b
golang.org/x/net/trace.init.1()
        /home/kaushal/go/src/golang.org/x/net/trace/trace.go:121 +0x42
golang.org/x/net/trace.init()
        /home/kaushal/go/src/golang.org/x/net/trace/trace_go17.go:22 +0x1cd
github.com/purpleidea/mgmt/vendor/google.golang.org/grpc.init()
        /home/kaushal/go/src/github.com/purpleidea/mgmt/vendor/google.golang.org/grpc/trace.go:120 +0x78
github.com/purpleidea/mgmt/vendor/github.com/coreos/etcd/etcdserver/etcdserverpb.init()
        /home/kaushal/go/src/github.com/purpleidea/mgmt/vendor/github.com/coreos/etcd/etcdserver/etcdserverpb/rpc.pb.gw.go:1912 +0x9b
github.com/purpleidea/mgmt/vendor/github.com/coreos/etcd/clientv3.init()
        /home/kaushal/go/src/github.com/purpleidea/mgmt/vendor/github.com/coreos/etcd/clientv3/watch.go:779 +0x5a
github.com/purpleidea/mgmt/etcd.init()
        /home/kaushal/go/src/github.com/purpleidea/mgmt/etcd/world.go:71 +0x9b
github.com/purpleidea/mgmt/lib.init()
        /home/kaushal/go/src/github.com/purpleidea/mgmt/lib/main.go:617 +0x8e
github.com/gluster/glusterd2/mgmt.init()
        /home/kaushal/go/src/github.com/gluster/glusterd2/mgmt/mgmt.go:53 +0x53
main.init()
        /home/kaushal/go/src/github.com/gluster/glusterd2/version.go:19 +0x89
```
mgmt and GD2 have vendored grpc individually. The grpc package imports `golang.org/x/net/trace` which registers a `/debug/requests` handler with the net/http on package intialization during program startup. The GD2 vendored grpc imports the vendored net/trace package. The mgmt grpc imports the global net/trace package. So we have 2 instances of net/trace pacakge during runtime, both of which attempt to register the a handler for the same path, leading to the panic we see.

One way to avoid this would be to vendor mgmt in GD2. This way mgmt would use the packages vendored by GD2 instead of its own vendored packages, avoiding the above situation.

purpleidea:
> mgmt and GD2 have vendored grpc individually

If your patch uses the mgmt implementation of etcd, then wouldn't this fix this issue, and also fix your etcd proxy issue?

prashanthpai:
> If your patch uses the mgmt implementation of etcd

Are you suggesting that we change existing etcd clustering present in glusterd2 and depend on mgmt for clustering (a functional change) just to fix a vendoring conflict !? That is not what I am looking for. mgmt can consume glusterd's etcd instance or for now, create it's own. In any case, I'd prefer mgmt to not replace existing etcd clustering at all but to co-exist as an optional way of doing things.

purpleidea:
> depend on mgmt for clustering

This is what we had discussed IRL as one of the aspects to work on. @amarts @atinmu

kshlm:
GD2 uses gRPC as well, and not just through etcd. So no, the problem wouldn't be fixed by using the mgmt instance of etcd.

I'm vendoring mgmt into GD2. Doing this solves the problem. mgmt is not locked to a specific version in `glide.yaml`. When you want to use a newer version of mgmt, or any other dependency, running `glide update` should move everything not pinned to the latest master.

purpleidea:
@kshlm Can't we both use the same grpc version? We should probably just pick the same one that is used by the latest stable etcd. Would that work for you?

kshlm:
> @kshlm Can't we both use the same grpc version? We should probably just pick the same one that is used by the latest stable etcd. Would that work for you?

GD2 is using the same version as the current stable etcd-3.1.7.

Even if mgmt switched to the same version, there would still be 2 instances of the gRPC package, `github.com/gluster/glusterd2/vendor/google.golang.org/grpc` and `github.com/purpleidea/mgmt/vendor/google.golang.org/grpc`. 

Both mgmt and GD2 need to source the package from the same path.

purpleidea:
On Fri, May 5, 2017 at 5:21 AM, Kaushal M <notifications@github.com> wrote:

> Both mgmt and GD2 need to source the package from the same path.


I feel like there must be a clean solution to this. I can't expect we're
the first to hit this. Want to open a ticket upstream about the http panic
you got ?

main.go
```diff
@@ -74,6 +75,7 @@ func main() {
 	super := initGD2Supervisor()
 	super.ServeBackground()
 	super.Add(servers.New())
+	super.Add(mgmt.New())
```
prashanthpai: I'd prefer this be done only when some config option (call it --mgmt or something) is turned on and not by default.

Even better if build tags can be used to conditionally compile and use `mgmt`. If this can be done easily, I'm okay for mgmt to be compiled by default while maintaining an option to exclude compiling it.

purpleidea:
On Tue, May 9, 2017 at 2:41 AM, Prashanth Pai <notifications@github.com> wrote:
>
> I'd prefer this be done only when some config option (call it --mgmt or something) is turned on and not by default.
>
> Even better if build tags can be used to conditionally compile and use mgmt. If this can be done easily, I'm okay for mgmt to be compiled by default while maintaining an option to exclude compiling it.

I have no problems not activating it unless there is a --mgmt like
option passed in, as long as it's not behind a build flag, since I'd
like it to get constantly built in in git master. Although to do the
etcd for glusterd2 stuff I thought you'd always want that running?

prashanthpai:
>  as long as it's not behind a build flag, since I'd like it to get constantly built in in git master

The build flag I requested is to disable building mgmt. Unless specified otherwise by build flag (example flag name `nomgmt`) , mgmt will be built by default in git master.

> Although to do the etcd for glusterd2 stuff I thought you'd always want that running?

As I've mentioned earlier, I'd prefer mgmt's management of etcd to be a "mode" that can be turned on and off. It should **co-exist amicably and not replace** current implementation. If it is deemed in future that mgmt does not satisfy glusterd's requirements, I'd want the provision for it to to be turned off and/or removed  without incurring major code changes or effort. A co-existing implementation also allows us to do an actual non-theoretical comparison of both "modes" - maintenance, correctness and performance wise.

prashanthpai:
I did a quick sanity check with mgmt being built but not turned on. Existing implementation seems to work fine. Merging.

purpleidea:
w00t! <fireworks emoji>

PR #308, opened 2017-05-19
Error handling cleanup

### LINKED ISSUE #296: Call Fatal() only from main()
Fatal() should only be called during startup and for non-recoverable errors, usually only in `main()`. 

```
[ppai@gd2-1 glusterd2]$ grep -nR "Fatal" . | grep -v vendor
./peer/self.go:17:		log.WithField("err", e).Fatal("Failed to list member in etcd cluster")
./peer/self.go:37:		}).Fatal("Failed to add peer into the etcd store")
./store/store.go:34:		log.WithError(e).Fatal("failed to create etcd client")
./store/store.go:42:		log.WithError(e).Fatal("failed to create an etcd session")
./utils/utils.go:254:			"xattr":     testXattr}).Fatal("removexattr failed")
./utils/utils.go:312:				}).Fatal("failed to create directory")
./utils/utils.go:320:			}).Fatal("failed to access directory")
./utils/utils.go:328:		}).Fatal("directory path is not a directory")
./utils/utils.go:337:		}).Fatal("directory path is not a writable")
./config.go:61:		}).Fatal("Failed to get current working directory.")
./config.go:88:		log.Fatal("Invalid peer address specified.")
./config.go:104:			log.Fatal("Invalid etcd addresses specified.")
./config.go:148:			}).Fatal("failed to read given config file")
./logging.go:31:		log.WithField("error", err).Fatal("Failed to parse log level")
./logging.go:45:			log.WithError(logFileErr).Fatalf("Failed to open log file %s", logFilePath)
./gdctx/myuuid.go:33:			}).Fatal("failed to read MyUUID from file")
./gdctx/myuuid.go:57:		}).Fatal("failed to write MyUUID to file")
./gdctx/global.go:67:		log.Fatal("SetHostnameAndIP: Could not get IP address")
./gdctx/global.go:73:		log.Fatal("SetHostnameAndIP: Could not get hostname")
Binary file ./build/glusterd2 matches
./servers/muxsrv/server.go:24:		log.WithError(err).Fatal("failed to create gd2-muxsrv listener")
./etcdmgmt/logging.go:20:		log.WithError(err).Fatalf("Failed to open etcd log file %s", etcdLogFile)
./main.go:48:		log.WithError(err).Fatalf("Failed to change working directory to %s", workdir)
./main.go:60:		log.WithError(err).Fatal("Could not fetch config options for etcd")
./main.go:64:		log.WithError(err).Fatal("Could not start embedded etcd server")
```

Thanks @purpleidea for reminding about this.

prashanthpai:
* Bubble up errors all the way upto main(). Only main() should call log.Fatal()
* Cleanup and simplify gdctx package

Closes #296 


purpleidea:
Not doing a comprehensive review, since my previous long review was mostly ignored, but glad to see this issue is being looked into! Two quick comments:

1) https://github.com/gluster/glusterd2/pull/308/files#diff-cd575165f84eed10d10eb1460b0927b4R303 -- usually you either want to return (and propagate up) OR log in place. Not both. Maybe I misread the patch if there's a reason for this.

2) There seems to be a lot of goto's in the code. This reads like C code, and much less like golang IMO. It's not that it's necessarily illegal or invalid, but far less common. Up to you if there's a better way or not. I personally have avoided these, and haven't see many in production, but the data is slim here.

HTH!
James

prashanthpai:
> you either want to return (and propagate up) OR log in place

Usually, that's the case. As you mentioned, this PR actually removes logging done at few inner functions. At places where I have retained the logging, it's always logged at level `DEBUG` which will not be the default when glusterd2 is shipped. It's useful when one function can return different errors (multiple exit points). A rule of thumb I follow is that inner most functions shouldn't log at all or use `DEBUG` level if logged.

> There seems to be a lot of goto's in the code. This reads like C code, and much less like golang IMO/

Couldn't agree more. I guess during the early days of development, C developers converted glusterd1's C code into Go verbatim. Some of them need to be fixed, yes. This PR isn't for that though. Patches to fix them are welcome :)

OTOH, purely my personal preference: I'd justify using `goto` for these usecases:
* getting out of nested loops
* ensuring single exit point for the function when the thing being returned needs some processing or there's cleanup to do

logging.go
```diff
@@ -34,8 +34,9 @@ func initLog(logdir string, logFileName string, logLevel string) {
 
 	l, err := log.ParseLevel(strings.ToLower(logLevel))
 	if err != nil {
-		log.SetOutput(os.Stderr)
-		log.WithField("error", err).Fatal("Failed to parse log level")
+		setLogOutput(os.Stderr)
+		log.WithField("error", err).Debug("Failed to parse log level")
```
kshlm: Should be `log.WithError`

gdctx/global.go
```diff
@@ -30,47 +33,63 @@ var (
...
+		default:
+			log.WithError(err).WithField("path", uuidFile).Debug(
+				"failed to read UUID from file")
+			return err
+		}
+	}
+	// use the UUID found in file
+	MyUUID = uuid.Parse(string(ubytes))
+	if MyUUID == nil {
+		return errors.New("Failed to parse UUID found in file")
```
kshlm: As per the new rule in #307 this error string should begin with a lowercase letter

purpleidea: Most of these changes can be caught with golint. :)

purpleidea: (I like forcing the computers to do the busy work for us... hehe!)

prashanthpai: @purpleidea golint doesn't catch them unfortunately. We have golint turned on in automated CI tests

purpleidea: @prashanthpai Actually it does:

```
./golint /tmp/foo.go
/tmp/foo.go:7:13: error strings should not be capitalized or end with punctuation or a newline
james@computer:~/tmp/code/gopath/bin$ cat /tmp/foo.go
package main

import "errors"

func main() {

	errors.New("Fooo!")

}
```

What doesn't happen is that if it's *only* capitalized, it often won't report unless you lower the confidence:

```
./golint -min_confidence 0.6 /tmp/foo.go
/tmp/foo.go:7:13: error strings should not be capitalized or end with punctuation or a newline
james@computer:~/tmp/code/gopath/bin$ cat /tmp/foo.go
package main

import "errors"

func main() {

	errors.New("Fooo")

}

```

HTH,
James

prashanthpai: > it often won't report unless you lower the confidence

Yeah. We run the CI tests with default (0.8) confidence.

Issue #252, opened 2017-02-27
Authentication for ReST API


prashanthpai:

aravindavk:
REST API authentication patch is merged. I will close this issue and open new one for REST users management.

PR #315, opened 2017-06-07
auth: Initial patch for authenticating local request

aravindavk:
Introduced new middleware for Authentication. This middleware
allows local requests without Auth.

Access to local users can be controlled using iptable rules

For example, Glusterd2 REST API port is 24007

    iptables -I OUTPUT -p tcp --dport 24007 -j REJECT
    iptables -I OUTPUT -p tcp --dport 24007 --match owner \
        --gid-owner gluster -j ACCEPT
    iptables -I OUTPUT -p tcp --dport 24007 --match owner \
        --uid-owner root -j ACCEPT

Updates: #252

### DIFF
--- a/middleware/auth.go
+++ b/middleware/auth.go
@@ -1,0 +1,32 @@
+package middleware
+
+import (
+	"net/http"
+	"strings"
+)
+
+func isAuthorized(r *http.Request) bool {
+	// TODO: Add configuration to enable this Auth
+	// This configuration should not be as command line option
+	// Because, one glusterd can be run with `--no-auth` and
+	// other glusterd with auth enabled.
+	if strings.HasPrefix(r.RemoteAddr, "127.0.0.1:") {
+		return true
+	}
+	// TODO: Add Auth for requests from Remote IPs
+	// Without that, this feature is no impact on
+	// existing(Without Auth). Returning true so that it
+	// will not break the existing consumers/tests till Auth introduced
+	return true
+}
+
+// Auth is a middleware which validates the incoming request is from local node or remote node
+func Auth(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if isAuthorized(r) {
+			next.ServeHTTP(w, r)
+		} else {
+			http.Error(w, "Unauthorized", http.StatusUnauthorized)
+		}
+	})
+}
--- a/servers/rest/rest.go
+++ b/servers/rest/rest.go
@@ -38,7 +38,7 @@ func NewMuxed(m cmux.CMux) *GDRest {
 
 // Serve begins serving client HTTP requests served by REST server
 func (r *GDRest) Serve() {
-	chain := alice.New(middleware.LogRequest, middleware.ReqIDGenerator).Then(r.Routes)
+	chain := alice.New(middleware.Auth, middleware.LogRequest, middleware.ReqIDGenerator).Then(r.Routes)
 	log.WithField("ip:port", r.listener.Addr().String()).Info("Started GlusterD ReST server")
 	if err := http.Serve(r.listener, chain); err != nil {
 		//TODO: Correctly handle valid errors. We could also be having errors when stopping


middleware/auth.go
```diff
@@ -0,0 +1,32 @@
...
+	return true
+}
+
+// Auth is a middleware which validates the incoming request is from local node or remote node
+func Auth(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if isAuthorized(r) {
+			next.ServeHTTP(w, r)
+		} else {
+			http.Error(w, "Unauthorized", http.StatusUnauthorized)
```
prashanthpai: A `403` (http.StatusForbidden) is more suited here because as per HTTP [RFC](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) a `401 Unauthorized` response MUST include a WWW-Authenticate header field to allow client to authenticate again.


Also, the body can contain something more specific which says that requests from non-local IPs will not be served.

servers/rest/rest.go
```diff
@@ -38,7 +38,7 @@ func NewMuxed(m cmux.CMux) *GDRest {
 
 // Serve begins serving client HTTP requests served by REST server
 func (r *GDRest) Serve() {
-	chain := alice.New(middleware.LogRequest, middleware.ReqIDGenerator).Then(r.Routes)
+	chain := alice.New(middleware.Auth, middleware.LogRequest, middleware.ReqIDGenerator).Then(r.Routes)
```
prashanthpai: The `LogRequest` should come at first because we want all request and response to be logged, including 401/403

```
INFO[2017-06-08T07:45:48+05:30] 127.0.0.1 - - [08/Jun/2017:07:45:48 +0530] "GET /v1/peers HTTP/1.1" 200 131 
```

middleware/auth.go
```diff
@@ -0,0 +1,32 @@
...
+	"net/http"
+	"strings"
+)
+
+func isAuthorized(r *http.Request) bool {
+	// TODO: Add configuration to enable this Auth
+	// This configuration should not be as command line option
+	// Because, one glusterd can be run with `--no-auth` and
+	// other glusterd with auth enabled.
+	if strings.HasPrefix(r.RemoteAddr, "127.0.0.1:") {
```
prashanthpai: Please check ipv6 address too for the string `[::1]`. On my system localhost resolves to ipv6 address by default.

purpleidea: Instead of strings.XXX I think it might be worth considering use of the proper types for this, eg: https://golang.org/pkg/net/#IP

kshlm: We should stop matching network addresses as strings. We do that in GD1 and in protocol/auth, and it isn't really good.

aravindavk:
@kshlm what are your plans to protect sunrpc port and etcd ports, are you planning to use the same token based approach or IPtables based approach to prevent access from non root(or non gluster) users.

middleware/auth.go
aravindavk: Makes sense, will change it to 403

middleware/auth.go
aravindavk: r.RemoteAddr pattern is like <host>:<random_port>(Not 24007, may be multiple go routine gets different port).

- r.RemoteAddr data type is string, which includes host name and port information. String comparision is one step compared to `net.SplitHostPort(ip)`, `net.ParseIP(host)` and `ip.Equal()`
- Also this field is not set by the user, no need to validate that it is valid IP or not, we just have to compare with existing pattern.
- We are not resolving IP here, so not finding enough benefit to use net.* functions for comparison.

kshlm: Doing string compares solve just this one case easily. The TODO below to allow Auth requests from other IPs would require you to use `net` comparisons anyway.

We should move the whole of Gluster to use CIDR representation and matching for auth. This is the standard. (I need to open an RFE  issue for this). GD2 should as well.

Also, the `net.Parse` functions don't do any resolution AFAIK, so the impact should be negligible.

kshlm:
> @kshlm what are your plans to protect sunrpc port and etcd ports, are you planning to use the same token based approach or IPtables based approach to prevent access from non root(or non gluster) users.

The token approach wouldn't work for sunrpc. If we would need to do the auth the same way we do know on 3.x, based on addresses. But we can do it better by using CIDR as mentioned in my other comment.

Etcd V3 API suggests using TLS for security, and using certificate based auth if needed.

TLS and certificates is the only method that would work for all three.

middleware/auth.go
aravindavk: TODO below is not for allow/deny based on IP. That will be based on Token based auth like JWT.(To be decided yet on this topic)

If IP filters to be implemented, that is entirely different feature since allowed IP list can't skip authentication.

Only `net.Parse` is not enough for comparison. I don't think constant for localhost is available in `net` package(like `LocalhostIP`), So the comparison is still based on string but with additional steps.

Example program,

    package main
     
    import (
        "net"
        "fmt"
    )
     
    func main(){
        input1 := "127.0.0.1:24007"
        input2 := "[::1]:24007"
        localhost := net.ParseIP("::1")
        host1, port1, err1 := net.SplitHostPort(input1)
        if err1 != nil{
            fmt.Println("IP split failed for input 1")
            return
        }
     
        host2, port2, err2 := net.SplitHostPort(input2)
        if err2 != nil{
            fmt.Println("IP split failed for input 2")
            return
        }   
        ip1 := net.ParseIP(host1)
        ip2 := net.ParseIP(host2)
        fmt.Println("Localhost: ", localhost,
            "\nInput1: ", input1,
            "\nInput2: ", input2,
            "\nHost 1: ", host1, " Port 1: ", port1,
            "\nHost 2: ", host2, " Port 2: ", port2,
            "\nInput 1 Equal to Localhost: ", ip1.Equal(localhost),
            "\nInput 2 Equal to Localhost: ", ip2.Equal(localhost),
        )
    }


Outputs: You can note from below output, `ip.Equal` is not suceeding for all the variant of localhost. There may be other tricks to achieve the same, but I am not aware of it.

Output When `localhost := "127.0.0.1"`

    Localhost:  127.0.0.1 
    Input1:  127.0.0.1:24007 
    Input2:  [::1]:24007 
    Host 1:  127.0.0.1  Port 1:  24007 
    Host 2:  ::1  Port 2:  24007 
    Input 1 Equal to Localhost:  true 
    Input 2 Equal to Localhost:  false

Output When `localhost := "::1"`

    Localhost:  ::1 
    Input1:  127.0.0.1:24007 
    Input2:  [::1]:24007 
    Host 1:  127.0.0.1  Port 1:  24007 
    Host 2:  ::1  Port 2:  24007 
    Input 1 Equal to Localhost:  false 
    Input 2 Equal to Localhost:  true

aravindavk:
> The token approach wouldn't work for sunrpc. If we would need to do the auth the same way we do know on 3.x, based on addresses. But we can do it better by using CIDR as mentioned in my other comment.

> Etcd V3 API suggests using TLS for security, and using certificate based auth if needed.

> TLS and certificates is the only method that would work for all three.

CIDR/TLS are for transport layer security. Will not solve multi user problem in system unless certificate access is limited by User permissions.

aravindavk:
This patch not required now.

Issue #135, opened 2016-09-22
Integrate gRPC with GlusterFS


kshlm:
We'll be using gRPC as the RPC framework of choice in GD2. #128 is tracking progress for this.

Related to this, we'll also need to integrate gRPC with GlusterFS to allow bricks, daemons and clients to communicate with GD2.

kshlm:
#128 changed the default inter GD2 RPC framework to gRPC. I'll leave this issue open till integration with glusterfsd is done.

prashanthpai:
We're using sunrpc for interaction with glusterfsd and other gluster daemons. Closing this issue.

PR #324, opened 2017-06-19
ElasticEtcd

kshlm:
This PR brings in an elastic etcd package and integrates it with the GD2 store and peer workflow.

There are still a lot of TODOs left to do, along with known issues. These will be noted as issues and addressed as we progress further with GD2.


purpleidea:
What's the story on this code, and why not use the mgmtconfig stuff as a base?

kshlm:
> What's the story on this code, and why not use the mgmtconfig stuff as a base?

I tried using mgmt/etcd, but it isn't usable as a standalone package. I saw if I could refactor it, to make it usable as a standalone package, but it had a lot of extra code for integration with mgmt, that was hard to refactor out. So, I went with a reimplementation of the same algorithm that mgmt/etcd uses. But this package is usable standalone, and could be used by mgmt itself if it wanted.

purpleidea:
> I tried using mgmt/etcd, but it isn't usable as a standalone package. I saw if I could refactor it, to make it usable as a standalone package, but it had a lot of extra code for integration with mgmt, that was hard to refactor out. So, I went with a reimplementation of the same algorithm that mgmt/etcd uses. But this package is usable standalone, and could be used by mgmt itself if it wanted.

Would have been nice if you had discussed with me as I would have been happy to help you refactor my implementation to help suit your needs. Of course, I don't believe this is truly necessary since the imported mgmt/lib already brings this in for you.

In any case, it's usually classy to leave a comment in the source and/or commit thanking the original author if you've used some of their code and/or algorithm.

Cheers!

kshlm:
> Would have been nice if you had discussed with me as I would have been happy to help you refactor my implementation to help suit your needs. Of course, I don't believe this is truly necessary since the imported mgmt/lib already brings this in for you.
> 
> In any case, it's usually classy to leave a comment in the source and/or commit thanking the original author if you've used some of their code and/or algorithm.
> 
> Cheers!

I did talk to you about the refactoring, and about doing it as a new package. It's been a long while though.

And yes. I will be giving credit. I've just not added much of the documentation yet.

store/store.go
```diff
@@ -2,108 +2,113 @@
...
-	livenessKeyPrefix = GlusterPrefix + "alive/"
+	GlusterPrefix = "gluster/"
+	sessionTTL    = 30 // used for etcd mutexes and liveness key
 )
 
-// Store variable can be imported by packages which need access to the store
-var Store *GDStore
+var (
+	// Store is the default GDStore that must to be used by the packages in GD2
+	Store *GDStore
```
kshlm: I'm trying to find a way to correctly guard this.
This current method with the mutex is half baked, not all access is guarded. In this method I'd have to wrap the all the methods of the etcd client, with a new function that uses this lock. This doesn't feel nice to me.

store/config.go
```diff
@@ -0,0 +1,141 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+
+	"github.com/gluster/glusterd2/pkg/elasticetcd"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/pelletier/go-toml"
```
kshlm: We don't have a dependency on direct dependency on yaml. We have an indirect on through spf13/viper.

I prefer toml to yaml. Mainly because toml ignores indentation, and yaml doesn't. This yaml feature has left me frustrated many times.

kshlm:
@prashanthpai I'll need your help to fix e2e testing. It fails with a panic.

kshlm:
> Feature request: It'll be really great to optionally compile embedded etcd. This is for environments like kubernetes and openshift where etcd instances can be in their own pods and glusterd2 binary will be light-weight containing only the business logic and doesn't contain embedded etcd.

Just tested this out, the resulting binary isn't any smaller. Not much to gain by doing this.

```
~/g/s/g/g/glusterd2 ❯❯❯ ls -l build                                                                                                                                                                                         
total 110784
-rwxr-xr-x 1 kaushal kaushal 35623912 Jun 22 19:10 glusterd2-with-elastic*
-rwxr-xr-x 1 kaushal kaushal 35544096 Jun 22 19:10 glusterd2-without-elastic*
```

kshlm:
> Started reviewing this. It's quite a big change. I still haven't looked at the elasticetcd package yet. It will take some time.

I've tried to keep the flow of the package clear and understandable. But if it helps do the review faster, I can sit with you and go over the elasticetcd package.

prashanthpai:
> Just tested this out, the resulting binary isn't any smaller. Not much to gain by doing this.

Interesting. Was this without mgmt (also pulls etcd) ?
Either ways, optional building of elastic etcd is not a blocker.

store/store.go
prashanthpai: We may also need mechanism (global lock or graceful server shutdown) to stop accepting new client connections (HTTP and sunrpc) during a store restart (add/remove peer). The goroutines that access the store will be serving client requests.

purpleidea:
On Fri, Jun 23, 2017 at 12:34 AM, Prashanth Pai
<notifications@github.com> wrote:
>
> We may also need mechanism (global lock or graceful server shutdown) to stop accepting new client connections (HTTP and sunrpc) during a store restart (add/remove peer). The goroutines that access the store will be serving client requests.

Is there a way to clean up cleanly with sunrpc? My (basic)
understanding was that due to the stream nature of the protocol (that
you explained) this wasn't quite possible, which I assume is one of
the reasons the golang RPC package was deprecated.

In any case, the idea to port glusterfsd away from sunrpc to grpc, or
something else is definitely a +1 in my book.

prashanthpai:
> Is there a way to clean up cleanly with sunrpc? My (basic)
understanding was that due to the stream nature of the protocol (that
you explained) this wasn't quite possible, which I assume is one of
the reasons the golang RPC package was deprecated.

You're right. The only way to do it is to move away from `net/rpc` and implement our own rpc framework. But right now, the benefit isn't too much in doing that. The sunrpc clients (glusterfs mount process and glusterfsd) are resilient to glusterd sunrpc server going down. IOW, they keep reconnecting on failures.

> In any case, the idea to port glusterfsd away from sunrpc to grpc, or
something else is definitely a +1 in my book.

Initially, [that was the plan](https://github.com/gluster/glusterd2/issues/135). But that won't happen as we realized that it's a bad idea. grpc doesn't have a consumable C interface and adding such an interface to grpc is complex and non-trivial amount of work. Even if it did have a C interface, we have to modify glusterfsd and glusterfs code to use grpc. Writing a sunrpc codec for `net/rpc` was quick and straight-forward. Performance isn't critical as it's not in the I/O path. Using sunrpc also meant that glusterfs and glusterfsd processes can seamlessly talk to both glusterd1 and glusterd2 i.e be drop-in replacement for old glusterd.

store/store.go
```diff
@@ -2,108 +2,124 @@
...
-	switch nodeID.(type) {
-	case uuid.UUID:
-		keySuffix = nodeID.(uuid.UUID).String()
-	case string:
-		keySuffix = nodeID.(string)
-		if uuid.Parse(keySuffix) == nil {
-			return false
+// New creates a new GDStore from the given Config.
+// If the given Config is nil, the saved store config is used.
+func New(conf *Config) (*GDStore, error) {
```
prashanthpai: This need not be exported. Only `store.Init()` calls this.

store/remote.go
```diff
@@ -0,0 +1,53 @@
...
+		log.WithError(e).Error("failed to create etcd client")
+		return nil, e
+	}
+	log.Debug("etcd client connection created")
+
+	// Create a new session (lease kept alive for the lifetime of a client)
+	// This is currently used for:
+	// * distributed locking (Mutex)
+	// * representing liveness of the client
+	s, e := concurrency.NewSession(c, concurrency.WithTTL(sessionTTL))
```
prashanthpai: You can move the `sessionTTL` to this file as it's the only place where it's used currently. But, it also means that in case of embedded store, when `publishLiveness()` is called, the TTL is different (default for the session i.e 60s)

pkg/elasticetcd/logging.go
```diff
@@ -0,0 +1,39 @@
...
+		f, err := os.OpenFile(path.Join(ee.conf.Dir, "elastic.log"), os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
+		if err != nil {
+			ee.log.Out = os.Stdout
+			return
+		}
+		ee.log.Out = f
+		ee.logFile = f
+	}
+
+	ee.log.Info("beginning etcd logging logging")
```
prashanthpai: nit: Not true when `DisableLogging` is true

pkg/elasticetcd/elastic.go
```diff
@@ -0,0 +1,89 @@
...
+	ee.conf = conf
+	ee.stopwatching = make(chan bool)
+	ee.initLogging()
+
+	// If no endpoints are given or if the default endpoint is set, assume that there is no existing server
+	if len(ee.conf.Endpoints) == 0 || isDefaultEndpoint(ee.conf.Endpoints) {
+		ee.log.Debug("no configured endpoints, starting own server")
+
+		if err := ee.startServer(""); err != nil && err != ErrClientNotAvailable {
+			ee.log.WithError(err).Debug("failed to start server")
```
prashanthpai: fd created in `ee.initLogging()` leaks here.

pkg/elasticetcd/logging.go
kshlm: But it (along with all the others) doesn't get logged anyway.

store/remote.go
kshlm: This can be done for the embedded store as well, but for now, it works.

store/store.go
kshlm: Most of the methods in store can actually be unexported. I'll go through the package again.

store/remote.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/gluster/glusterd2/pkg/elasticetcd"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/coreos/etcd/clientv3"
+	"github.com/coreos/etcd/clientv3/concurrency"
+)
+
+func newRemoteStore(conf *Config) (*GDStore, error) {
+	if len(conf.Endpoints) == 1 && conf.Endpoints[1] == elasticetcd.DefaultEndpoint {
```
kshlm: I'm not sure if we'd have a lot of users trying to do this. I expect most users to either use the embedded elastic etcd or use the already available etcd clusters in kubernetes sort of deployments.

That doesn't mean we cannot support this.

Right now I need to ensure that the user has provided a proper endpoint, when GD2 is asked not to start the embedded store. This is the only way (I can think of right now), to check if the user has given proper endpoints that aren't the default. If not, we end up trying to connect to the default, and just stall for a long time (changing the timeout below will help though).

pkg/elasticetcd/leader.go
```diff
@@ -0,0 +1,269 @@
...
+				switch {
+				case err == context.Canceled:
+					return
+				default:
+					continue
+				}
+			}
+			ee.log.Debug("won election to become leader")
+			// Resign as leader if you cannot start the leader funtions
+			if err := ee.startLeader(); err != nil {
```
prashanthpai: startLeader() in its current form never errors out.

pkg/elasticetcd/leader.go
```diff
@@ -0,0 +1,269 @@
...
+
+	f := func(resp clientv3.WatchResponse) {
+		for _, ev := range resp.Events {
+			i, err := strconv.Atoi(string(ev.Kv.Value))
+			if err != nil {
+				ee.log.WithError(err).Error("could not parse idealsize value, ignoring update")
+				continue
+			}
+			if i != ee.conf.IdealSize {
+				ee.log.WithField("idealsize", i).Debug("idealsize changed, doing nominations again")
```
prashanthpai: Who can change the idealsize ? Is it only the newly joining member ? Or is there an API to change this during runtime ?

pkg/elasticetcd/doc.go
```diff
@@ -0,0 +1,24 @@
...
+//	- Connect to the given endpoints (or your own etcd endpoints if you were the first one up)
+//	- Volunteer to be a server, and wait to be nominated as a server
+//		- If you are nominated, start your embedded etcd server and join the existing cluster
+//		- If your nomination is removed, stop your embedded etcd server
+//	- Begin a campaign to become the leader of the elastic cluster
+// 		- When elected as the leader, make nominations from the volunteer list, to keep the right number of servers.
+// 		- Watch for changes to the volunteer list, online servers and the ideal size, and make/remove nominations as required.
+//
+// Right now the server nominations are selected in a round-robin fashion, using the list of volunteers sorted by name.
+//
```
kshlm: ElasticEtcd stop wasn't being done in a safe way. The watchers were left running, while the client was stopped. 
The client nomination would be removed as the session ended when it was stopped.
There was a brief moment where the nomination watcher was triggered before it was canceled, leading to the `doNominations` handler being run. This would panic when it tried to use the now nil `ee.cli`. This panic would prevent the etcd server from cleanly transferring the leadership and stopping, leading to an inconsistent cluster.

This is now fixed, with the watchers being stopped first before stopping the client.

pkg/elasticetcd/doc.go
```diff
@@ -0,0 +1,24 @@
+// Package elasticetcd implements an autoscaling etcd cluster.
+// This package was inspired by, and implements the elastic scaling algorithm as described in, the github.com/purpleidea/mgmt/etcd package.
+//
+// The elastic scaling algorithm is as follows,
+// 	- The elasticetcd instance is started, and can be passed a list of existing etcd endpoints
+//  	- If no endpoints are given, assume you are the first instance up and start your embedded etcd server
+//	- Connect to the given endpoints (or your own etcd endpoints if you were the first one up)
+//	- Volunteer to be a server, and wait to be nominated as a server
```
kshlm: Yes, all elasticetcd instances will volunteer to be a server. As a package, I would like elasticetcd to continue this way, and leave out the choice to be/not be a server to whoever uses the package. Like how the store package does in this PR. Without the server part, elasticetcd is just a etcd client, so you're better off just using etcd client directly. If elasticetcd were to become a standalone app, then that app could provide such a feature.

But, as mentioned in one of the TODOs below, I would also like to have the ability to use different server selection algorithms available for elasticetcd.

pkg/elasticetcd/leader.go
kshlm: startLeader() underwent a lot of refactors as I developed this. Somewhere along the line, the errors that were present just got moved out. I'll have a look to see if I can get errors propagated back here.

pkg/elasticetcd/leader.go
kshlm: I had forgotten about this. Right now, there isn't an API to change idealsize, but if required it can be changed directly in etcd. Everyone who can access this etcd cluster can change it right now.

We would need to protect access to the elastic namespace. Etcd v3 has users and roles, which can be used for this. I'll add a todo for this.

pkg/elasticetcd/leader.go
```diff
@@ -0,0 +1,269 @@
...
+	// its server.
+	// This is only required for the first nominee added. When more than one
+	// server is present the etcd cluster will serve requests when a member is
+	// added.
+	// TODO: Try to figure a way to avoid the sleeping
+
+	ee.addToNominees(host, urls)
+
+	time.Sleep(2 * time.Second)
+
```
kshlm: The sleep here kinda takes care of this as a side effect most of the times. Maybe we just leave it here forever.

pkg/elasticetcd/server.go
```diff
@@ -0,0 +1,125 @@
...
+		ee.log.Debug("initial cluster not given, setting to self")
+		conf.InitialCluster = conf.InitialClusterFromName(conf.Name)
+	}
+	return conf
+}
+
+func (ee *ElasticEtcd) initEtcdLogging() {
+	ee.server.logFile = new(nilWriteCloser)
+	if !ee.conf.DisableLogging {
+		f, err := os.OpenFile(path.Join(ee.conf.LogDir, "etcd.log"), os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
```
kshlm: I don't agree with aborting on failure here. Logging isn't the core function of this package, and I wouldn't want a failure here to stop the whole of GD2.

pkg/elasticetcd/volunteer.go
```diff
@@ -0,0 +1,71 @@
...
+		ee.log.WithError(err).Error("could not prepare nominees map")
+	}
+
+	// Check if you are in the nominees, and start/stop you embedded server as
+	// required
+	if _, ok := nominees[ee.conf.Name]; ok {
+		ee.log.Debug("nominated, starting server")
+		// Sleeping to allow leader to add me as a etcd cluster member
+		time.Sleep(2 * time.Second)
+		ee.startServer(nominees.String())
```
kshlm: `ee.startServer`/`ee.stopServer` check internally for the current state of the etcd server before doing their operations. Other than some unnecessary logs, doing this frequently shouldn't be a problem.

Nominations happen on changes to the volunteer list or idealsize. I don't expect these to change very fast. If you are already a server, changes to the volunteer list wouldn't affect you, as any existing online servers would still be volunteered. Your nomination can be removed only when idealsize is lowered. So unless someone keeps bumping an lowering idealsize repeatedly, it shouldn't be a problem as well.

I would like to see some sort of rate limitation for nominations though. I'll a TODO for this as well.

prashanthpai:
> @prashanthpai I'll need your help to fix e2e testing. It fails with a panic.

This should fix the panics: https://github.com/gluster/glusterd2/pull/330

kshlm:
@prashanthpai The func tests pass now. But as noted in #332, we have an issue with 2 node clusters.

I'll also be opening issues for the other TODOs and known issues in this PR.

PR #331, opened 2017-06-29
Enable functional tests

kshlm:


prashanthpai:
@kshlm  This will require tests to be always run as root. That ok ?

kshlm:
> @kshlm This will require tests to be always run as root. That ok ?

Gluster needs to run as root to function, so it's okay if functional tests need to run as root.

PR #333, opened 2017-07-03
Add an RPM specfile

kshlm:


scripts/build.sh
```diff
@@ -13,6 +13,7 @@ fi
 GOBUILD_TAGS="novirt noaugeas "
 VERSION=$($(dirname $0)/pkg-version --full)
 LDFLAGS="-X github.com/gluster/glusterd2/gdctx.GlusterdVersion=$VERSION"
+LDFLAGS+=" -B 0x$(head -c20 /dev/urandom | od -An -tx1 | tr -d ' \n')"
```
purpleidea: AFAICT, this would make the builds un-reproducible. As a result, maybe best to avoid doing this. More info here: https://reproducible-builds.org/

kshlm: Fedora requires this though.

purpleidea:
On Mon, Jul 3, 2017 at 9:55 AM, Kaushal M <notifications@github.com> wrote:

> Fedora requires this though.



Can you provide more info about what this is and why?

kshlm:
> Can you provide more info about what this is and why?

https://fedoraproject.org/wiki/Releases/FeatureBuildId has some more info. Rpmbuild fails if the generated binaries are not tagged with a build-id.

purpleidea:
On Mon, Jul 3, 2017 at 10:03 AM, Kaushal M <notifications@github.com> wrote:

> https://fedoraproject.org/wiki/Releases/FeatureBuildId has some more
> info. Rpmbuild fails if the generated binaries are not tagged with a
> build-id.



A buildid is fine, but it should not be random. It should probably be based
off of the git sha1 of the commit.

kshlm:
> A buildid is fine, but it should not be random. It should probably be based
> off of the git sha1 of the commit.

Probably. But rpms are built from tarballs, not directly from git. I don't know how I could pass a git sha1 there.

purpleidea:
On Mon, Jul 3, 2017 at 10:08 AM, Kaushal M <notifications@github.com> wrote:

> Probably. But rpms are built from tarballs, not directly from git. I don't
> know how I could pass a git sha1 there.



I'm not an RPM expert, so I don't know, sorry. Just saw this and figured
you'd like to know. I'd recommend pinging the RPM pro's on IRC.

scripts/build.sh
prashanthpai: @purpleidea I was also looking for this sometime back and Go [doesn't officially support](https://github.com/golang/go/issues/16860) reproducible builds yet.

@kshlm We should base build id based on git sha-1 eventually. https://fedoraproject.org/wiki/PackagingDrafts/Go#Build_ID

extras/rpms/glusterd2.spec
```diff
@@ -0,0 +1,81 @@
...
+%setup -q -n %{name}-%{commit}
+
+%build
+mkdir -p src/%(dirname %{import_path})
+ln -s ../../../ src/%{import_path}
+
+pushd src/%{import_path}
+# Install vendored packages
+# TODO: See if we can build with unbundled packages
+make vendor-install
```
aravindavk: I think this should be as part of make dist. Good if we can avoid network calls during RPM/packaging create.

This is fine for current release, we can take care of this issue later.

kshlm: This is a good idea. I'll do this next time. Makes is easier to do a release.

Issue #205, opened 2016-12-27
CLI


kshlm:
We need to provide a CLI application for managing Gluster. The CLI app will be a consumer of the ReST API.

aravindavk:
Since `glusterd2` CLI is now REST client, it can be used to manage or get information from multiple Clusters.

User/Auth management part is TBD. We may have to add `.env` or additional parameter for CLI command to accept Username and Secret

For Example(TBD - This will change depending on Auth we support):

    gluster -h <HOST> -u <USERNAME> -p <SECRET>  # This opens gluster shell `gluster>`
    gluster -h <HOST> -u <USERNAME>              # This prompts for Secret and then opens gluster shell
    gluster -h <HOST> -u <USERNAME> -p <SECRET> <CMD> <ARGS> # Same as `glusterd1` CLI but additionally takes HOST, USERNAME and SECRET parameters

We should take this opportunity to fix/improve some of the issues we currently face with existing CLI commands.

- In `glusterd1` optional arguments are parsed as positional arguments, we should change this in `glusterd2`. For example, in volume create command Use `--replica 2` instead of `replica 2`. This eliminates lot of parsing overhead and bugs during parsing.
- Split force argument into meaningful flags. For Example, `--allow-root-dir`, `--reuse-bricks` during Volume Create(REST APIs should be enhanced to make this work)
- Depricate XML output and support JSON?
- Which is better? `volume info` or `volume-info`(First option may need extra work to parse arguments, for second option can take advantage of existing parsing library)

Please add if any other things to consider while designing `glusterd2` CLIs.

kshlm:
Going from first to last in the bullets

- +1 from me. This makes it easier to parse flags for each command.
- +1
- As the rest api itself provides JSON, cli should only need to provide the
raw responses if requested.
- there are packages that provide this behaviour. So we should be okay with
'gluster volume info' style of commands.


On 28 Dec 2016 15:21, "Aravinda VK" <notifications@github.com> wrote:

Since glusterd2 CLI is now REST client, it can be used to manage or get
information from multiple Clusters.

User/Auth management part is TBD. We may have to add .env or additional
parameter for CLI command to accept Username and Secret

For Example(TBD - This will change depending on Auth we support):

gluster -h <HOST> -u <USERNAME> -p <SECRET>  # This opens gluster
shell `gluster>`
gluster -h <HOST> -u <USERNAME>              # This prompts for Secret
and then opens gluster shell
[...]

PR #334, opened 2017-07-04
cli: Initial patch for Glusterd2 CLI

aravindavk:
Updates: #205
Signed-off-by: Aravinda VK <mail@aravindavk.in>

### DIFF
--- a/cli/cmd/peer.go
+++ b/cli/cmd/peer.go
@@ -49,7 +49,7 @@ var poolCmd = &cobra.Command{
 
 var peerProbeCmd = &cobra.Command{
 	Use:   "probe <HOSTNAME>",
-	Short: helpPeerCmd,
+	Short: helpPeerProbeCmd,
 	Run: func(cmd *cobra.Command, args []string) {
 		validateNArgs(cmd, 1, 1)
 		hostname := cmd.Flags().Args()[0]
--- a/cli/cmd/peer.go
+++ b/cli/cmd/peer.go
@@ -1,0 +1,113 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/olekukonko/tablewriter"
+	"github.com/spf13/cobra"
+)
+
+const (
+	helpPeerCmd       = "Gluster Peer Management"
+	helpPeerProbeCmd  = "probe peer specified by <HOSTNAME>"
+	helpPeerDetachCmd = "detach peer specified by <HOSTNAME>"
+	helpPeerStatusCmd = "list status of peers"
+	helpPoolListCmd   = "list all the nodes in the pool (including localhost)"
+)
+
+var (
+	// Peer Detach Command Flags
+	flagPeerDetachForce bool
+)
+
+func init() {
+	peerCmd.AddCommand(peerProbeCmd)
+
+	peerDetachCmd.Flags().BoolVarP(&flagPeerDetachForce, "force", "f", false, "Force")
+	peerCmd.AddCommand(peerDetachCmd)
+
+	peerCmd.AddCommand(peerStatusCmd)
+
+	poolCmd.AddCommand(poolListCmd)
+
+	RootCmd.AddCommand(peerCmd)
+	RootCmd.AddCommand(poolCmd)
+}
+
+var peerCmd = &cobra.Command{
+	Use:   "peer",
+	Short: helpPeerCmd,
+}
+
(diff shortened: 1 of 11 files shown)


aravindavk:
Still WIP.

aravindavk:
Implemented Peer probe and peer detach. All others(peer status, pool list, volume ops) are just place holder now.

Looks like Peer detach API accepts ID instead of fqdn, that is not taken care.
Cleanup required in Error and success messages.
Makefile changes are remaining for cli dependency handling and other checks

Note: This depends on restclient patch.

cli/cmd/peer.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+var poolCmd = &cobra.Command{
+	Use:   "pool",
+	Short: helpPeerCmd,
+}
+
+var peerProbeCmd = &cobra.Command{
+	Use:   "probe <HOSTNAME>",
+	Short: helpPeerCmd,
```
kshlm: Should be `helpPeerProbeCmd`.

cli/cmd/root.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+var (
+	flagXMLOutput  bool
+	flagJSONOutput bool
+)
+
+func init() {
+	initRESTClient()
+	initLog("./cli.log", "INFO")
```
kshlm: This should be configurable from now itself.

cli/main.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/gluster/glusterd2/cli/cmd"
+)
+
+func main() {
+	// Migrate old format Args into new Format. Modifies os.Args()
```
kshlm: Is there a way to get the log initialized here before we do anything else?

Right now, it's being initialized during package initialization of `cmd`, which technically happens before this point, but it doesn't allow for much configurability.

cli/cmd/root.go
prashanthpai: Also, please check the return value for error.

cli/cmd/volume.go
```diff
@@ -0,0 +1,168 @@
...
+		volname := cmd.Flags().Args()[0]
+		fmt.Println("START:", volname)
+	},
+}
+
+var volumeStopCmd = &cobra.Command{
+	Use:   "stop",
+	Short: helpVolumeStopCmd,
+	Run: func(cmd *cobra.Command, args []string) {
+		validateNArgs(cmd, 1, 1)
```
prashanthpai: I so miss Python's decorators :)

cli/cmd/volume.go
```diff
@@ -0,0 +1,168 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+)
+
+const (
+	helpVolumeCmd       = "Gluster Volume Management"
```
prashanthpai: As purely a personal preference, unless these were used at multiple places, I'd place them in the declaration of `cobra.Command` itself.

scripts/build-cli.sh
```diff
@@ -0,0 +1,27 @@
...
+  OUTDIR=$1
+fi
+
+GOBUILD_TAGS="novirt noaugeas "
+VERSION=$($(dirname $0)/pkg-version --full)
+REPO_PATH="github.com/gluster/glusterd2"
+GIT_SHA=`git rev-parse --short HEAD || echo "undefined"`
+LDFLAGS="-X ${REPO_PATH}/version.GlusterdVersion=$VERSION -X ${REPO_PATH}/version.GitSHA=$GIT_SHA"
+LDFLAGS+=" -B 0x$(head -c20 /dev/urandom | od -An -tx1 | tr -d ' \n')"
+BIN=gluster
```
aravindavk: Makes sense. will change it to `glustercli`

cli/cmd/volume.go
aravindavk: Placed it separately only to facilitate easy translation of these messages in future.

cli/main.go
aravindavk: I moved initialization code before cmd initialization. Not yet made configurable. 

With the help of `PersistentPreRun` option in `cobra` I can initialize log there, but it is not possible to initialize log file before parsing the args. (Logger will not be available during Args Migrate). I need to write custom parser to parse `--log-file` and `--log-level` outside the `cobra.Command` initialization.

I will address this with separate patch, as of now added constants for the same.

I will address this in separate patch.

Issue #265, opened 2017-02-27
Management API package - REST API SDKs


prashanthpai:
This management API package should be similar to libgfapi but for management. For now, we can focus on go bindings that allows access to glusterfs cluster management operations. This API can be independently used y external consumers and internally by our own CLI implementation.

kshlm:
Requires #251, and will be a part of #205

aravindavk:
Restclient API design is getting complicated than expected, following alternatives shows the issues to use with CLI. 

## Alternative 1:

Restclient always returns object, CLI/consumer can marshal it into whatever format

Example:

    func PeerStatus() ([]Peer, error)

This is good API design, only problem is that it is doing extra Marshal and Unmarshal
which is not required. CLI calls this API, restclient gets JSON output from
server and then unmarshals into respective object. CLI consumes this and again
marshals it to get JSON output.(`gluster peer status --json`)

If we follow this approach, then we should be careful while using multiple
fields in struct for the same entity or derived fields. For example,

    package main
     
    import (
        "encoding/json"
        "encoding/xml"
     
        "fmt"
    )
     
    type Peer struct {
        ID            string `xml:"id" json:"id"`
        Name          string `xml:"name" json:"name"`
        Connected     bool   `xml:"-" json:"connected"`
        ConnectedFlag int    `xml:"connected" json:"-"`
    }
     
    func updatePeerStatus(peer *Peer, connected bool) {
        peer.ConnectedFlag = 0
        if connected {
        	peer.Connected = connected
        	peer.ConnectedFlag = 1
        }
    }
     
    func main() {
        peer := Peer{ID: "p1", Name: "Peer 1"}
        updatePeerStatus(&peer, true)
     
        fmt.Println("Print object")
        fmt.Println(peer)
     
        fmt.Println("Direct JSON output")
        data1, _ := json.Marshal(&peer)
        fmt.Println(string(data1))
     
        fmt.Println("Direct XML output")
        data2, _ := xml.Marshal(&peer)
        fmt.Println(string(data2))
     
        fmt.Println("XML output(JSON->Object->XML)")
        var p Peer
        xml.Unmarshal(data1, &p)
        data3, _ := xml.Marshal(&p)
        fmt.Println(string(data3))
    }

Output shows that we got wrong value for `<connected>` tag in last XML output.

    Print object
    {p1 Peer 1 true 1}
    Direct JSON output
    {"id":"p1","name":"Peer 1","connected":true}
    Direct XML output
[...]

PR #337, opened 2017-07-18
restclient: Golang REST Client library

aravindavk:
REST Client for Glusterd2 APIs. Initially these APIs will be used by
CLI project, it can be used by any application which is written in Go
lang.

Example:

    package main

    import (
        "fmt"
        "github.com/gluster/glusterd2/pkg/restclient"
    )

    func main(){
        client := restclient.NewRESTClient("http://localhost:24007", "", "")
        err := client.VolumeStart("gv1")
        if err != nil{
                fmt.Println("Failed to Start Volume gv1")
        } else{
                fmt.Println("Started Volume gv1 successfully")
        }
    }

Signed-off-by: Aravinda VK mail@aravindavk.in
Updates: #265

### DIFF
--- a/e2e/peer_ops_test.go
+++ b/e2e/peer_ops_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/gluster/glusterd2/pkg/api"
 	"github.com/stretchr/testify/require"
 
 	"github.com/gluster/glusterd2/peer"
@@ -36,8 +37,13 @@ func TestAddRemovePeer(t *testing.T) {
 	r.True(g3.IsRunning())
 
 	// add peer: ask g1 to add g2 as peer
-	reqBody := strings.NewReader(fmt.Sprintf(`{"addresses": ["%s"]}`, g2.PeerAddress))
-	resp, err := http.Post("http://"+g1.ClientAddress+"/v1/peers", "application/json", reqBody)
+	peerAddReq := api.PeerAddReq{
+		Addresses: []string{g2.PeerAddress},
+	}
+	reqBody, err := json.Marshal(peerAddReq)
+	r.Nil(err)
+
+	resp, err := http.Post("http://"+g1.ClientAddress+"/v1/peers", "application/json", strings.NewReader(string(reqBody)))
 	r.Nil(err)
 	defer resp.Body.Close()
 	r.Equal(http.StatusCreated, resp.StatusCode)
--- a/e2e/peer_ops_test.go
+++ b/e2e/peer_ops_test.go
@@ -1,18 +1,10 @@
 package e2e
 
 import (
-	"encoding/json"
-	"fmt"
-	"io/ioutil"
-	"net/http"
-	"strings"
 	"testing"
 	"time"
 
-	"github.com/gluster/glusterd2/pkg/api"
 	"github.com/stretchr/testify/require"
-
-	"github.com/gluster/glusterd2/peer"
 )
 
 func TestAddRemovePeer(t *testing.T) {
@@ -36,48 +28,25 @@ func TestAddRemovePeer(t *testing.T) {
 	defer g3.EraseWorkdir()
 	r.True(g3.IsRunning())
 
-	// add peer: ask g1 to add g2 as peer
-	peerAddReq := api.PeerAddReq{
-		Addresses: []string{g2.PeerAddress},
-	}
-	reqBody, err := json.Marshal(peerAddReq)
-	r.Nil(err)
+	client := initRestclient(g1.ClientAddress)
 
-	resp, err := http.Post("http://"+g1.ClientAddress+"/v1/peers", "application/json", strings.NewReader(string(reqBody)))
-	r.Nil(err)
-	defer resp.Body.Close()
-	r.Equal(http.StatusCreated, resp.StatusCode)
+	_, err2 := client.PeerProbe(g2.PeerAddress)
+	r.Nil(err2)
 
 	time.Sleep(5 * time.Second)
 
 	// add peer: ask g1 to add g3 as peer
-	reqBody = strings.NewReader(fmt.Sprintf(`{"addresses": ["%s"]}`, g3.PeerAddress))
-	resp, err = http.Post("http://"+g1.ClientAddress+"/v1/peers", "application/json", reqBody)
-	r.Nil(err)
-	defer resp.Body.Close()
-	r.Equal(http.StatusCreated, resp.StatusCode)
+	_, err3 := client.PeerProbe(g3.PeerAddress)
+	r.Nil(err3)
 
 	time.Sleep(5 * time.Second)
 
 	// list and check you have 3 peers in cluster
-	resp, err = http.Get("http://" + g1.ClientAddress + "/v1/peers")
-	r.Nil(err)
-	defer resp.Body.Close()
-	r.Equal(http.StatusOK, resp.StatusCode)
-
-	data, err := ioutil.ReadAll(resp.Body)
-	r.Nil(err)
-	var peers []peer.Peer
-	err = json.Unmarshal(data, &peers)
-	r.Nil(err)
+	peers, err4 := client.Peers()
+	r.Nil(err4)
 	r.Len(peers, 3)
 
 	// remove peer: ask g1 to remove g2 as peer
-	delURL := fmt.Sprintf("http://%s/v1/peers/%s", g1.ClientAddress, g2.PeerID())
-	req, err := http.NewRequest("DELETE", delURL, nil)
-	r.Nil(err)
-	resp, err = http.DefaultClient.Do(req)
-	r.Nil(err)
-	defer resp.Body.Close()
-	r.Equal(http.StatusNoContent, resp.StatusCode)
+	err5 := client.PeerDetachByID(g2.PeerID())
+	r.Nil(err5)
 }
--- a/e2e/restart_test.go
+++ b/e2e/restart_test.go
@@ -1,14 +1,11 @@
 package e2e
 
 import (
-	"encoding/json"
-	"fmt"
 	"io/ioutil"
-	"net/http"
 	"os"
-	"strings"
 	"testing"
 
+	"github.com/gluster/glusterd2/pkg/api"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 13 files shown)


pkg/restclient/examples/main.go
```diff
@@ -0,0 +1,28 @@
+package main
+
+import (
+	"fmt"
+	"github.com/gluster/glusterd2/restclient"
```
prashanthpai: As you may already know, this is causing the CI failure. It has to be `github.com/gluster/glusterd2/pkg/restclient`

pkg/restclient/common.go
```diff
@@ -0,0 +1,62 @@
...
+	if respType == "xml" || respType == "json" {
+		req.Header.Set("Accept", "application/"+respType)
+	}
+	req.Header.Set("Content-Type", "application/json")
+	resp, err1 := http.DefaultClient.Do(req)
+	if err1 != nil {
+		return "", err1
+	}
+
+	if resp.StatusCode == 404 {
```
prashanthpai: 404 are also returned for genuine errors such as volume not found or peer not found and doesn't only represent invalid endpoint.

prashanthpai: For now, it'll be better to define request success based on the response status code being returned in `2xx` series and failure as any status code returned in `4xx` (client error) and `5xx` (server error) series.

aravindavk: Idea behind this is to handle not implemented APIs smoothly in CLI, For example Same CLI can work with multiple versions of Glusterd2, if one Glusterd2(Cluster) doesn't have a API implemented then CLI can handle it. I think this check can be moved to CLI itself instead of having it in restclient.

pkg/restclient/volume.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/gluster/glusterd2/pkg/api"
+)
+
+// VolumeCreate creates Gluster Volume
+func (c *RESTClient) VolumeCreate(volname string, bricks []string, replica int, force bool) error {
```
aravindavk: Makes sense. Let me think about using VolumeCreateOptions struct or something similar for optional parameters.

aravindavk:
> I'm not really sure how useful restclient package will be for external programs other than the CLI

- Automation scripts
- Alternative CLIs
- Easy integration with external Golang programs
- In this patch, connection related things are not yet implemented, once that is available it adds more value to consume APIs programatically from external nodes.

pkg/restclient/peer.go
```diff
@@ -0,0 +1,27 @@
...
+// PeerProbe adds a peer to the Cluster
+func (c *RESTClient) PeerProbe(host string) error {
+	peerAddReq := api.PeerAddReq{
+		Addresses: []string{host},
+	}
+	reqBody, err := json.Marshal(peerAddReq)
+	if err != nil {
+		return err
+	}
+	return httpRESTAction("POST", c.baseURL+"/v1/peers", strings.NewReader(string(reqBody)), 201)
```
kshlm: General comment for this and the rest of the client methods that have been implemented.

You could do much more stuff inside `httpRestAction`.
- If you pass in the `RestClient` or make `httpRestAction` a method on `RestClient` you can form the URL inside, instead of concatenating here.
- You could just pass the request struct in here, and do the marshalling in `httpRESTAction`.
- You can avoid passing the expected code, you can have a map of action and expected code and use it instead.

Also, use the `http.Status*` enums for status codes, instead of the values directly.

pkg/restclient/common.go
```diff
@@ -0,0 +1,73 @@
+package restclient
```
kshlm: Instead of calling this package, you can just call it `gd2client` or just `client`. We don't expect there to be any other type of GD2 client package. 
This will make more sense once we restructure the source tree, then this package would appear as `glusterd2/rest/client`.

pkg/restclient/common.go
```diff
@@ -0,0 +1,73 @@
...
+	"encoding/json"
+	"io"
+	"io/ioutil"
+	"net/http"
+)
+
+const baseURL = "http://localhost:24007"
+
+// RESTClient represents Glusterd2 REST Client
+type RESTClient struct {
```
kshlm: Just `Client` is a good name for this. The doc comment anyway specifies what sort of a client this is.

pkg/restclient/common.go
```diff
@@ -0,0 +1,73 @@
...
+	username string
+	password string
+}
+
+type httpError struct {
+	Error string `json:"Error"`
+}
+
+// NewRESTClient creates new instance of RESTClient
+func NewRESTClient(baseURL string, username string, password string) *RESTClient {
```
kshlm: This can just be `New`, so that users will be calling this as `client.New()` instead of `restclient.NewRESTClient`. Still makes the same amount of sense.

pkg/restclient/common.go
```diff
@@ -0,0 +1,73 @@
...
+	if err2 != nil {
+		return "", err2
+	}
+	if resp.StatusCode != expectStatusCode {
+		return "", raiseAPIUnexpectedStatusCodeError(expectStatusCode, resp.StatusCode, parseHTTPError(output))
+	}
+	return parseHTTPError(output), nil
+}
+
+func httpRESTAction(method string, url string, body io.Reader, expectStatusCode int) error {
```
kshlm: This could become a method on the `RestClient` called `Do` along the lines of [`net/http.(*Client).Do`](https://golang.org/pkg/net/http/#Client.Do).

And as mentioned else where, you could also form the full URL and marshal the body in here.

pkg/restclient/errors.go
```diff
@@ -0,0 +1,23 @@
...
+// APIUnexpectedStatusCodeError is custom error when expected
+// status code does not match with return status code
+type APIUnexpectedStatusCodeError struct {
+	msg      string
+	expected int
+	actual   int
+	resp     string
+}
+
+func raiseAPIUnexpectedStatusCodeError(expected int, actual int, resp string) error {
```
kshlm: This function isn't actually required. The `msg` is static and isn't actually used when forming the error string. In the one place this is actually being called, you can just create an in-place `UnexpectedStatusError` object.

pkg/restclient/errors.go
```diff
@@ -0,0 +1,23 @@
+package restclient
+
+import (
+	"fmt"
+)
+
+// APIUnexpectedStatusCodeError is custom error when expected
+// status code does not match with return status code
+type APIUnexpectedStatusCodeError struct {
```
kshlm: Name can be simplified, to possibly `UnexpectedStatusError`.

e2e/peer_ops_test.go
```diff
@@ -25,8 +27,13 @@ func TestAddRemovePeer(t *testing.T) {
...
 	// add peer: ask g1 to add g2 as peer
-	reqBody := strings.NewReader(fmt.Sprintf(`{"addresses": ["%s"]}`, g2.PeerAddress))
-	resp, err := http.Post("http://"+g1.ClientAddress+"/v1/peers", "application/json", reqBody)
+	peerAddReq := api.PeerAddReq{
+		Addresses: []string{g2.PeerAddress},
+	}
+	reqBody, err := json.Marshal(peerAddReq)
+	r.Nil(err)
+
+	resp, err := http.Post("http://"+g1.ClientAddress+"/v1/peers", "application/json", strings.NewReader(string(reqBody)))
```
aravindavk: This patch is not meant to change the e2e behavior. Only changes related to pkg/api are changed here. I will send separate patch to fix e2e tests using restclient.

pkg/restclient/errors.go
aravindavk: Ok

pkg/restclient/errors.go
```diff
@@ -0,0 +1,23 @@
...
+	resp     string
+}
+
+func raiseAPIUnexpectedStatusCodeError(expected int, actual int, resp string) error {
+	return &APIUnexpectedStatusCodeError{"Unexpected Status Code", expected, actual, resp}
+}
+
+func (e *APIUnexpectedStatusCodeError) Error() string {
+	msg := fmt.Sprintf("Status Code: %d Error: %s", e.actual, e.resp)
+	return msg
```
aravindavk: I will add it.

pkg/restclient/peer.go
aravindavk: All points makes sense.

> You can avoid passing the expected code, you can have a map of action and expected code and use it instead.

Expected code may differ from request to request, keeping map of action is not generic. Plugins restclient may expect different status codes, which can't be added to map.

pkg/restclient/common.go
aravindavk: I prefer having all the external facing libraries under `glusterd2/pkg`.
With this patch I will keep it as `glusterd2/pkg/restclient`, we can change this in future if required.

pkg/restclient/common.go
aravindavk: Ok

pkg/restclient/common.go
aravindavk: Ok

pkg/restclient/common.go
aravindavk: It is different from http.Do, I was trying to map it to Action/Command in Gluster which doesn't require to parse the response body. Only depends on the status code to decide success or failure. (Doesn't return response)

aravindavk:
Do not merge.

Realized that we are doing many things ID based, So even action command outputs are important to parse and report it back as . (Volume Create, Peer probe etc). Other commands/actions need not parse the HTTP response, only return code check is enough. For example, Volume start.

I will make that change and refresh this pull request.

e2e/restart_test.go
```diff
@@ -20,23 +17,21 @@ func TestRestart(t *testing.T) {
...
-	r.Equal(http.StatusCreated, resp.StatusCode)
+	createReq := api.VolCreateReq{
+		Name: "vol1",
+		Bricks: []string{
+			gd.PeerAddress + ":" + dir,
+		},
+		Force: true,
+	}
+	_, errVolCreate := client.VolumeCreate(createReq)
+	r.Nil(errVolCreate)
```
aravindavk: Behavior is very confusing in Go :(

Do you have any doc link which explains this better?

    resp, err := SomeFunc()
    resp, err := SomeOtherFunc()  # Not valid
    resp, err1 := SomeOtherFunc()  # Valid since one new variable
    resp, err = SomeOtherFunc()  # Valid
    resp1, err = SomeOtherFunc() # Not Valid, resp1 is new variable

prashanthpai: Sometimes you can declare `var err error` at the beginning of the function.

PR #338, opened 2017-07-18
Explicitly add git SHA and other info in version output

prashanthpai:
...and move version information to its own package

Example output:
```sh
# glusterd2 --version
glusterd version: v4.0dev-7.1.gitb5a02a5
git SHA: b5a02a5
go version: go1.8
go OS/arch: linux/amd64
```


prashanthpai:
> We still need to figure out how to pass this information for RPM (and other packages). Packages don't build directly from git, so you cannot get the SHA during those builds.

We can write a shell script to generate the source tarball that'll be used in building source RPM. The tarball can be made to contain a text file with git SHA. This text file will be read during the build process. You [already have](https://github.com/gluster/glusterd2/blob/master/scripts/pkg-version#L4) something similar today for version.

kshlm:
> We can write a shell script to generate the source tarball that'll be used in building source RPM. The tarball can be made to contain a text file with git SHA. This text file will be read during the build process. You already have something similar today for version.

I had forgotten that the version script could read from file. That makes things easier.

PR #339, opened 2017-07-20
Read options from non-xlators too

prashanthpai:
There are shared objects such as socket.so in 'rpc-transport' and
'auth' directories. Load them too.
    
    NOTE:
    This also depends on https://review.gluster.org/#/c/17753/ which
    means going forward, glusterd2 will need gluster version installed
    from source (master branch) until 3.12 branches out and becomes
    available as an RPM.


kshlm:
@prashanthpai, would it possible to have a way to turn this off till 3.12 is released?

prashanthpai:
> @prashanthpai, would it possible to have a way to turn this off till 3.12 is released?

Yes, I just have to comment out a line for now. I'll do that.

PR #341, opened 2017-07-26
Use proper store config when restarting store during probe

kshlm:
Instead of using the provided store config values, store was being
restarted with the default store config. This prevented GD2 from working
when non-default config is used.

A test has been added to the e2e suite to check this.

### DIFF
--- a/e2e/peer_ops_test.go
+++ b/e2e/peer_ops_test.go
@@ -29,7 +29,7 @@ func TestAddRemovePeer(t *testing.T) {
 	defer g2.EraseWorkdir()
 	r.True(g2.IsRunning())
 
-	g3, err := spawnGlusterd("./config/3.yaml")
+	g3, err := spawnGlusterd("./config/3.yaml", true)
 	r.Nil(err)
 	defer g3.Stop()
 	defer g3.EraseWorkdir()
--- a/e2e/peer_ops_test.go
+++ b/e2e/peer_ops_test.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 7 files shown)


kshlm:
@ppai, what actually happens for TestAddRemovePeer for you? Does the test fail out right or does it hang?

The hang happens because of the know issue with peer detach in 2 node cluster. A quick solution for that would be to add another peer to cluster.

prashanthpai:
> what actually happens for TestAddRemovePeer for you? Does the test fail out right or does it hang?

It never fails. It hangs (exactly like how it does in manual testing). And from the print statements that I put in, I know that it's hung on list peer op and not remove peer in the TestAddRemovePeer test.

kshlm:
@prashanthpai Is this still not working for you? I'd like to get this in quickly.

kshlm:
@prashanthpai I believe I've finally solved this for good. There were a couple more fundamental bugs, that have been fixed.

I was finally able to hit what you did today, and was able to debug it better. I have no idea how or why this worked for till now, as clustering shouldn't have worked with the bugs.

e2e/peer_ops_test.go
```diff
@@ -24,12 +29,42 @@ func TestAddRemovePeer(t *testing.T) {
...
+	// add peer: ask g1 to add g3 as peer
+	reqBody = strings.NewReader(fmt.Sprintf(`{"addresses": ["%s"]}`, g3.PeerAddress))
+	resp, err = http.Post("http://"+g1.ClientAddress+"/v1/peers", "application/json", reqBody)
+	r.Nil(err)
+	defer resp.Body.Close()
+	r.Equal(http.StatusCreated, resp.StatusCode)
+
+	time.Sleep(5 * time.Second)
+
+	// list and check you have 2 peers in cluster
```
prashanthpai: nit: `3` peers

e2e/peer_ops_test.go
```diff
@@ -24,12 +29,42 @@ func TestAddRemovePeer(t *testing.T) {
...
+
 	// add peer: ask g1 to add g2 as peer
 	reqBody := strings.NewReader(fmt.Sprintf(`{"addresses": ["%s"]}`, g2.PeerAddress))
 	resp, err := http.Post("http://"+g1.ClientAddress+"/v1/peers", "application/json", reqBody)
 	r.Nil(err)
 	defer resp.Body.Close()
-	r.Equal(resp.StatusCode, 201)
+	r.Equal(http.StatusCreated, resp.StatusCode)
+
+	time.Sleep(5 * time.Second)
```
prashanthpai: Do we need this sleep ?

kshlm: This is just to allow elasticetcd to properly form the etcd cluster. We need to add a cleaner method to detect this, and once we do that, these sleeps can be removed.

PR #345, opened 2017-07-27
Add volume set API

prashanthpai:
...and allow setting volume options during volume create as well

### DIFF
--- a/commands/volumes/common.go
+++ b/commands/volumes/common.go
@@ -1,0 +1,45 @@
+package volumecommands
+
+import (
+	"strings"
+
+	"github.com/gluster/glusterd2/xlator"
+)
+
+func areOptionNamesValid(optsFromReq map[string]string) bool {
+
+	var xlOptFound bool
+	for o := range optsFromReq {
+
+		// assuming option to be of the form <domain>.<xlator-option>
+		// and <domain> will be the xlator type.
+		// Example: cluster/afr.eager-lock
+		// we know for certain that this isn't true
+
+		tmp := strings.Split(strings.TrimSpace(o), ".")
+		if len(tmp) != 2 {
+			return false
+		}
+		xlatorType := tmp[0]
+		xlatorOption := tmp[1]
+
+		options, ok := xlator.AllOptions[xlatorType]
+		if !ok {
+			return false
+		}
+
+		xlOptFound = false
+		for _, option := range options {
+			for _, key := range option.Key {
+				if xlatorOption == key {
+					xlOptFound = true
+				}
+			}
+		}
+		if !xlOptFound {
+			return false
+		}
+	}
+
+	return true
+}
--- a/commands/volumes/common.go
+++ b/commands/volumes/common.go
@@ -1,9 +1,18 @@
 package volumecommands
 
 import (
+	"os"
 	"strings"
 
+	"github.com/gluster/glusterd2/gdctx"
+	"github.com/gluster/glusterd2/servers/sunrpc"
+	"github.com/gluster/glusterd2/transaction"
+	"github.com/gluster/glusterd2/utils"
+	"github.com/gluster/glusterd2/volgen"
+	"github.com/gluster/glusterd2/volume"
 	"github.com/gluster/glusterd2/xlator"
+
+	"github.com/pborman/uuid"
 )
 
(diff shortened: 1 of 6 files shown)


commands/volumes/volume-create.go
```diff
@@ -3,26 +3,24 @@ package volumecommands
...
-	Transport    string   `json:"transport,omitempty"`
-	ReplicaCount int      `json:"replica,omitempty"`
-	Bricks       []string `json:"bricks"`
-	Force        bool     `json:"force,omitempty"`
+	Name         string            `json:"name"`
+	Transport    string            `json:"transport,omitempty"`
+	ReplicaCount int               `json:"replica,omitempty"`
+	Bricks       []string          `json:"bricks"`
+	Force        bool              `json:"force,omitempty"`
+	Options      map[string]string `json:"options,omitempty"`
```
aravindavk: Options can be configured in Volume create itself?

commands/volumes/common.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+		if !xlOptFound {
+			return false
+		}
+	}
+
+	return true
+}
+
+func generateVolfiles(c transaction.TxnCtx) error {
```
aravindavk: This is moved from other file right? I am skipping review of this below functions in this file, is that fine?

commands/volumes/volume-create.go
prashanthpai: Yes. I can't think of a reason why it shouldn't be. What do you think ?

commands/volumes/common.go
prashanthpai: Yes, it's moved. That's fine 👍

commands/volumes/volume-option.go
```diff
@@ -0,0 +1,138 @@
...
+
+func volumeOptionsHandler(w http.ResponseWriter, r *http.Request) {
+
+	p := mux.Vars(r)
+	volname := p["volname"]
+	reqID, logger := restutils.GetReqIDandLogger(r)
+
+	volinfo, err := volume.GetVolume(volname)
+	if err != nil {
+		restutils.SendHTTPError(w, http.StatusNotFound, errors.ErrVolNotFound.Error())
```
aravindavk: 404 error is confusing. "400 Bad Request" can be used here I think since invalid Volume name is given

commands/volumes/volume-create.go
aravindavk: Cool, it is added advantage to accept volume create and options together.

commented just for clarification.

commands/volumes/common.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+func areOptionNamesValid(optsFromReq map[string]string) bool {
+
+	var xlOptFound bool
+	for o := range optsFromReq {
+
+		// assuming option to be of the form <domain>.<xlator-option>
+		// and <domain> will be the xlator type.
+		// Example: cluster/afr.eager-lock
```
kshlm: We should in the future allow options with just the xlator name as the domain, ie. allow `afr.eager-lock` as well instead of just `cluster/afr.eager-lock`, to make this a little more user friendly.

commands/volumes/volume-option.go
```diff
@@ -0,0 +1,138 @@
...
+		return
+	}
+
+	var req VolOptionRequest
+	if err := utils.GetJSONFromRequest(r, &req); err != nil {
+		restutils.SendHTTPError(w, http.StatusUnprocessableEntity, errors.ErrJSONParsingFailed.Error())
+		return
+	}
+
+	if !areOptionNamesValid(req.Options) {
```
kshlm: It'll be better here if you could log the invalid option, and also send it back to the user. Do the same also in volume create.
An explicit InvalidOptionError type will be useful here.

I'll hold merging this PR, till this is addressed.

commands/volumes/common.go
prashanthpai: Can be done in future. The only concern is conflicting names. For example, `cache-size` can belong to any of the perf xlators.

kshlm: You would still have the xlator id being prefixed. But just not the full id. So `performace/md-cache.cache-size` would be also allowed as `md-cache.cache-size`.

Issue #332, opened 2017-06-29
Unhealthy cluster after removing a peer in 2 node cluster


kshlm:
(Only valid once #324 is merged)
Deleting the 2nd peer in a 2 node cluster will leave the 1st node unresponsive. This happens when using the embedded elastic store.

This happens because the etcd instance is left without quorum to operate when the 2nd peer leaves. With 3 or more peers, the etcd cluster still has quorum and can continue to serve requests.

Solving this requires changes to elasticetcd and the deletepeer code, to introduce special cases for handling deletes in a 2node cluster.

prashanthpai:
Related to issue #323 ?

atinmu:
Is this issue still valid?

PR #347, opened 2017-07-27
CI test PR [Do NOT merge]

prashanthpai:


prashanthpai:
retest this please

prashanthpai:
@kshlm just adding the listing operation causes the CI test to fail (in list peer on my machine but remove peer on CI machines) which shouldn't happen. The same 2 node test passes when it doesn't do the listing!

prashanthpai:
I'm closing this PR. This has demonstrated the race that I could see and the [issue with quorum in a two-node cluster](https://github.com/gluster/glusterd2/issues/332)

PR #351, opened 2017-08-01
volgen: Template file based graph and volfile generator

kshlm:
Volgen aims to work as follows,

- Developers create text files to define the structure of various
GlusterFS graphs. (Format isn't finalized yet)
- Volgen reads these template files, to create a an empty template
graph.
- From this template graph, volgen can generate final graphs by filling
in information from volinfo
- This graph will be converted to a volfile using a common volfile
template.

This current commit handles linear graphs and can generate a volfile for
a brick with some options missing. As the generated volfile isn't yet
workable, this hasn't been integrated with the rest of GD2.


purpleidea:
Can you briefly describe the graph format for the stupid ones like me that don't know as much about gluster internals? Is this a DAG? Is it a list? Thanks!

kshlm:
> Can you briefly describe the graph format for the stupid ones like me that don't know as much about gluster internals? Is this a DAG? Is it a list? Thanks!

GlusterFS volume graphs are DAGs, with the xlators as nodes and the direction of the edges representing the flow of requests. Each glusterfs process has it's own graph of xlators, specifically created for the function it needs to perform. The graphs are described in the volfiles, which are used by the glusterfsd process to build and load the actual graph.

In a most cases, GlusterFS graphs are either linear with no branches, or symmetric trees. Brick graphs are linear and client graphs are symmetric (when having more than one brick). We have a few possible asymmetric graphs, for example the graphs for tiered volumes.

The volfiles describe the final graph including all the branches, names and options of the xlators. The template file in this PR, just describes the order/hierarchy of the xlators in the tree. The volgen package reads this template to build a template graph. It then generates the final graph by filling in missing information using the volume info, and writes out the final graph as a volfile.

Glusterd-1 also did the same, except the order was hard coded and each graph type had unique generator functions. This made it hard to modify these graphs. For GD2 we want to remove this sort of hardcoding, and provide tools to make it simple for developers to define graphs. This PR is an effort in that direction.

Issue #356, opened 2017-08-11
Add option to change hostname in specifying brick (or use UUID)


prashanthpai:
When a node's IP changes, there should be an option to edit the brick host:path representation OR we should only use UUID:brick representation always. That makes things really easy.

Reference:
https://bugzilla.redhat.com/show_bug.cgi?id=1036566

joejulian:
Original reference:
Bug 765437: [FEAT] Use uuid in volume info file for servers instead of hostname or ip address
https://bugzilla.redhat.com/show_bug.cgi?id=765437

atinmu:
@prashanthpai Isn't this already being implemented?

prashanthpai:
> @prashanthpai Isn't this already being implemented?

This issue was broken up into multiple tasks - the first of which has been completed in PR #357 . There's more work to be done.

atinmu:
@rishubhjain @prashanthpai Anything pending on this?

PR #357, opened 2017-08-18
Make UUID:path as the ONLY way of specifying brick during volume create

prashanthpai:
* Remove Hostname field from Brickinfo
* Remove "remote-host" hardcoded in client volfile
* Fix unit tests, logging and e2e tests accordingly
* Add String() method to Brickinfo type

This partially fixes issues #356 

TODO (separate PR): Add REST API to change IP address of the peer.

### DIFF
--- a/brick/glusterfsd.go
+++ b/brick/glusterfsd.go
@@ -87,7 +87,8 @@ func (b *Glusterfsd) SocketFile() string {
 	// First we form a fake path to the socket file
 	// Example: /var/lib/glusterd/vols/<vol-name>/run/<host-name>-<brick-path>
 	brickPathWithoutSlashes := strings.Trim(strings.Replace(b.brickinfo.Path, "/", "-", -1), "-")
-	fakeSockFileName := fmt.Sprintf("%s-%s", b.brickinfo.Hostname, brickPathWithoutSlashes)
+	// FIXME: The brick can no longer clean this up on clean shut down
+	fakeSockFileName := fmt.Sprintf("%s-%s", b.brickinfo.NodeID.String(), brickPathWithoutSlashes)
 	volumedir := utils.GetVolumeDir(b.brickinfo.VolumeName)
 	fakeSockFilePath := path.Join(volumedir, "run", fakeSockFileName)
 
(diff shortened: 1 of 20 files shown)


prashanthpai:
> I'm not so okay about just supporting UUIDs for specifying bricks in the APIs, as UUIDs are not human friendly

The CLI can be the one providing the human-friendly front-end and REST API can stay "raw" or "low-level" considering the fact that it's almost always programs (restclient,CLI,tendrl etc) using the REST API and not humans.

> We will need to provide a human friendly way to specify peers (and volumes and other entities)

Yes, definitely. I'm only suggesting for CLI to be the place for it.

It's straight-forward for volumes as volume names are themselves unique. Peer names aren't unique as of today and it'll be a huge burden on users if we mandate the peer names to be unique. The REST API can mandate using only UUIDs and the UIs/frontend/CLI can add the human-friendly aspect.

From my previous experience, when there are more than one ways to do something via a published API, it can lead to problems w.r.t confusion for API consumers, deprecation of one of the ways in future and internal debugging.

If this is really wanted, these are the TODOs I could think of:

1, *Name* field in peer structure will be settable via conf file (initial starup) and during peer attach.
2. *Name* field (along with addresses) must be changeable later via REST API
3. **IF** *Name* is unique among all peers, allow volume create requests to go through.

**1** and **2** will have to be worked on anyway I think.

kshlm:
If the human friendliness is handled in the CLI, I'm okay with it. I'll hold on merging this patch though, so that we do implement it in the CLI.

aravindavk:
+1 for ID based brick info. I think CLI calls are not very expensive. One extra call to get peers info before calling volume create, add brick, remove brick, info/status. Rest of the commands are unaffected.

If we can add hosts field to BrickInfo and generate when volinfo/status requested. So that even info/status call need not call extra API.

prashanthpai:
> If we can add hosts field to BrickInfo and generate when volinfo/status requested. So that even info/status call need not call extra API.

This can be done. We certainly want internal structs and the structs marshalled and returned to clients to be different. Currently they're the same.

prashanthpai:
@kshlm The CLI support has been added now.

```sh
$ glustercli volume create test 192.168.56.25:/export/brick1/data force
test Volume created successfully
Volume ID:  f3ce2396-1fee-4cee-9000-ade4585e29fa
$
$ glustercli volume delete test
Volume test deleted successfully
$ 
[ppai@gd2-1 glusterd2]$ glustercli volume create test 9fbcb443-fc00-437c-b78d-d2f464b3df8c:/export/brick1/data force
test Volume created successfully
Volume ID:  b42dfeed-3561-4619-a6c4-a2bd2e00f889
```

PR #360, opened 2017-08-29
Update quick-start-guide

kshlm:


kshlm:
@prashanthpai Thanks!

PR #362, opened 2017-08-30
cli: Add volume options for the Glusterd2 CLI.

devyanikota:
Added the volume create option for the Glusterd2 CLI.
Currently, a volume can be created in the following way:

`./glusterd2 create volume <force> <volname> <bricks>`

Where the options are to be given in this certain order.
We use the flag package to parse the arguments from the command line.
As suggested by @kshlm that needs to be changed in favor of cobra.
WIP.


cli/cmd/volume.go
```diff
@@ -75,8 +83,29 @@ var volumeCreateCmd = &cobra.Command{
 	Short: helpVolumeCreateCmd,
 	Run: func(cmd *cobra.Command, args []string) {
 		validateNArgs(cmd, 2, 0)
-		volname := cmd.Flags().Args()[0]
-		bricks := cmd.Flags().Args()[1:]
+		force := cmd.Flags().Args()[0]
```
aravindavk: force flag is already added. Check the same file for `flagCreateCmdForce`

cli/cmd/volume.go
```diff
@@ -75,8 +83,29 @@ var volumeCreateCmd = &cobra.Command{
...
+		volname := cmd.Flags().Args()[1]
+		bricks := cmd.Flags().Args()[2:]
+		force_, err := strconv.ParseBool(force)
```
aravindavk: This is not required, already available as flag. as mentioned in previous comment

cli/cmd/volume.go
```diff
@@ -75,8 +83,29 @@ var volumeCreateCmd = &cobra.Command{
 	Short: helpVolumeCreateCmd,
 	Run: func(cmd *cobra.Command, args []string) {
 		validateNArgs(cmd, 2, 0)
-		volname := cmd.Flags().Args()[0]
-		bricks := cmd.Flags().Args()[1:]
+		force := cmd.Flags().Args()[0]
+		volname := cmd.Flags().Args()[1]
+		bricks := cmd.Flags().Args()[2:]
```
aravindavk: Change back to [0] and [1:] as earlier

cli/cmd/volume.go
```diff
@@ -75,8 +83,29 @@ var volumeCreateCmd = &cobra.Command{
...
+			Name: volname,
+			Bricks: bricks,
+			Force: force_,
+		})
+		if err != nil {
+			log.WithField("volume", volname).Println("volume creation failed")
+			failure(fmt.Sprintf("Volume creation failed with %s", err.Error()), 1)
+		}
+		fmt.Println("Volume created successfully\n")
+		table := tablewriter.NewWriter(os.Stdout)
```
aravindavk: Try to follow the same format as existing `gluster volume info` command. Table may not be suitable here.

aravindavk:
> ./glusterd2 create volume <force> <volname> <bricks>

It should be `./glustercli volume create <volname> <bricks> --force`

devyanikota:
@aravindavk nod nod. I manually typed that command. It is ./glustercli
Will integrate the suggested changes. Thanks.

devyanikota:
Hey @aravindavk, worked on the suggested changes. Can you please check, thanks.

cli/cmd/volume.go
```diff
@@ -74,10 +79,20 @@ var volumeCreateCmd = &cobra.Command{
 	Use:   "create",
 	Short: helpVolumeCreateCmd,
 	Run: func(cmd *cobra.Command, args []string) {
-		validateNArgs(cmd, 2, 0)
+		validateNArgs(cmd, 2, 4)
```
aravindavk: Why 4? 0 means unlimited args.

cli/cmd/volume.go
```diff
@@ -74,10 +79,20 @@ var volumeCreateCmd = &cobra.Command{
...
+		vol, err := client.VolumeCreate(api.VolCreateReq{
+			Name: volname,
+			Bricks: bricks,
+			Force: flagCreateCmdForce,
+		})
+		if err != nil {
+			log.WithField("volume", volname).Println("volume creation failed")
+			failure(fmt.Sprintf("Volume creation failed with %s", err.Error()), 1)
+		}
+		fmt.Println(vol.ID)
```
aravindavk: Add other fields of Volume as summary. (Same format as `gluster volume info` with glusterd1)

cli/cmd/volume.go
devyanikota: Rectified :)

cli/cmd/volume.go
devyanikota: Removed.

cli/cmd/volume.go
devyanikota: Added.

cli/cmd/volume.go
devyanikota: Changed the same.

kshlm:
add to whitelist

cli/cmd/volume.go
```diff
@@ -3,6 +3,11 @@ package cmd
 import (
 	"fmt"
 	"github.com/spf13/cobra"
+
```
kshlm: When we import packages we group packages in blocks in the following order.
- Stdlib packages
- Packages from our project
- External packages

Please group them correctly here.

cli/cmd/volume.go
```diff
@@ -77,7 +82,27 @@ var volumeCreateCmd = &cobra.Command{
...
+			Name: volname,
+			Bricks: bricks,
+			Replica: flagCreateCmdReplicaCount,
+			Force: flagCreateCmdForce,
+		})
+		if err != nil {
+			log.WithField("volume", volname).Println("volume creation failed")
+			failure(fmt.Sprintf("Volume creation failed with %s", err.Error()), 1)
+		}
+		fmt.Println("Volume created successfully\n")
```
kshlm: Why not use the tabulator here to get pretty output?

Also, the `volume create` command doesn't show this much information in the old cli.

aravindavk: ID is required in the output since user should know the ID of volume created, showing volinfo will be like added value I feel. Previous patchset had tabulator, but I requested this format to match old format.

cli/cmd/volume.go
devyanikota: Noted. Will rectify the same.

cli/cmd/volume.go
kshlm: I'd say just the ID would be enough extra information. But we can update the output format later.

devyanikota: Updating the patch. Adding the volume info in the volumeInfoCmd().
thanks.

devyanikota:
Hi, @aravindavk I have made the suggested changes, can you please review the same.
Thank you.

cli/cmd/volume.go
```diff
@@ -78,7 +79,18 @@ var volumeCreateCmd = &cobra.Command{
 		validateNArgs(cmd, 2, 0)
 		volname := cmd.Flags().Args()[0]
 		bricks := cmd.Flags().Args()[1:]
```
prashanthpai: Can you please check if this will work when `replica` and `transport` keywords are specified in the CLI.

aravindavk: replica and transport are flags. --replica, --transport etc. So it will work.

Callback func has two arguments, `func(cmd *cobra.Command, args []string)` second argument contains os.Args, but cmd.Flags().Args() contains only arguments excluding all the flags. So this approach should work.

https://godoc.org/github.com/spf13/pflag#FlagSet

PR #365, opened 2017-08-31
cli: Added cli for volume start and stop.

nnDarshan:
Updates: #205
Signed-off-by: nnDarshan <darshan.n.2024@gmail.com>


nnDarshan:
@aravindavk @kshlm Pls review

cli/cmd/volume.go
```diff
@@ -82,22 +83,32 @@ var volumeCreateCmd = &cobra.Command{
 }
 
 var volumeStartCmd = &cobra.Command{
-	Use:   "start [flags] <VOLNAME>",
+	Use:   "start <VOLNAME>",
```
aravindavk: You can retain [flags], which will be added later for backward compatibility

cli/cmd/volume.go
```diff
@@ -82,22 +83,32 @@ var volumeCreateCmd = &cobra.Command{
...
+			log.WithField("volume", volname).Println("volume start failed")
+			failure(fmt.Sprintf("volume start failed with %s", err.Error()), 1)
+		}
+		fmt.Printf("Volume %s started successfully\n", volname)
 	},
 }
 
 var volumeStopCmd = &cobra.Command{
-	Use:   "stop",
+	Use:   "stop <VOLNAME>",
```
aravindavk: same as above

cli/cmd/volume.go
nnDarshan: ack

cli/cmd/volume.go
nnDarshan: ack

kshlm:
add to whitelist

cli/cmd/volume.go
```diff
@@ -87,17 +88,27 @@ var volumeStartCmd = &cobra.Command{
...
+			log.WithField("volume", volname).Println("volume start failed")
+			failure(fmt.Sprintf("volume start failed with: %s", err.Error()), 1)
+		}
+		fmt.Printf("Volume %s started successfully\n", volname)
 	},
 }
 
 var volumeStopCmd = &cobra.Command{
-	Use:   "stop",
+	Use:   "stop [flags] <VOLNAME>",
```
kshlm: What are the 'flags' here? Do you even need to mention it here? I thought cobra automatically adds all defined flags to the commands help outpu.

aravindavk: I just verified about [flags], it is optional. cobra will add it if not added. Adds it to the end.

Sorry @nnDarshan, I requested you to add flag in last patchset. It is optional.

cobra doesn't have any facility for positional arguments as for as I know. So positional args required.

PR #367, opened 2017-08-31
volume-options: store req.Options in volinfo

### LINKED ISSUE #366: GET /v1/volumes/<volname> doesn't display volume options set through volume options API
While I was trying to set some volume options through volume options API, the command succeeded. However get volume API returned blank options field.

atinmu:
Fixes #366

### DIFF
--- a/commands/volumes/volume-option.go
+++ b/commands/volumes/volume-option.go
@@ -58,6 +58,10 @@ func volumeOptionsHandler(w http.ResponseWriter, r *http.Request) {
 		restutils.SendHTTPError(w, http.StatusBadRequest, fmt.Sprintf("invalid option specified: %s", err.Error()))
 		return
 	}
+	// Update the options into volinfo
+	for k, v := range req.Options {
+		volinfo.Options[k] = v
+	}
 
 	lock, unlock, err := transaction.CreateLockSteps(volinfo.Name)
 	if err != nil {


kshlm:
add to whitelist

commands/volumes/volume-option.go
```diff
@@ -58,6 +58,10 @@ func volumeOptionsHandler(w http.ResponseWriter, r *http.Request) {
 		restutils.SendHTTPError(w, http.StatusBadRequest, fmt.Sprintf("invalid option specified: %s", err.Error()))
 		return
 	}
+	// Update the options into volinfo
+	for k, v := range req.Options {
+		volinfo.Options[k] = v
+	}
```
kshlm: There is a loop that does exactly this on line 106, before `txn.Do()` is called. Why doesn't it work there?

commands/volumes/volume-option.go
```diff
@@ -58,6 +58,10 @@ func volumeOptionsHandler(w http.ResponseWriter, r *http.Request) {
 		restutils.SendHTTPError(w, http.StatusBadRequest, fmt.Sprintf("invalid option specified: %s", err.Error()))
 		return
 	}
+	// Update the options into volinfo
```
aravindavk: Already this code is available in line 106. Looks like some other issue while storing in volinfo

atinmu: Correct, I overlooked this code. Let me figure it out where the problem is.

prashanthpai:
PR #374  fixes the issue.

PR #371, opened 2017-09-14
Allow option of the form <graph>.<xlator>.<option>

prashanthpai:
Volume options can now be of the form `<graph>.<xlator>.<option>` where `<graph>`
is optional and when omitted, the option change shall be applied (by volgen)
to instances of the xlator loaded in all graphs.

### DIFF
--- a/commands/volumes/common.go
+++ b/commands/volumes/common.go
@@ -23,22 +23,6 @@ func (e invalidOptionError) Error() string {
 	return e.option
 }
 
-func splitOptionName(option string) (string, string, string) {
-	// Option can be of the form <graph>.<xlator>.<option>
-	// where <graph> is optional and when omitted, the option change shall
-	// be applied to instances of the xlator loaded in all graphs.
-
-	tmp := strings.Split(strings.TrimSpace(option), ".")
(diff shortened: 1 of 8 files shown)


prashanthpai:
retest this please

commands/volumes/common.go
```diff
@@ -23,22 +23,33 @@ func (e invalidOptionError) Error() string {
 	return e.option
 }
 
+func splitOptionName(option string) (string, string, string) {
```
kshlm: It will be nice to have this function exported, and possibly in a more common place, as I'd like to use it in volgen as well. We would need a volume-set or an options package to perform option validation. This function will be a nice addition in it.

prashanthpai: I'll move this to `utils` package or `volume/volume-utils.go` for now. The `xlator` package currently has code such as definition of `Option` struct. There's no clear conceptual distinction between an xlator option and a volume option. Practically, most of the times they mean the same. We can consolidate these into one package in future.

prashanthpai:
retest this please

prashanthpai:
retest this please

PR #378, opened 2017-09-27
Fix TestVolumeOptions and running unit test as non-root

prashanthpai:
TestVolumeOptions used to be skipped on CentOS CI as it didn't have glusterfs
installed from source. Fixed that test to reflect change in volume create API.
`TestVolumeOptions` is not skipped anymore. GlusterFS is installed from source, for now.

`make test` when run as non-root only runs unit tests now.

### DIFF
--- a/e2e/volume_ops_test.go
+++ b/e2e/volume_ops_test.go
@@ -76,7 +76,7 @@ func TestVolumeOptions(t *testing.T) {
 	volname := "testvol"
 	createReq := api.VolCreateReq{
 		Name:   volname,
-		Bricks: []string{gds[0].PeerAddress + ":" + brickPath},
+		Bricks: []string{gds[0].PeerID() + ":" + brickPath},
 		Force:  true,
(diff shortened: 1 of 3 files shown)


e2e/volume_ops_test.go
```diff
@@ -76,7 +76,7 @@ func TestVolumeOptions(t *testing.T) {
 	volname := "testvol"
 	createReq := api.VolCreateReq{
 		Name:   volname,
-		Bricks: []string{gds[0].PeerAddress + ":" + brickPath},
+		Bricks: []string{gds[0].PeerID() + ":" + brickPath},
```
atinmu: Even PeerAddress should work here, isn't it?

prashanthpai: No. This was changed to supporting only UUID in PR https://github.com/gluster/glusterd2/pull/357
The CLI supports both UUID and peer address.

PR #379, opened 2017-09-27
CLI: Volume set option for gd2 cli

devyanikota:
Updates #205

### DIFF
--- a/cli/cmd/common.go
+++ b/cli/cmd/common.go
@@ -23,7 +23,9 @@ func failure(msg string, err int) {
 func validateNArgs(cmd *cobra.Command, min int, max int) {
 	nargs := len(cmd.Flags().Args())
 	if nargs < min || (max != 0 && nargs > max) {
-		cmd.Usage()
-		os.Exit(1)
+		if nargs-1 % 2 == 0 {
+			cmd.Usage()
+			os.Exit(1)
+		}
 	}
 }
--- a/cli/cmd/common.go
+++ b/cli/cmd/common.go
@@ -23,9 +23,7 @@ func failure(msg string, err int) {
 func validateNArgs(cmd *cobra.Command, min int, max int) {
 	nargs := len(cmd.Flags().Args())
 	if nargs < min || (max != 0 && nargs > max) {
-		if nargs-1 % 2 == 0 {
-			cmd.Usage()
-			os.Exit(1)
-		}
+		cmd.Usage()
+		os.Exit(1)
 	}
 }
--- a/cli/cmd/volume.go
+++ b/cli/cmd/volume.go
@@ -4,8 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"strings"
-	"encoding/json"
-	"os"
 
 	"github.com/gluster/glusterd2/pkg/api"
 	"github.com/pborman/uuid"
@@ -42,9 +40,6 @@ var (
 
 	// Stop Command Flags
 	flagStopCmdForce bool
-
-	// Set Command Flags
-	flagSetCmdData string
 )
 
 func init() {
@@ -69,11 +64,8 @@ func init() {
 	// Volume Delete
 	volumeCmd.AddCommand(volumeDeleteCmd)
 
-    //Volume Set
-	volumeSetCmd.Flags().StringVarP(&flagSetCmdData, "data", "d", "", "Volume Options")
-	volumeCmd.AddCommand(volumeSetCmd)
-
 	volumeCmd.AddCommand(volumeGetCmd)
+	volumeCmd.AddCommand(volumeSetCmd)
 	volumeCmd.AddCommand(volumeResetCmd)
 	volumeCmd.AddCommand(volumeInfoCmd)
 	volumeCmd.AddCommand(volumeStatusCmd)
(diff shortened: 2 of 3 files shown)


devyanikota:
@aravindavk is there a way we can get the value of the Flag Option.
Am trying to extract the volume options entered by the user using os.Args[] (Line 229) because:
On extracting using Flags.Args(), it omits the first option key :/

Example:
`$./glustercli volume set testvol --option afr.eager-lock on foo bar`

Output of the created json:
`[on foo bar]`

The solution here can be to extract the value of the flag 'option' and append the rest of the array of options.

aravindavk:
> Am trying to extract the volume options entered by the user using os.Args[] (Line 229) because:
On extracting using Flags.Args(), it omits the first option key :/

Why you need `--option`?

cli/cmd/volume.go
```diff
@@ -202,13 +210,36 @@ var volumeGetCmd = &cobra.Command{
...
+
 var volumeSetCmd = &cobra.Command{
 	Use:   "set",
 	Short: helpVolumeSetCmd,
 	Run: func(cmd *cobra.Command, args []string) {
-		validateNArgs(cmd, 3, 3)
+		validateNArgs(cmd, 3, 0)
 		volname := cmd.Flags().Args()[0]
-		fmt.Println("SET:", volname)
+		options := os.Args[5:]
```
aravindavk: Do not take Args directly from os.Args, This will have raw arguments which are not parsed. Use cmd.Flags().Args()

cli/cmd/volume.go
```diff
@@ -202,13 +210,36 @@ var volumeGetCmd = &cobra.Command{
...
+func volumeOptionJsonHandler(cmd *cobra.Command, volname string, options []string) error {
+	vopt := make(map[string]string)
+	for op,val := range options {
+		if op % 2 == 0 {
+			vopt[val] = options[op+1]
+		}else{
+			continue
+		}
+	}
+	var op = map[string]interface{}{"options": vopt}
```
aravindavk: Avoid interface where not required. Try,

var op = map[string]map[string]string

cli/cmd/volume.go
```diff
@@ -64,8 +69,11 @@ func init() {
 	// Volume Delete
 	volumeCmd.AddCommand(volumeDeleteCmd)
 
-	volumeCmd.AddCommand(volumeGetCmd)
+    //Volume Set
+	volumeSetCmd.Flags().StringVarP(&flagSetCmdData, "data", "d", "", "Volume Options")
```
aravindavk: Why this is required?

cli/cmd/common.go
```diff
@@ -23,7 +23,9 @@ func failure(msg string, err int) {
 func validateNArgs(cmd *cobra.Command, min int, max int) {
 	nargs := len(cmd.Flags().Args())
 	if nargs < min || (max != 0 && nargs > max) {
-		cmd.Usage()
-		os.Exit(1)
+		if nargs-1 % 2 == 0 {
```
aravindavk: why this change here? This is only required in VolumeSet, move this to that func itself

cli/cmd/volume.go
```diff
@@ -202,13 +210,36 @@ var volumeGetCmd = &cobra.Command{
 	},
 }
 
+func volumeOptionJsonHandler(cmd *cobra.Command, volname string, options []string) error {
+	vopt := make(map[string]string)
+	for op,val := range options {
+		if op % 2 == 0 {
+			vopt[val] = options[op+1]
+		}else{
+			continue
```
aravindavk: else statement not adding any value here, can be removed

cli/cmd/volume.go
devyanikota: I tried using cmd.Flags().Args()
but, like I explained earlier:
$./glustercli volume set testvol --data afr.eager-lock on foo bar
The volume options array being created is [on foo bar].
The first volume option entered is being ommitted.
Hence, thought of extracting the value of the flag: 'data' and then append it to the array.

cli/cmd/volume.go
devyanikota: Noted.

cli/cmd/volume.go
devyanikota: um, I had named the flag as 'option' earlier, rectified and renamed : '--data'

cli/cmd/volume.go
devyanikota: Right. ok. Noted.

cli/cmd/common.go
devyanikota: My bad here. Sorry. Rectified.

cli/cmd/volume.go
aravindavk: Both of them are not required. Do not use any flags for volume set. Just accept all as arguments.

    $./glustercli volume set testvol afr.eager-lock on foo bar

cli/cmd/volume.go
aravindavk: That is because of `--option` or `--data` flag. Both of them not required. Just use 

    $./glustercli volume set testvol afr.eager-lock on foo bar

PR #384, opened 2017-10-10
logrus: Fix repo name

prashanthpai:
Ref: https://github.com/sirupsen/logrus/issues/543

Reported by: kmohanan


prashanthpai:
> Was it working earlier because the previous path is somehow symlinked?

The old repo still can be cloned.

Issue #271, opened 2017-02-27
Support geo-replication


prashanthpai:

aravindavk:
Details of Geo-replication plugin implementation

## Create command                                                                                                                                                                                    
Existing,
                                                                                                                                                                                                       
- Verify Slave Volume version, Space available in Slave Volume etc                                                                                                                                                   
- Record session creation information in VOLINFO file in all peer nodes                                                                                                                                              
- Enables "changelog" and "index" features in Master Volume                                                                                                                                                          
- Creates config file in all Master nodes                                                                                                                                                                            
- Creates session directory in `/var/lib/glusterd/geo-replication/` and creates                                                                                                                                      
  monitor.status file inside that directory.                                                                                                                                                                         
- Pushes ssh keys of master nodes to Slave nodes and add to authorized_keys file                                                                                                                                     
  in Slave nodes.                                                                                                                                                                                                    
                                                                                                                                                                                                                     
With Glusterd2,                                                                                                                                                                                                      
                                                                                                                                                                                                                     
- Verification yet to be decided(Since this involves reaching Slave Cluster)                                                                                                                                         
- Record session creation information in etcd store.                                                                                                                                                                 
- Enables "changelog" and "index" features using Volume Set APIs                                                                                                                                                     
- Session directory and monitor status file not required since the status can be                                                                                                                                     
  maintained in etcd. Geo-rep monitor process can create session directory when                                                                                                                                      
  started.                                                                                                                                                                                                           
- Push ssh keys APIs yet to be decided.                                                                                                                                                                              
                                                                                                                                                                                                                     
Note: Verification, SSH Key generation and distribution steps are manual now.                                                                                                                                        
                                                                                                                                                                                                                     
## Start Command                                                                                                                                                                                     
                                                                                                                                                                                                                     
Existing,
                                                                                                                                                                                                                     
- Start Geo-rep monitor process in all peer nodes and update monitor status as                                                                                                                                       
  "Started"                                                                                                                                                                                                          
                                                                                                                                                                                                                     
With Glusterd2,                                                                                                                                                                                                      
                                                                                                                                                                                                                     
- Register Geo-rep monitor as enabled daemon in Glusterd2(with required args)                                                                                                                                        
- Start the daemon in all master nodes(No need to start in all peer nodes)                                                                                                                                           
                                                                                                                                                                                                                     
## Stop Command                                                                                                                                                                                      
                                                                                                                                                                                                                     
[...]

PR #385, opened 2017-10-10
plugins/geo-rep: Geo-replication Session Create

aravindavk:
REST API to create Geo-replication session create

    POST /v1/geo-replication/:mastervolid/:slavevolid

Request body,

    {
        "mastervol": MASTER_VOL_NAME,
        "slavevol": SLAVE_VOL_NAME,
        "slavehosts": LIST_OF_SLAVE_HOSTS,
        "slaveuser": SLAVE_USER
    }

Updates: #271

### DIFF
--- a/plugins/georeplication/api/req.go
+++ b/plugins/georeplication/api/req.go
@@ -1,7 +1,7 @@
 package georeplication
 
-// GeorepCreateRequest represents REST API request to create Geo-rep session
-type GeorepCreateRequest struct {
+// GeorepCreateReq represents REST API request to create Geo-rep session
+type GeorepCreateReq struct {
 	MasterVol  string   `json:"mastervol"`
 	SlaveUser  string   `json:"slaveuser"`
 	SlaveHosts []string `json:"slavehosts"`
--- a/plugins/georeplication/api/req.go
+++ b/plugins/georeplication/api/req.go
@@ -1,0 +1,9 @@
+package georeplication
+
+// GeorepCreateRequest represents REST API request to create Geo-rep session
+type GeorepCreateRequest struct {
+	MasterVol  string   `json:"mastervol"`
+	SlaveUser  string   `json:"slaveuser"`
+	SlaveHosts []string `json:"slavehosts"`
+	SlaveVol   string   `json:"slavevol"`
+}
--- a/plugins/georeplication/api/resp.go
+++ b/plugins/georeplication/api/resp.go
@@ -54,22 +54,3 @@ type GeorepSession struct {
 	Workers    []GeorepWorker    `json:"workers"`
 	Options    map[string]string `json:"options"`
 }
-
-// NewGeorepSession creates new instance of GeorepSession
-func NewGeorepSession(mastervolid uuid.UUID, slavevolid uuid.UUID, req GeorepCreateReq) *GeorepSession {
-	slaveUser := req.SlaveUser
-	if req.SlaveUser == "" {
-		slaveUser = "root"
-	}
-	return &GeorepSession{
-		MasterID:   mastervolid,
-		SlaveID:    slavevolid,
-		MasterVol:  req.MasterVol,
-		SlaveVol:   req.SlaveVol,
-		SlaveHosts: req.SlaveHosts,
-		SlaveUser:  slaveUser,
-		Status:     GeorepStatusCreated,
-		Workers:    []GeorepWorker{},
-		Options:    make(map[string]string),
-	}
-}
--- a/plugins/georeplication/api/resp.go
+++ b/plugins/georeplication/api/resp.go
@@ -54,3 +54,22 @@ type GeorepSession struct {
 	Workers    []GeorepWorker    `json:"workers"`
 	Options    map[string]string `json:"options"`
 }
+
+// NewGeorepSession creates new instance of GeorepSession
+func NewGeorepSession(mastervolid uuid.UUID, slavevolid uuid.UUID, req GeorepCreateReq) *GeorepSession {
+	slaveUser := req.SlaveUser
+	if req.SlaveUser == "" {
+		slaveUser = "root"
+	}
+	return &GeorepSession{
+		MasterID:   mastervolid,
+		SlaveID:    slavevolid,
+		MasterVol:  req.MasterVol,
+		SlaveVol:   req.SlaveVol,
+		SlaveHosts: req.SlaveHosts,
+		SlaveUser:  slaveUser,
+		Status:     GeorepStatusCreated,
+		Workers:    []GeorepWorker{},
+		Options:    make(map[string]string),
+	}
+}
--- a/plugins/georeplication/api/resp.go
+++ b/plugins/georeplication/api/resp.go
@@ -1,0 +1,56 @@
+package georeplication
+
+import (
+	"github.com/pborman/uuid"
+)
+
+const (
+	// GeorepStatusCreated represents Created State
+	GeorepStatusCreated = "Created"
+
+	// GeorepStatusStarted represents Started State
+	GeorepStatusStarted = "Started"
+
+	// GeorepStatusStopped represents Stopped State
+	GeorepStatusStopped = "Stopped"
+
+	// GeorepStatusPaused represents Paused State
+	GeorepStatusPaused = "Paused"
+)
+
+// GeorepWorker represents Geo-replication Worker
+type GeorepWorker struct {
+	MasterNode                 string `json:"master_node"`
+	MasterNodeID               string `json:"node_id"`
+	MasterBrickPath            string `json:"master_brick_path"`
+	MasterBrick                string `json:"master_brick"`
+	Status                     string `json:"status"`
+	LastSyncedTime             string `json:"last_synced_time"`
+	LastSyncedTimeUTC          string `json:"last_synced_time_utc"`
+	LastEntrySyncedTime        string `json:"last_synced_entry"`
+	SlaveNode                  string `json:"slave_node"`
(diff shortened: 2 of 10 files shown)


plugins/georeplication/api/resp.go
```diff
@@ -0,0 +1,56 @@
+package georeplication
+
+import (
+	"github.com/pborman/uuid"
+)
+
+const (
```
prashanthpai: Just a FYI: These can be like enums instead. There's [stringer](https://godoc.org/golang.org/x/tools/cmd/stringer) tool which can make it strings. I got to know about this while reviewing the volgen PR. You can take a look at it if it makes things simpler/easier for you.

plugins/georeplication/api/resp.go
```diff
@@ -0,0 +1,56 @@
...
+	MetaOps                    string `json:"meta"`
+	EntryOps                   string `json:"entry"`
+	DataOps                    string `json:"data"`
+	FailedOps                  string `json:"failures"`
+	CrawlStatus                string `json:"crawl_status"`
+}
+
+// GeorepSession represents Geo-replication session
+type GeorepSession struct {
+	MasterID   uuid.UUID         `json:"master_volume_id"`
```
prashanthpai: Check if the string representation of UUID is what that is seen in response.

plugins/georeplication/init.go
```diff
@@ -0,0 +1,46 @@
...
+func (p *Plugin) SunRPCProgram() sunrpc.Program {
+	return nil
+}
+
+// RestRoutes returns list of REST API routes to register with Glusterd
+func (p *Plugin) RestRoutes() route.Routes {
+	return route.Routes{
+		route.Route{
+			Name:        "GeoreplicationCreate",
+			Method:      "PUT",
```
prashanthpai: Personally, I really prefer `PUT` over `POST` for resource creation and `POST` for resource modification.

But it's been `POST` for all requests in REST API so far. We should open up this for discussion. We will need the APIs to be uniform i.e creation of volumes, snapshot, georep to be all `PUT` or all `POST`.

plugins/georeplication/rest.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// Same handler for both Create and Update, if HTTP method is POST then
+	// it is update request
+	updateRequest := false
+	if r.Method == http.MethodPost {
+		updateRequest = true
+	}
+
+	// Validate UUID format of Master and Slave Volume ID
+	masterid, slaveid, err := validateMasterAndSlaveIDFormat(w, masteridRaw, slaveidRaw)
```
prashanthpai: To be consistent with rest of the code, you can return SendHTTPError() from the handler itself.

```go
 if uuid.Parse(masteridRaw) == nil || uuid.Parse(slaveidRaw) == nil {
    restutils.SendHTTPError(w, http.StatusBadRequest, "Invalid Master/Slave Volume ID")
    return
}
```
Or you can put all of the below validations (MasterVol, SlaveHosts, SlaveVol) in a single validation function. This will also make the handler function smaller and more readable.

plugins/georeplication/init.go
prashanthpai: Also, `mastervolid` and `slaveid` can also be part of request body. Any reason for it being part of URL ?

plugins/georeplication/rest.go
```diff
@@ -0,0 +1,193 @@
...
+			restutils.SendHTTPError(w, http.StatusInternalServerError, err.Error())
+			return
+		}
+		// Fail if update request received for non existing Geo-rep session
+		if updateRequest {
+			restutils.SendHTTPError(w, http.StatusBadRequest, "geo-replication session not found")
+			return
+		}
+
+		geoSession = new(georepapi.GeorepSession)
```
prashanthpai: Consider populating this struct instance (including members filled below) in it's own function, may be something like:

```go
func newGeorepSession(req *georepapi.GeorepCreateRequest) *georepapi.GeorepSession
```

plugins/georeplication/rest.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// TODO: Transaction step function for setting Volume Options
+	// As a workaround, Set volume options before enabling Geo-rep session
+
+	txn.Nodes = vol.Nodes()
+	txn.Steps = []*transaction.Step{
+		lock,
+		{
+			DoFunc: "georeplication-create.Commit",
+			Nodes:  []uuid.UUID{gdctx.MyUUID},
```
prashanthpai: As this run only one node, is this made a transaction only for the locking ?

plugins/georeplication/rest.go
```diff
@@ -0,0 +1,193 @@
...
+			"mastervolid": masterid,
+			"slavevolid":  slaveid,
+		}).Error("failed to create geo-replication session")
+		restutils.SendHTTPError(w, http.StatusInternalServerError, e.Error())
+		return
+	}
+
+	geoSession.Status = georepapi.GeorepStatusCreated
+
+	e = addOrUpdateSession(geoSession)
```
prashanthpai: Can't this part of `txnGeorepCreate` ?

plugins/georeplication/api/req.go
```diff
@@ -0,0 +1,9 @@
+package georeplication
+
+// GeorepCreateRequest represents REST API request to create Geo-rep session
+type GeorepCreateRequest struct {
```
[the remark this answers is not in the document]
aravindavk: SlaveUser is optional, if not provided defaults to "root"

plugins/georeplication/api/resp.go
aravindavk: Let me explore that, thanks. Is that libray already part of glusterd2?

plugins/georeplication/api/resp.go
aravindavk: Yes. Sample output

HTTP/1.1 200 OK
Content-Type: application/json; charset=UTF-8
X-Request-Id: 52e4ac7a-a470-4a9c-952f-687bebe442f0
Date: Thu, 12 Oct 2017 07:26:23 GMT
Content-Length: 255

{"master_volume_id":"254218fe-37ac-40e7-9b76-980700217862","slave_volume_id":"bc8834b4-c710-4fd6-b07e-f25380595a62","master_volume":"gv1","slave_user":"root","slave_hosts":["f26"],"slave_volume":"gv3","monitor_status":"Created","workers":[],"options":{}}

plugins/georeplication/init.go
aravindavk: > Also, mastervolid and slaveid can also be part of request body. Any reason for it being part of URL ?

Since it is unique combination for a georep session and required always. It also makes more sense for other commands start/stop/status/config (Ex: POST /v1/geo-replication/:masterid/:slaveid/start)

aravindavk: I will change this to POST

plugins/georeplication/rest.go
aravindavk: Initially I had that in handler itself, but observed that I have to repeat the stuff almost for all Georep API calls. (MasterVol, SlaveHosts and SlaveVol) validation is required only for Create/Update. For all the other apis Start/Stop/Delete/Status/Config we don't need to validate these fields except Master ID and Slave ID

plugins/georeplication/rest.go
aravindavk: Ok

plugins/georeplication/rest.go
aravindavk: There are some todos for Geo-rep create steps. I think I need to create a working directory in all Volume nodes before starting. I will work on those steps later.

plugins/georeplication/rest.go
aravindavk: Ok, I will move this to `txn`

plugins/georeplication/api/req.go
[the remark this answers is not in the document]
prashanthpai: In that case, you can set it as [omitempty](https://golang.org/pkg/encoding/json/#Marshal)

plugins/georeplication/api/resp.go
prashanthpai: > Is that libray already part of glusterd2?

It's part of standard go tools and isn't an importable library. The volgen PR has example on how to use it. The stringer tool will generate code. All you have to do it include a one line comment in the go source file where you have your types (enum-like consts) defined. Example:

```go
//go:generate stringer -type=GeorepStatus
```

aravindavk: I will include this in another patch later.

plugins/georeplication/api/req.go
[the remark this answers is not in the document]
aravindavk: omitempty is required for response struct right? In this case if SlaveUser is not passed then it will be empty string.

plugins/georeplication/api/resp.go
```diff
@@ -54,3 +54,22 @@ type GeorepSession struct {
 	Workers    []GeorepWorker    `json:"workers"`
 	Options    map[string]string `json:"options"`
 }
+
+// NewGeorepSession creates new instance of GeorepSession
+func NewGeorepSession(mastervolid uuid.UUID, slavevolid uuid.UUID, req GeorepCreateReq) *GeorepSession {
```
prashanthpai: This shouldn't be part of `api` package though which is supposed to contain only API requests and responses. In future, I hope we will be able to generate API doc from the package or generate package code from the API spec.

You can place this function near the handler. And if it's only invoked there, you can as well choose not export it.

aravindavk:
@prashanthpai updated the patch, please check

PR #389, opened 2017-10-13
plugins/georep: Geo-replication session start

aravindavk:
REST API:

    POST /v1/geo-replication/:mastervolid/:slavevolid/start

This API depends on the Geo-rep refactoring
patch(https://review.gluster.org/18257), not added any
tests since start of gsyncd will fail.

I need to find best way to not hard code the path of gsyncd
binary. That can be addressed in different patch.

Updates: #271

### DIFF
--- a/plugins/georeplication/transactions.go
+++ b/plugins/georeplication/transactions.go
@@ -1,8 +1,15 @@
 package georeplication
 
 import (
+	"github.com/gluster/glusterd2/daemon"
 	georepapi "github.com/gluster/glusterd2/plugins/georeplication/api"
 	"github.com/gluster/glusterd2/transaction"
+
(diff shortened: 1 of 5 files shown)


plugins/georeplication/transactions.go
```diff
@@ -1,8 +1,15 @@
...
 import (
+	"github.com/gluster/glusterd2/daemon"
 	georepapi "github.com/gluster/glusterd2/plugins/georeplication/api"
 	"github.com/gluster/glusterd2/transaction"
+
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	gsyncdStartMaxRetries = 10
```
prashanthpai: This isn't needed as gsyncd may not run into port clashes unlike bricks.

PR #390, opened 2017-10-16
doc: Update usage of UUID in volume create

prashanthpai:


prashanthpai:
@kmohanan Check if this update needs more information.

PR #391, opened 2017-10-18
Start daemons on GD restart

kshlm:


main.go
```diff
@@ -97,6 +98,9 @@ func main() {
 	super.Add(servers.New())
 	addMgmtService(super)
 
+	// Restart previously running daemons
+	daemon.StartAllDaemons()
```
prashanthpai: Some of the rest handlers might have to communicate with these daemons and as soon as the rest server is up, glusterd2 will accept incoming requests. So it'll be safer to start the daemons first and then start supervisor chain of servers.

kshlm: The daemons need the sunrpc server to be up though. I think its acceptable that calls can fail for a little while.

kshlm:
> LGTM 👍 I have one requested one small change.
> It would also be nice to have a e2e test case for this with bricks as the example daemon.

It will be much easier to test this if volume status were working. Without it, I'll need to shell out a lot to test this.

prashanthpai:
retest this please

kshlm:
retest this please

kshlm:
retest this please

PR #400, opened 2017-11-02
Switch to dep

kshlm:
'dep' is the official dependency management experiment of the Golang
community. It is expected to become the official tool and the latest glide
release encourages users to switch to dep.

Some quick help,
- `dep ensure` installs dependencies that have been locked
- `dep ensure -upate` updates dependencies to the latest versions and
updates the lock file
- `dep ensure -add <package>` adds a new package.

`make vendor-install` and `make vendor-update` still work as expected.

[1]: https://github.com/golang/dep
[2]: https://github.com/Masterminds/glide/releases/tag/v0.13.0


kshlm:
DO NOT MERGE YET.

PR is functional and ready for general development use. But as `dep` isn't yet available as a package for either Fedora or CentOS.

Till a work around can be implemented don't merge this.

kshlm:
retest this please

kshlm:
`make test` keeps failing in the CI with some weird error about not being able to find packages. That is strange. It works well enough on my laptop. I'm setting up a new centos vm to test it in a similar environment.

prashanthpai:
> 05:40:34 can't load package: package github.com/gluster/glusterd2/vendor/github.com/prashanthpai/sunrpc/examples/arith: C source files not allowed when not using cgo or SWIG: arith_clnt.c arith_svc.c arith_xdr.c c-client.c c-server.c

I've seen this warning before. It's not a big deal but I can fix it if that's a blocker.

prashanthpai:
@kshlm The latest commit adds back `glide.*` files.

kshlm:
> @kshlm The latest commit adds back glide.* files.

I've fixed it now.
And that should be last of the changes I need to do.
Once this merged, I will tag a new release, and try to build rpms.

scripts/dist.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# Create the VERSION file first
+$(dirname $0)/gen-version.sh
+
+# Use a temp dir to build tar file
+TMPDIR=$(mktemp -d)
+
+echo "Creating dist tarball $ARCHIVE"
+# Exclude .git, releases, builds and older tarballs
+tar --exclude .git --exclude releases --exclude build --exclude 'glusterd2-*.tar.xz*' $EXCLUDEVENDOR --transform "s/^\./$TARNAME/" -cf $TMPDIR/$TARBALL . || exit 1
```
prashanthpai: This can become problematic when these list of dirs/files to be excluded grows or is different on my machine. Would you prefer to use `git-archive` instead which is cleaner ?

kshlm: The reason for creating the dist command was to also package vendored packages. The vendored packages aren't in git, so git-archive wouldn't contain them. So I started from with this to archive everything, and went back excluding things as required.

It should be possible create a git-archive tar and then add the vendored package to it. I'll see what I can do.

prashanthpai: In the new commits, did you intentionally leave out including the `VERSION` file from tarball ?

kshlm: That was unintended.

PR #401, opened 2017-11-03
rest: Authentication middleware

aravindavk:
Added REST API authentication middleware. On Glusterd2 start, it
generates `$WORKDIR/auth` file with random string if not exists.

Local CLI/Curl can use that auth file and add Authorization request.

Note: User management is not covered in this patch. Limited to local
consumers only who can access `$WORKDIR/auth`. Auth is disabled by
default which can be enabled using config file option `restauth=true`

CLI needs new flag `--auth-file` to accept path for auth file similar
to Glusterd1 CLI flag `glusterd-sock`

Updates: #252

### DIFF
--- a/middleware/auth.go
+++ b/middleware/auth.go
@@ -1,0 +1,92 @@
+package middleware
+
+import (
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/gluster/glusterd2/gdctx"
+
+	"github.com/dgrijalva/jwt-go"
+)
+
+const (
+	internalUser = "glustercli"
+)
+
+var (
+	requiredClaims = []string{"iss", "exp"}
+)
+
+func getAuthSecret(issuer string) string {
+	if issuer == internalUser {
+		return gdctx.LocalAuthToken
+	}
+
+	// TODO: Look for issuer secret in etcd if not internal user, this depends on User management feature
+
+	return ""
+}
+
+// Auth is a middleware which authenticates HTTP requests
+func Auth(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// If Auth disabled Return as is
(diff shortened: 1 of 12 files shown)


prashanthpai:
Reviewing this. Please do not merge yet.

middleware/auth.go
```diff
@@ -0,0 +1,92 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/gluster/glusterd2/gdctx"
+
+	"github.com/dgrijalva/jwt-go"
+)
+
+const (
+	internalUser = "glustercli"
```
prashanthpai: Anyone can create CLI of their own using this string and `restclient` package. I'm not sure how useful or required this is. Am I missing something here ?

P.S: Every client must be treated the same way by glusterd2- `glustercli` or tendrl or any custom client written by someone. One way to define "internal" is by checking if the client is part of "Trusted Storage Pool" or co-located on the same node etc.

aravindavk: Request from same trusted pool doesn't mean that user can be trusted. "glustercli" can be made as reserved username when we have users/apps management. 

I think new CLI/consumer should register with its own username and password instead of allowing access to the auth file to external users. Let me know your thoughts.

prashanthpai: > "glustercli" can be made as reserved username when we have users/apps management.

It shouldn't be is what I'm hinting at. If this is a POC/prototype user which will be removed in future, then it's okay.

aravindavk: "glustercli" will be there to identify default user. Any user creation will be stored in etcd. Username will be specified in payload of Authorization token, if it is internal user then secret will be fetched from auth file, else fetch secret from etcd.

prashanthpai: So, default user = internal user ?

IIUC, this default/internal user will be present before first "real" user is created and stored. It's like an admin. Is it ?

PR #405, opened 2017-11-07
e2e: Add test for volume expand operation.

kmlmhnn:

### DIFF
--- a/e2e/volume_ops_test.go
+++ b/e2e/volume_ops_test.go
@@ -45,6 +45,9 @@ func TestVolume(t *testing.T) {
 	// Create the volume
 	t.Run("Create", testVolumeCreate)
 
+	// Expand the volume
+	t.Run("Expand", testVolumeExpand)
+
 	// Run tests that depend on this volume
 	t.Run("Start", testVolumeStart)
 	t.Run("Mount", testVolumeMount)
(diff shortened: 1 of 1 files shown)


prashanthpai:
test this pr

e2e/volume_ops_test.go
```diff
@@ -79,6 +82,29 @@ func testVolumeCreate(t *testing.T) {
...
+		brickPaths = append(brickPaths, brickPath)
+	}
+
+	expandReq := api.VolExpandReq{
+		ReplicaCount: 2,
+		Bricks: []string{
+			gds[0].PeerID() + ":" + brickPaths[0],
+			gds[1].PeerID() + ":" + brickPaths[1],
+			gds[0].PeerID() + ":" + brickPaths[2],
+			gds[1].PeerID() + ":" + brickPaths[3],
```
kshlm: You're using the exact same bricks as the ones that were used to create the volume. So ideally this would be a negative test case.
Keep the negative case, and add a new positive case, with new bricks.

kshlm:
add to whitelist

e2e/volume_ops_test.go
kmlmhnn: But doesn't ioutil.TempDir create new temporary directories each time?

kshlm: You're right. My bad.

PR #406, opened 2017-11-08
Volume operations test

kmlmhnn:


pkg/restclient/volume.go
```diff
@@ -28,6 +29,14 @@ func (c *Client) Volumes(volname string) ([]api.Volinfo, error) {
 	return []api.Volinfo{vol}, err
 }
 
+// VolumeStatus returns the status of a Gluster volume
+func (c *Client) VolumeStatus(volname string) (volume.VolStatus, error) {
+	url := fmt.Sprintf("/v1/volumes/%s/status", volname)
+	var volStatus volume.VolStatus
```
prashanthpai: The CLI/restclient should never import anything from `github.com/gluster/glusterd2/volume`. We should properly define request and response structs in the `api` package. I'll merge this one for now.

PR #410, opened 2017-11-09
Remove brick entry from store on volume stop

### LINKED ISSUE #408: gd2 starts bricks on restart even if volume is in stopped state

prashanthpai:
This should prevent glusterd2 from spawning bricks of volume that is
in stopped state, during glusterd2's restart.

Fixes #408


prashanthpai:
> Changes look fine. Can we cover this scenario in e2e/restart_test.go?

I wanted to do it. But it requires a small refactor in volume status output. Is it okay if I add it after those changes ?

kshlm:
> I wanted to do it. But it requires a small refactor in volume status output. Is it okay if I add it after those changes ?

Sounds fine to me. I'll merge this now.

PR #413, opened 2017-11-10
volgen: Add gfproxy-client and gfproxyd volgen code

poornimag:


poornimag:
I have few doubts/requests:

1. I get these errors when i run glusterd2:
patchy-client
unknown variable string: brick.hostname
patchy-server
unknown variable string: brick.path
This is while generating gfproxy-client volfile, that is because, the default value of
remote-host in protocol-client is "{{ brick.hostname }}" and
remote-subvolume in protocol-client is "{{ brick.path }}"

In case of gfproxy -client volfile, the the remote-host is not filled in volfile, and remote-subvolume is "gfproxyd-<VOLNAME>" which is not brick.path.

So how do i handle setting a specific value for remote-subvolume?

2. There are scenarios where the io-stats is named as <VOLUMENAME> instead of <VOLUMENAME>-io-stats. This is required for remote-subvolume search. How do we do this?

poornimag:
Thanks @prashanthpai for the help with the patch.

kshlm:
add to whitelist

kshlm:
> I have few doubts/requests:
> 
> I get these errors when i run glusterd2:
> patchy-client
> unknown variable string: brick.hostname
> patchy-server
> unknown variable string: brick.path
> This is while generating gfproxy-client volfile, that is because, the default value of
> remote-host in protocol-client is "{{ brick.hostname }}" and
> remote-subvolume in protocol-client is "{{ brick.path }}"
> In case of gfproxy -client volfile, the the remote-host is not filled in volfile, and remote-subvolume is "gfproxyd-" which is not brick.path.
> 
> So how do i handle setting a specific value for remote-subvolume?
> 
> There are scenarios where the io-stats is named as instead of -io-stats. This is required for remote-subvolume search. How do we do this?

The graph generate method has an `extra` argument which is a [string]string map. You can create a map, with keys brick.hostname and brick.client set to the values you want and pass it to the generate function.

volgen/volgen.go
```diff
@@ -52,16 +58,42 @@ func GenerateClientVolfile(vol *volume.Volinfo) error {
...
-		return err
-	}
-	if _, err := store.Store.Put(context.TODO(), volfilePrefix+vol.Name, buf.String()); err != nil {
-		return err
-	}
+	return cg.WriteToFile(getGfproxydVolFilePath(vol.Name))
+}
+
+// GenerateClientVolfile generates the client volfile and stores it in etcd
+func GenerateClientVolfile(vol *volume.Volinfo) error {
```
kshlm: It will be better to refactor this into a couple of functions, `GenerateClientVolfiles` and `GenerateClientVolfile`.

`GenerateClientVolfile` should accept volinfo, template, volfile path and extra map as arguments. `GenerateClientVolfiles` can then call `GenerateClientVolfile` with required params.

This applies to `DeleteClientVolfile` as well.

volgen/templates/gfproxy.graph
```diff
@@ -0,0 +1,10 @@
+protocol/server
+debug/io-stats
```
kshlm: To name io-stats as "gfproxy-*" or with volname, instead of the normal "<volname-io>-stats", add the name you want here. For eg. this line could be,
`debug/io-stats, gfpoxy-{{ volname }}`

volgen/volgen.go
```diff
@@ -40,9 +46,9 @@ func Generate(vol *volume.Volinfo) error {
 	return nil
 }
 
-// GenerateClientVolfile generates the client volfile and stores it in etcd
-func GenerateClientVolfile(vol *volume.Volinfo) error {
-	ct, err := GetTemplate(fuseTmpl, vol.GraphMap)
+// GenerateGfproxydVolfile generates the gfproxyd volfile and stores it in etcd
+func GenerateGfproxydVolfile(vol *volume.Volinfo) error {
```
kshlm: Please put the GFproxy related volgen code in a separate file. Makes it easier to maintain.

volgen/templates/gfproxy.graph
poornimag: Done

volgen/volgen.go
poornimag: Done

volgen/volgen.go
poornimag: After making the below mentioned changes, there were only 2 4-line functions remaning, that were specific to gfproxy, should i move them to different file? The best would be to have plugins for volgen as well?

glusterd2/volgen/static.go
```diff
@@ -58,4 +58,23 @@ protocol/client`,
...
 protocol/client`,
 	},
+	{
+		name: "gfproxy-fuse.graph",
+		content: `debug/io-stats
+performance/write-behind
+protocol/client`,
+	},
+	{
+		name: "gfproxy.graph",
```
prashanthpai: Shouldn't this be `gfproxyd.graph` as defined at `gfproxydTmpl` ?

volgen/templates/gfproxy.graph
```diff
@@ -0,0 +1,10 @@
+protocol/server
```
prashanthpai: The file name. Shouldn't it be `gfproxyd.graph` ?

poornimag: Yes, missed it

glusterd2/volgen/static.go
poornimag: Done

aravindavk:
Alternate approach for volgen is implemented in https://github.com/gluster/glusterd2/pull/497, which supports cluster volfiles, volume volfiles and brick volfiles. gfproxy volfiles are also available with that patch. Please check.

poornimag:
@aravindavk Oh ok, will check it out. But i have on e question, not related to gfproxy, but parallel-readdir. The position of readdir-ahead xlator is either above dht or below dht based on the a volume set option, that will be more easier with this patch i believe?

aravindavk:
> @aravindavk Oh ok, will check it out. But i have on e question, not related to gfproxy, but parallel-readdir. The position of readdir-ahead xlator is either above dht or below dht based on the a volume set option, that will be more easier with this patch i believe?

Check this below file(gfproxy example), for any new volfile support just add one new file like this
https://github.com/gluster/glusterd2/pull/497/files#diff-881216ff8e45eb669f7745c694654537

Gopkg.lock
```diff
@@ -348,6 +348,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "dc95f6a3bf9f2aa98d30b085c01eefb0ad1023a1f3e67d64563d64558d7871f1"
+  inputs-digest = "1afa5e0c8fa1ae1a19b5d3acf7256192943a5dae20bb0cf84906bc9e1c777d1b"
```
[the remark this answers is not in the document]
prashanthpai: Right. This shouldn't be there as no new dependency was added.

aravindavk:
@poornimag let us know if any help needed for this patch.

glusterd2/volume/struct.go
```diff
@@ -41,6 +41,8 @@ const (
 	VkeyScrubFrequency = "bit-rot.scrub-freq"
 	// VkeyScrubThrottle is the key for controls scrubber throttle
 	VkeyScrubThrottle = "bit-rot.scrub-throttle"
+	//VkeyFeaturesGfproxyd is the key which enables or disables gfproxyd
+	VkeyFeaturesGfproxyd = "config.gfproxyd"
```
aravindavk: Unrelated to your patch.

@prashanthpai, @kshlm what is the best way to maintain a feature is enabled or not in volinfo?

plugins/gfproxyd/gfproxyd.go
```diff
@@ -0,0 +1,92 @@
...
+func newgfproxyd(volname string) (*gfproxyd, error) {
+	g := &gfproxyd{volname: volname}
+	binarypath, e := exec.LookPath(gfproxydBin)
+	if e != nil {
+		return nil, e
+	}
+	g.binarypath = binarypath
+	g.gfproxyID = fmt.Sprintf("gfproxyd-%s", volname)
+	g.gfproxydPort = strconv.Itoa(pmap.AssignPort(0, g.gfproxyID))
+	g.volfileID = fmt.Sprintf("%s.gfproxyd", volname)
```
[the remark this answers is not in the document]
aravindavk: Change volfileid to "gfproxyd/<volname>"

plugins/gfproxyd/rest.go
```diff
@@ -0,0 +1,142 @@
...
+		logger.WithFields(log.Fields{
+			"error":   err.Error(),
+			"volname": volname,
+		}).Error("failed to start gfproxy daemon")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err.Error(), api.ErrCodeDefault)
+		return
+	}
+
+	// Enable gfproxyd
+	volinfo.Options[volume.VkeyFeaturesGfproxyd] = "on"
```
aravindavk: This is not persistent. since not saved in store, you can add "storeVolume" transaction step. Ref: https://github.com/gluster/glusterd2/blob/master/plugins/bitrot/rest.go#L155

plugins/gfproxyd/rest.go
```diff
@@ -0,0 +1,142 @@
...
+		logger.WithFields(log.Fields{
+			"error":   err.Error(),
+			"volname": volname,
+		}).Error("failed to start gfproxy daemon")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err.Error(), api.ErrCodeDefault)
+		return
+	}
+
+	// Disable gfproxyd
+	volinfo.Options[volume.VkeyFeaturesGfproxyd] = "off"
```
aravindavk: same as above, not persistent.

poornimag: Done

plugins/gfproxyd/rest.go
poornimag: Done

aravindavk: Need to be updated before `txt.Do` and make volinfo available to "vol-option.UpdateVolinfo" using Ctx.Set (Before `txn.Do`)

    volinfo.Options[volume.VkeyFeaturesGfproxyd] = "on"

    if err := txn.Ctx.Set("volinfo", volinfo); err != nil {
    		logger.WithError(err).Error("failed to set volinfo in transaction context")
    		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err.Error(), api.ErrCodeDefault)
    		return
    }

poornimag: And i should reset it to off, if it fails to do the transaction?.

aravindavk: You can add rollback function later which can unset from volinfo.

poornimag: And i should reset it to off, if it fails to do the transaction?

glusterd2/volume/struct.go
aravindavk: This can be changed as "features.gfproxyd".

plugins/gfproxyd/rest.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	// Transaction which starts gfproxyd on all nodes.
+	txn := transaction.NewTxn(ctx)
+	defer txn.Cleanup()
+
+	txn.Ctx.Set("volname", volname)
+
+	// Enable gfproxyd
+	volinfo.Options[volume.VkeyFeaturesGfproxyd] = "on"
```
aravindavk: After this step, `txn.Ctx.Set("volinfo"` is required for  "vol-option.UpdateVolinfo" transaction step

plugins/gfproxyd/rest.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	// Transaction which stop gfproxyd on all nodes.
+	txn := transaction.NewTxn(ctx)
+	defer txn.Cleanup()
+
+	txn.Ctx.Set("volname", volname)
+
+	// Disable gfproxyd
+	volinfo.Options[volume.VkeyFeaturesGfproxyd] = "off"
```
aravindavk: Same as above. Ctx.Set "volinfo" is required

aravindavk:
@poornimag rebase required.

aravindavk:
Resolved conflict from UI.

glusterd2/volume/struct.go
```diff
@@ -46,6 +46,8 @@ const (
 	VkeyScrubFrequency = "bit-rot.scrub-freq"
 	// VkeyScrubThrottle is the key for controls scrubber throttle
 	VkeyScrubThrottle = "bit-rot.scrub-throttle"
+	//VkeyFeaturesGfproxyd is the key which enables or disables gfproxyd
+	VkeyFeaturesGfproxyd = "features.gfproxyd"
```
kshlm: I don't know when this started, but features shouldn't be modifying core packages to add their stuff. Ideally, to add these sorts of keys, the core package should provide some sort of interface that plugins use to add things they want.

As I see, currently this isn't used else where out of the gfproxyd plugin. So you can just move it to the plugin.

pkg/errors/error.go
```diff
@@ -38,5 +38,7 @@ var (
 	ErrUnknownValue            = errors.New("unknown value specified")
 	ErrGetFailed               = errors.New("failed to get value from the store")
 	ErrUnmarshallFailed        = errors.New("failed to unmarshall from json")
+	ErrGfproxydAlreadyEnabled  = errors.New("Gfproxyd is already enabled")
```
kshlm: This package is something I'd like to get rid of. Please keep the errors related to a package, within the package. This should be in the `gfproxyd` plugin package, in a seperate `errors.go` preferably.

Also, (again) plugins shouldn't be modifying core packages.

glusterd2/volume/volume-utils.go
```diff
@@ -45,3 +45,12 @@ func IsQuotaEnabled(v *Volinfo) bool {
 	}
 	return false
 }
+
+// IsGfproxydEnabled returns true if gfproxyd is enabled for a volume and false otherwise
+func IsGfproxydEnabled(v *Volinfo) bool {
```
kshlm: This should be in the gfproxyd package. Again, plugins shouldn't be modifying core packages.

plugins/gfproxyd/gfproxyd.go
```diff
@@ -0,0 +1,92 @@
...
+func newgfproxyd(volname string) (*gfproxyd, error) {
+	g := &gfproxyd{volname: volname}
+	binarypath, e := exec.LookPath(gfproxydBin)
+	if e != nil {
+		return nil, e
+	}
+	g.binarypath = binarypath
+	g.gfproxyID = fmt.Sprintf("gfproxyd-%s", volname)
+	g.gfproxydPort = strconv.Itoa(pmap.AssignPort(0, g.gfproxyID))
+	g.volfileID = fmt.Sprintf("gfproxyd/%s", volname)
```
kshlm: I'm assuming the gfproxy volfile is already generated. Right?

plugins/gfproxyd/rest.go
kshlm: You shouldn't be directly setting options in a volinfo, without trying to verify that it is supported on all peers. The volume set step functions could be used to do the same. I don't know if its possible right now, but a TODO here will help remind us later.

glusterd2/volume/struct.go
aravindavk: Volfile generation may use it. This is common problem for many features. For example bitrot. Unless we provide a way to store enabled features this we can't avoid.

I asked the same question here https://github.com/gluster/glusterd2/pull/413#discussion_r167149739

I feel we should allow this till we implement infra to store this info.

plugins/gfproxyd/gfproxyd.go
aravindavk: Yes

poornimag:
Can this be merged before 4.0?

plugins/gfproxyd/transactions.go
```diff
@@ -0,0 +1,38 @@
+package gfproxyd
+
+import (
+	"github.com/gluster/glusterd2/glusterd2/daemon"
+	"github.com/gluster/glusterd2/glusterd2/transaction"
+)
+
```
prashanthpai: If you prefer, you can separate the common code in these functions into separate function. [Here](https://github.com/gluster/glusterd2/blob/517a731e829fce24807cdb3b82d13b9809307eda/plugins/glustershd/transactions.go#L47-L53) is an example.

plugins/gfproxyd/rest.go
```diff
@@ -0,0 +1,164 @@
...
+	if !volume.IsGfproxydEnabled(volinfo) {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, errors.ErrGfproxydAlreadyDisabled.Error(), api.ErrCodeDefault)
+		return
+	}
+
+	// Transaction which stop gfproxyd on all nodes.
+	txn := transaction.NewTxn(ctx)
+	defer txn.Cleanup()
+
+	txn.Ctx.Set("volname", volname)
```
prashanthpai: volname is part of volinfo which you set below in the txn context

plugins/gfproxyd/rest.go
```diff
@@ -0,0 +1,164 @@
...
+		},
+		unlock,
+	}
+
+	err = txn.Do()
+	if err != nil {
+		logger.WithFields(log.Fields{
+			"error":   err.Error(),
+			"volname": volname,
+		}).Error("failed to start gfproxy daemon")
```
prashanthpai: "failed to `stop` gfproxy daemon"

plugins/gfproxyd/init.go
```diff
@@ -0,0 +1,46 @@
...
+			Pattern:     "/volumes/{name}/gfproxy/disable",
+			Version:     1,
+			HandlerFunc: gfproxydDisableHandler},
+	}
+}
+
+// RegisterStepFuncs registers transaction step functions with
+// Glusterd Transaction framework
+func (p *Plugin) RegisterStepFuncs() {
+	transaction.RegisterStepFunc(txnGfproxydStart, "gfproxyd-start.Commit")
```
prashanthpai: Lose the `Commit` suffix from the names.

plugins/gfproxyd/gfproxyd.go
```diff
@@ -0,0 +1,92 @@
...
+	g.logfile = path.Join(config.GetString("logdir"), "glusterfs", fmt.Sprintf("gfproxyd-%s.log", volname))
+	g.pidfilepath = fmt.Sprintf("%s/gfproxyd-%s.pid", config.GetString("rundir"), volname)
+
+	shost, _, _ := net.SplitHostPort(config.GetString("clientaddress"))
```
prashanthpai: Glusterd2 can be configured to listen on any port and this code needs to take that into consideration.

```go
shost, sport, _ := net.SplitHostPort(config.GetString("clientaddress"))
...
buffer.WriteString(fmt.Sprintf(" --volfile-server-port %s", sport))
```

You can refer to usage of the above code in `glusterd2/brick/glusterfsd.go`

plugins/gfproxyd/rest.go
```diff
@@ -0,0 +1,164 @@
...
+	if volume.IsGfproxydEnabled(volinfo) {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, errors.ErrGfproxydAlreadyEnabled.Error(), api.ErrCodeDefault)
+		return
+	}
+
+	// Transaction which starts gfproxyd on all nodes.
+	txn := transaction.NewTxn(ctx)
+	defer txn.Cleanup()
+
+	txn.Ctx.Set("volname", volname)
```
prashanthpai: You can use `Name` field of volinfo which you already set in txn set.

prashanthpai:
> Can this be merged before 4.0?

@poornimag  Can you address review comments ?

prashanthpai:
@poornimag Can you please rebase and address the earlier review comments ?

amarts:
@poornimag is it possible to look into this, and address final comments? Any further help required?

glusterd2/volume/struct.go
aravindavk: As @kshlm suggested, move this to `plugins/gfproxyd`

glusterd2/volume/volume-utils.go
aravindavk: Same here, move this function to `plugins/gfproxyd`

pkg/errors/error.go
aravindavk: Same here

plugins/gfproxyd/gfproxyd.go
aravindavk: This is required, since tests will be run with different ports for glusterd2

plugins/gfproxyd/rest.go
aravindavk: For now this behavior is fine for internal options.

aravindavk:
@poornimag please rebase this and address review comments.

plugins/gfproxyd/gfproxyd.go
```diff
@@ -0,0 +1,92 @@
...
+	if shost == "" {
+		shost = "localhost"
+	}
+
+	var buffer bytes.Buffer
```
aravindavk: This needs change now, convert this to string list

plugins/gfproxyd/init.go
```diff
@@ -0,0 +1,46 @@
...
+type Plugin struct {
+}
+
+// Name returns name of plugin
+func (p *Plugin) Name() string {
+	return "gfproxyd"
+}
+
+// SunRPCProgram returns sunrpc program to register with Glusterd
+func (p *Plugin) SunRPCProgram() sunrpc.Program {
```
aravindavk: This function can be removed now, made changes to plugin interface

PR #415, opened 2017-11-13
cli: Make log file and log level configurable via CLI options.

kmlmhnn:

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"fmt"
+
 	"github.com/spf13/cobra"
 )
 
@@ -9,6 +11,10 @@ var RootCmd = &cobra.Command{
 	Use:   "gluster",
 	Short: "Gluster Console Manager (command line utility)",
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
+		err := initLog(flagLogFile, flagLogLevel)
+		if err != nil {
+			fmt.Println("Error initializing log file ", err)
+		}
 		initRESTClient(flagHostname, flagCacert, flagInsecure)
 	},
 }
@@ -19,6 +25,8 @@ var (
 	flagHostname   string
 	flagCacert     string
 	flagInsecure   bool
+	flagLogFile    string
+	flagLogLevel   string
 )
 
 func init() {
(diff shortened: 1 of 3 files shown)


kmlmhnn:
Logging messages from argsMigrate will not work as expected since it is called before initializing the log file.
Also the the log file and log level options specified are not persistent. Therefore they have to be specified each time glustercli is invoked if non-default values are required.

cli/cmd/root.go
```diff
@@ -27,6 +35,10 @@ func init() {
 	RootCmd.PersistentFlags().BoolVarP(&flagJSONOutput, "json", "", false, "JSON Output")
 	RootCmd.PersistentFlags().StringVarP(&flagHostname, "host", "", "http://localhost:24007", "Host")
 
+	// Log options
+	RootCmd.PersistentFlags().StringVarP(&flagLogFile, "log", "", "./cli.log", "Log file")
```
prashanthpai: nit: Please name this as `logfile`

cli/main.go
```diff
@@ -2,69 +2,16 @@ package main
...
-		setLogOutput(os.Stderr)
-		log.WithError(err).Debug("Failed to open log file %s", logFilePath)
-		return err
-	}
-	setLogOutput(logFile)
-	logWriter = logFile
-	return nil
-}
 func main() {
-	err := initLog(defaultLogFile, defaultLogLevel)
```
kshlm: I'd like it for logging intialization to remain here. Log should be intialized as early as possible, so that we can catch all logs. Doing this would allow logs for `argsMigrate()` as well.

I know trying to do this with the cobra package will not be easy, as it seems that the first level of flags can be added only to the root command. But can you explore if it is possible to add and make use of flags before a root command is created?

If it isn't possible, we can merge this as it is now.

aravindavk: I feel we can skip that part now since we can print the errors to stderr(or as user friendly error message) in case we need to log something in argsMigrate.

I think adding separate parsing only for these two fields may not add value. Let me know your thoughts.

kshlm: There need not be separate parsing just for these two flags. The flags can be defined in RootCmd, but we could make use of it here.
I'm consolidating the logging intialization into a single package with my refactorings. I can take a shot at doing this with it. I'm okay with merging this for now.

@kmlmhnn fix @prashanthpai's comment, and we can get this merged.

PR #422, opened 2017-11-14
Separate internal structs from the ones sent to client

### LINKED ISSUE #407: Move all request and response structs to api package
...and differentiate internally used structs (like volinfo, volstatus) from the structs sent as response to users.

prashanthpai:
**Why?**

Because we want to limit what we expose over REST API. The internal structs that we store in etcd can have internal states or attributes that we shouldn't expose to the client.

This PR does not introduce any functional changes.

Closes #407


commands/volumes/volume-status.go
```diff
@@ -69,8 +77,8 @@ func checkStatus(ctx transaction.TxnCtx) error {
 			}
 		}
 
-		brickStatus := &brick.Brickstatus{
-			BInfo:  binfo,
+		brickStatus := &brickstatus{
```
kshlm: Shouldn't you be using `api.BrickStatus` instead of a this local struct?

prashanthpai: This is is an internal struct used only during the scope of the transaction.

kshlm: Why though? Both the structs have the same members.

kshlm: Okay. I see that the Info members are different, (api.Brickinfo vs brick.Brickinfo). I'm approving the PR.

PR #436, opened 2017-11-20
WIP: Add validation functions for the available option types.

devyanikota:

### DIFF
--- a/glusterd2/xlator/options/options.go
+++ b/glusterd2/xlator/options/options.go
@@ -1,5 +1,14 @@
 package options
 
+import (
+	"errors"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	validate "github.com/asaskevich/govalidator"
+)
+
 // OptionType is a type which represents the type of xlator option.
 type OptionType int
 
@@ -48,6 +57,9 @@ const (
 	OptionFlagDoc
 )
 
+// ErrInvalidArg is an Invalid Argument error
+var ErrInvalidArg = errors.New("Invalid Argument")
+
 // Option is a struct which represents one single xlator option exported by
 // the translator.
 type Option struct {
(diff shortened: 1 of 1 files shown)


xlator/options.go
```diff
@@ -1,26 +1,32 @@
...
-	OptionTypeStr
-	OptionTypeInt
-	OptionTypeSizet
-	OptionTypePercent
-	OptionTypePercentOrSizet
 	OptionTypeBool
-	OptionTypeXlator
-	OptionTypePath
-	OptionTypeTime
+	OptionTypeClientAuthAddr
```
kshlm: The order here needs to match the order of the `volume_option_type_t` enum in options.h. Don't sort it alphabetically.

xlator/options.go
```diff
@@ -65,3 +71,176 @@ type Option struct {
...
+		err = ValidateTime(o, val)
+	case OptionTypeXlator:
+		err = ValidateXlator(o, val)
+	default:
+		err = ValidateOption(o, val)
+	}
+	return err
+}
+
+// ValidateBool validates if the option is of type boolean
```
kshlm: Could you please move all the validation functions into a single file?

xlator/options.go
```diff
@@ -65,3 +71,176 @@ type Option struct {
...
+func ValidateBool(o *Option, val string) error {
+	if val == "" {
+		err := errors.New("No argument passed")
```
kshlm: It is better to define common error types for common validation errors, instead of creating new errors every time.

xlator/options.go
```diff
@@ -65,3 +71,176 @@ type Option struct {
...
+// ValidateInternetAddressList validates the Internet Address List
+func ValidateInternetAddressList(o *Option, val string) error {
+	err := errors.New("Argument invalid")
+	return err
+}
+
+// ValidatePath validates if the option is a valid path
+func ValidatePath(o *Option, val string) error {
+	var err error
+	if strings.Contains(val, "../") == true {
```
kshlm: You can use the functions in the `filepath` package to do this validation.

xlator/options.go
```diff
@@ -65,3 +71,176 @@ type Option struct {
...
+		err = errors.New("Invalid percent format")
+	} else if l > 1 && strings.ContainsRune(val, '%') {
+		if val[0] == '%' {
+			err = errors.New("Invalid percent format")
+		}
+		percent, err = strconv.ParseFloat(val[:(l-1)], 64)
+		if err != nil {
+			return err
+		}
+	}
```
kshlm: This should accept percents with and without the percent symbol. The current validation fails if a the symbol isn't present.

xlator/options.go
```diff
@@ -65,3 +71,176 @@ type Option struct {
...
+	return err
+}
+
+// ValidateBool validates if the option is of type boolean
+func ValidateBool(o *Option, val string) error {
+	if val == "" {
+		err := errors.New("No argument passed")
+		return err
+	} else if strings.EqualFold(val, "1") == true ||
```
kshlm: You could do this in a switch case.
```golang
switch strings.ToLower(val) {
  case "on", "yes", "true",...,"off","no","false",...:
    return nil
  default:
     return err
}
```

xlator/options.go
devyanikota: oops, I wasn't aware of that.
Ack, will do.

xlator/options.go
devyanikota: Ack, will do.

glusterd2/xlator/options.go
```diff
@@ -65,3 +74,152 @@ type Option struct {
...
+	return ErrInvalidArg
+}
+
+// ValidateInternetAddressList validates the Internet Address List
+func ValidateInternetAddressList(o *Option, val string) error {
+	return ErrInvalidArg
+}
+
+// ValidatePath validates if the option is a valid path
+// Use filepath package instead
```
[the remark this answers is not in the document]
devyanikota: Hehe. um, just so I don't forget? I wanted to work on other validations prior this :)
Will update the PR.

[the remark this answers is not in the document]
kshlm: A TODO label would help highlight it better.

kshlm:
@devyanikota Just a heads up about potential merge conflict issues. I have a new PR #448 that creates a new `options` package, by moving out the options stuff from `xlator`. The conflicts should be really easy to solve though. Let me know if you would need any help.

Also, please review the PR.

prashanthpai:
@devyanikota I was planning to use generic validation library such as [govalidator](https://github.com/asaskevich/govalidator) for validating request structs of incoming http requests. See if it's useful here too. It has a long list of [validation functions](https://godoc.org/github.com/asaskevich/govalidator)

glusterd2/xlator/options/options.go
```diff
@@ -70,6 +82,164 @@ type Option struct {
...
+		return ErrInvalidArg
+	}
 	return nil
 }
+
+// ValidateInt validates if the option is of type Int
+func ValidateInt(o *Option, val string) error {
+	if validate.IsInt(val) != true {
+		return ErrInvalidArg
+	}
```
kshlm: Good. Along with validating if the value can be parsed as the correct type, you will also need to validate if it is in the right range.

If either `o.Min` or `o.Max` are not zero, it should be ensured that the value is in the range. Depending on `o.ValidateType` you will need to check both min and max (default) or only min or only max.

The same needs to be done for Double and Size types as well. Have a common function to validate range.

glusterd2/xlator/options/options.go
```diff
@@ -70,6 +82,164 @@ type Option struct {
...
+}
+
+// ValidateStr validates if the option is of type Str
+func ValidateStr(o *Option, val string) error {
+	l := len(val)
+	if l == 0 {
+		return ErrInvalidArg
+	} else if len(strings.TrimSpace(val)) == 0 {
+		return ErrInvalidArg
+	}
```
kshlm: If `o.Value` is not empty, you need to check if the given value is among the valid values.

devyanikota: @kshlm rectified, please check :)

glusterd2/xlator/options/options.go
devyanikota: Added ValidateRange() function to check if option in expected range or not.

glusterd2/xlator/options/options.go
```diff
@@ -70,6 +82,208 @@ type Option struct {
...
+	return nil
+}
+
+// ValidateStr validates if the option is of type Str
+func ValidateStr(o *Option, val string) error {
+	l := len(val)
+	if l == 0 {
+		return ErrInvalidArg
+	} else if len(strings.TrimSpace(val)) == 0 {
```
kshlm: TrimSpace and check length just once. No need to be check length twice.

glusterd2/xlator/options/options.go
```diff
@@ -70,6 +82,208 @@ type Option struct {
...
+		return ErrInvalidArg
+	} else {
+		for _, op := range o.Value {
+			if val != op {
```
kshlm: This will always fail validation unless val is the first o.Value.

glusterd2/xlator/options/options.go
```diff
@@ -70,6 +82,208 @@ type Option struct {
 // Option.
 // Returns are error if it is not possible, nil otherwise.
 func (o *Option) Validate(val string) error {
-	// TODO: Do actual validation
+	var err error
+	switch t := o.Type; t {
```
kshlm: You need no assign to `t` here. Just `switch o.Type` will work.

glusterd2/xlator/options/options.go
```diff
@@ -48,6 +57,9 @@ const (
 	OptionFlagDoc
 )
 
+// ErrInvalidArg is an Invalid Argument error
+var ErrInvalidArg = errors.New("Invalid Argument")
```
kshlm: Invalid value makes much better sense, than invalid arg.

glusterd2/xlator/options/options.go
```diff
@@ -70,6 +82,208 @@ type Option struct {
...
+// ValidateRange validates if option in correctrange.
+func ValidateRange(o *Option, val string) error {
+	v, err := strconv.ParseFloat(val, 64)
+	if err != nil {
+		return err
+	}
+	if o.ValidateType == OptionValidateBoth && o.Min != 0 && o.Max != 0 {
+		return nil
+	} else if o.ValidateType == OptionValidateMin && v < o.Min {
+		err := errors.New("option is smaller than minimum value")
```
kshlm: You can create specific error types for range errors.

glusterd2/xlator/options/options.go
```diff
@@ -70,6 +82,208 @@ type Option struct {
...
+		err := errors.New("option is out of range")
+		return err
+	}
 	return nil
 }
+
+// ValidateBool validates if the option is of type boolean
+func ValidateBool(o *Option, val string) error {
+	if val == "" {
+		err := errors.New("No argument passed")
```
kshlm: A special error for empty values can be created as well.

glusterd2/xlator/options/options.go
devyanikota: Rectified. thank you.

devyanikota:
@kshlm worked on your suggested changes.
Noted: will start writing unit tests.
Thank you for the review :)

glusterd2/xlator/options/options.go
```diff
@@ -212,23 +243,33 @@ func ValidateSizeList(o *Option, val string) error {
...
+		return ErrEmptyArg
+	}
+	for _, op := range o.Value {
+		if val == op {
+			return nil
 		}
-		return nil
+		f = f + 1
+	}
+	if f != 0 {
```
kshlm: If you get out of the loop without finding a match you can just return an error here.

glusterd2/xlator/options/options.go
```diff
@@ -212,23 +243,33 @@ func ValidateSizeList(o *Option, val string) error {
...
+	} else if v < o.Min || v > o.Max {
+		return ErrInvalidRange
+	}
+	return nil
 }
 
 // ValidateStr validates if the option is of type Str
 func ValidateStr(o *Option, val string) error {
-	l := len(val)
+	l := len(strings.TrimSpace(val))
```
kshlm: Save the trimmed string, and use it when checking against expected values.

glusterd2/xlator/options/options.go
```diff
@@ -212,23 +243,33 @@ func ValidateSizeList(o *Option, val string) error {
...
-		return ErrInvalidArg
-	} else if len(strings.TrimSpace(val)) == 0 {
-		return ErrInvalidArg
-	} else {
-		for _, op := range o.Value {
-			if val != op {
-				return ErrInvalidArg
-			}
+		return ErrEmptyArg
+	}
```
kshlm: If `o.Value` is empty `return nil` right here. You can avoid using `f`.

glusterd2/xlator/options/options.go
kshlm: Use `ErrInvalidRange` for these as well.

glusterd2/xlator/options/options.go
devyanikota: eeks! an overlook, am sorry.

prashanthpai:
@devyanikota You missed adding `govalidator` using `dep` as dependency.

devyanikota:
@prashanthpai it's not working for me: 'dep ensure', was unable to add.
Me and Kaushal tried to figure out why but it seems like a fresh setup solves the same.
Am working on a fresh setup, will do it in a different commit. Thanks.

devyanikota:
@prashanthpai I just noticed, you already sent the commit, thanks :)

Issue #438, opened 2017-11-21
Do context plumbing


prashanthpai:
For better propagation of request-scoped artefacts and at a later stage, tracing.

atinmu:
@prashanthpai Can we target it for mvp-4?

PR #441, opened 2017-11-21
Do some initial context plumbling

prashanthpai:
Updates issue #438


glusterd2/transaction/transaction.go
```diff
@@ -31,14 +32,9 @@ type Txn struct {
...
-func NewTxn(id string) *Txn {
+func NewTxn(ctx context.Context) *Txn {
 	t := new(Txn)
-
-	if t.ID = uuid.Parse(id); t.ID == nil {
-		t.ID = uuid.NewRandom()
-		log.WithField("reqid", t.ID.String()).Warn("Invalid UUID set as request ID. Generated new request ID")
-	}
-
+	t.ID = uuid.Parse(ctx.Value(gdctx.ReqIDKey).(string))
```
kshlm: With middlewares like the disk management plugin, we could have multiple transactions launched for a single request. Using req-id as transaction id doesn't make good sense in such a case. We should start using unique ids per transaction.

prashanthpai: I've had the same thought too. I'll do that separately and also have `txnid` derived from `reqid` or come up with some other mechanism to co-relate the two.

Issue #445, opened 2017-11-22
Directory not found during make install


rishubhjain:
I am using centos 7, during "make install" I get an error " install: failed to access ‘/usr/local/share/glusterd2/templates’: No such file or directory " which is because of line https://github.com/gluster/glusterd2/blob/master/Makefile#L66.

kshlm:
Fixed with #446

PR #446, opened 2017-11-22
Update Makefile

rishubhjain:
#445 
Please verify


kshlm:
add to whitelist

kshlm:
@rishubhjain I didn't notice this earlier. But you submitted this PR by creating a branch on this repository. You shouldn't don't do that.

You need to submit PRs from branches on your own fork. Follow instructions on the [development guide](https://github.com/gluster/glusterd2/blob/master/doc/development-guide.md) on how to do this for Go projects.

This was possible because you are a member of glusterd2-developers team, who have write permissions on this repo. But the permissions were only given so that you could assign issues to yourself. I would like to prevent this from happening, but for now Github doesn't have enough permissions to do it. So it developers need to be careful with what their pushes.

PR #448, opened 2017-11-22
Refactor out options from the xlator package

kshlm:
A new `options` package has been created, by refactoring out option related
types from the `xlator` package.  The new package contains utility functions to
help find and validate options.


kshlm:
@prashanthpai I went this way to avoid circular dependencies. I have been thinking about creating a seperate package for types. By keeping all types in a single package, we can avoid circular dependencies. It would allow me to populate Options from within its package.

I wanted to do create such a package for Volume and Brick earlier. But during the refactor, it didn't seem required as both work well enough now. But now, for Xlator and Option, I feel like it is a good idea.

What do you think? Should I go with a separate types package? I maybe could (ab)use the type aliases in Go1.9 to make it look like the types still belong to their original packages.

prashanthpai:
@kshlm My preference would be to avoid creating a types package at the root level. Can both `xlator` and `option` be separate package with `option` package just containing the type definitions with utility functions and methods ? IOW, the global `options.Option` can be part of `xlator` package as they are really xlator options.

PR #449, opened 2017-11-24
Include liveness information in the responses from peer list and get requests.

kmlmhnn:


pkg/api/peer_req_resp.go
```diff
@@ -7,6 +7,7 @@ type Peer struct {
 	ID        uuid.UUID `json:"id"`
 	Name      string    `json:"name"`
 	Addresses []string  `json:"addresses"`
+	IsAlive   bool      `json:"isalive"`
```
prashanthpai: Please change this field name to `Online` and json string to `online`

Issue #453, opened 2017-11-28
Replicate volume support in GD2


karthik-us:
Missing template graph for replicate volume. Creating pure replicate volume is currently not supported in GD2. This issue is to track the related work in that area.

atinmu:
## Checklist
- [ ] ReST APIs availibility in wiki
- [ ] CLI commands
- [ ] Sufficient logging
- [ ] Availibility of metrics data if applicable
- [ ] Xlator options table upto date?
- [ ] Test cases & test coverage of the functionality

PR #460, opened 2017-11-29
volgen: Adding replicate.graph template

karthik-us:
Adding template for replicate graph to allow creation of pure
replicate volumes.

Fixes: #453
Signed-off-by: karthik-us <ksubrahm@redhat.com>

### DIFF
--- a/glusterd2/volgen/templates/replicate.graph
+++ b/glusterd2/volgen/templates/replicate.graph
@@ -1,2 +1,3 @@
+cluster/distribute
 cluster/replicate
 protocol/client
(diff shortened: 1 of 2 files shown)


glusterd2/volgen/templates/replicate.graph
```diff
@@ -0,0 +1,2 @@
+cluster/replicate
```
kshlm: This should be the same as the distreplicate graph, it needs to include cluster/distribute before cluster/replicate as well. All gluster volumes need to include distribute.

karthik-us: Done

kshlm:
add to whitelist

Issue #211, opened 2017-01-06
Define gluster cluster ID


prashanthpai:
Currently, there is no separate cluster ID associated with a glusterd2 instance. For all practical purposes, the etcd cluster ID is considered to be glusterd2 cluster ID. However, not all glusterd instances shall be etcd servers in future. We need better definition/mapping of gluster cluster ID and etcd cluster ID. This should allow not so common deployments like - making one beefy server run multiple glusterd2 instances and these instances can belong to different clusters altogether.

prashanthpai:
@kshlm @aravindavk @atinmu @amarts 

I'm planning to take this up. I propose the following approach:

* When glusterd comes up for the first time, it shall generate a cluster ID and be a single node cluster. This is how it is today but there's no cluster ID yet.

* Let's assume we have two independent glusterd instances named `node1` and `node2`. At the beginning, each of these shall have their own unique cluster ID. If `node1` peer probes `node2`, then `node1`'s cluster ID shall prevail as the cluster ID for the newly formed two-node cluster. IOW, peer probe request shall be treated as an invite to join an existing cluster.

* Any node in a fully formed cluster (two or more nodes) can then peer probe any other single glusterd2 instance.

* A single node gluster instance will be disallowed from peer probing a fully formed cluster (two or more nodes) cluster. IOW, joining an existing formed cluster shall be invite only. 

Further....

* Cluster ID shall be part of all inter-node communications (like transaction RPCs). This will prevent cross-cluster communication mismatch.
* External utilities (like heketi and others) can depend on cluster ID to uniquely identify a cluster instead of generating and maintaining their own.
* Etcd namespace shall be segregated (prefixed) by cluster ID. Hence the same etcd store (if external) can be utilised by multiple glusterfs clusters.

No cluster operations based on cluster ID shall be made available by glusterd2 though. Peer attach and detach shall work as before. To the end user, the only noticeable change shall be an additional `ClusterID` field in the peer API responses.

What are your thoughts ?

atinmu:
>What are your thoughts ?

Looks good to me.

prashanthpai:
> To the end user, the only noticeable change shall be an additional ClusterID field in the peer API responses

On a second thought. The Cluster ID field shall be the same for all peers in the same store. Hence not adding it to the peer object itself. The cluster ID shall be sent as a response header for all requests.

prashanthpai:
Sample output:

```sh
[...]

PR #463, opened 2017-11-30
Add Cluster ID

prashanthpai:
When glusterd comes up for the first time, it shall generate a cluster ID and be a single node cluster. 

Let's assume we have two independent glusterd instances named node1 and node2. At the beginning, each of these shall have their own unique cluster ID. If node1 peer probes node2, then node1's cluster ID shall prevail as the cluster ID for the newly formed two-node cluster. IOW, peer probe request shall be treated as an invite to join an existing cluster.

Any node in a fully formed cluster (two or more nodes) can then peer probe any other single glusterd2 instance.

A single node gluster instance will be disallowed from peer probing a fully formed cluster (two or more nodes) cluster. IOW, joining an existing formed cluster shall be invite only.

The Cluster ID will be returned as a response header `X-Gluster-Cluster-Id` in all responses. This is similar to etcd API responses.

**Store changes**:
Store access is allowed over a wrapped etcd client handle which automatically adds the right gluster store prefix. Consumers of store now cannot pollute parent namespace.

Closes issue #211


glusterd2/commands/peers/peer-rpc-svc.go
```diff
@@ -26,20 +27,25 @@ func (p *PeerService) RegisterService(s *grpc.Server) {
...
-	logger := log.WithField("remotepeer", req.PeerID)
+	logger := log.WithFields(log.Fields{
+		"remotepeer":    req.PeerID,
+		"remotecluster": req.ClusterID})
 
 	logger.Info("handling new incoming join cluster request")
 
 	// Handling a Join request happens as follows,
 	// 	- TODO: Ensure no ongoing operations (transactions/other peer requests) are happening
-	// 	- TODO: Check is peer is part of another cluster
```
kshlm: This is actually a step in the the list of steps to check, and should remain for the whole comment to make sense. Just remove the 'TODO:' prefix.

prashanthpai:
Response headers output:

```sh
curl -i -X GET http://127.0.0.1:24007/version
HTTP/1.1 200 OK
Content-Type: application/json; charset=UTF-8
X-Gluster-Cluster-Id: 4db2ccd9-d64e-4f50-ae67-c1bded86ca2a
X-Gluster-Node-Id: 93d21e85-0080-48ad-9b61-19dcddaac970
X-Request-Id: e169852e-a2af-4085-94aa-16c394309757
Date: Thu, 30 Nov 2017 12:13:16 GMT
Content-Length: 63

{"glusterd-version":"v4.0dev-9.95.gitba6637b","api-version":1}
```

glusterd2/commands/peers/peer-rpc-svc.go
prashanthpai: Oops! `Check is` should have been `Check if`. I will sneak in the grammatical fix in one of the future PRs.

PR #474, opened 2017-12-07
Internal events framework

kshlm:
The events framework allows events to be generated and broadcast internally
within the GD2 cluster, with `events.Broadcast(Event)`.

`Event`s are either local or global. Local events are broadcast just locally
within a GD2. Global events are broadcast globally to other GD2s.

Interested parties can register handlers for all or specific events using
`events.Register(EventHandler, events ...string)`

### DIFF
--- a/glusterd2/daemon/events.go
+++ b/glusterd2/daemon/events.go
@@ -1,0 +1,33 @@
+package daemon
+
+import (
+	"strconv"
+
+	"github.com/gluster/glusterd2/glusterd2/events"
+)
+
+type daemonEvent string
+
+const (
+	daemonStarting       daemonEvent = "daemon.starting"
+	daemonStarted                    = "daemon.started"
+	daemonStartFailed                = "daemon.startfailed"
+	daemonStopping                   = "daemon.stopping"
+	daemonStopped                    = "daemon.stopped"
+	daemonStopFailed                 = "daemon.stopfailed"
+	daemonStartingAll                = "daemon.startingall"
+	daemonStartedAll                 = "daemon.startedall"
+	daemonStartAllFailed             = "daemon.startallfailed"
+)
+
+// newEvent returns an event of given type with daemon data filled
+func newEvent(d Daemon, e daemonEvent, pid int) *events.Event {
+	data := make(map[string]string)
+	data["name"] = d.Name()
+	data["id"] = d.ID()
+	data["binary"] = d.Path()
+	data["args"] = d.Args()
+	data["pid"] = strconv.Itoa(pid)
+
+	return events.New(string(e), data, false)
+}
--- a/glusterd2/daemon/events.go
+++ b/glusterd2/daemon/events.go
@@ -22,12 +22,13 @@ const (
 
 // newEvent returns an event of given type with daemon data filled
 func newEvent(d Daemon, e daemonEvent, pid int) *events.Event {
-	data := make(map[string]string)
-	data["name"] = d.Name()
-	data["id"] = d.ID()
-	data["binary"] = d.Path()
-	data["args"] = d.Args()
-	data["pid"] = strconv.Itoa(pid)
+	data := map[string]string{
+		"name":   d.Name(),
+		"id":     d.ID(),
+		"binary": d.Path(),
+		"args":   d.Args(),
+		"pid":    strconv.Itoa(pid),
+	}
 
 	return events.New(string(e), data, false)
 }
--- a/glusterd2/events/eventhandler.go
+++ b/glusterd2/events/eventhandler.go
@@ -70,8 +70,8 @@ func Register(h Handler, events ...string) HandlerID {
 	in := make(chan *Event)
 	id := addHandler(in)
 
+	handlers.wg.Add(1)
 	go func() {
-		handlers.wg.Add(1)
 		handleEvents(in, h, events...)
 		handlers.wg.Done()
 	}()
@@ -95,8 +95,8 @@ func handleEvents(in <-chan *Event, h Handler, events ...string) {
 
 	for e := range in {
 		if interested(e, events) {
+			wg.Add(1)
 			go func() {
-				wg.Add(1)
 				h.Handle(e)
 				wg.Done()
 			}()
--- a/glusterd2/events/eventhandler.go
+++ b/glusterd2/events/eventhandler.go
@@ -6,13 +6,24 @@ import (
 	"sync"
 )
 
-// Handler is a function that is registered to be called when an event happens
-// event is the name of the event that happened and data is any extra data that
-// was attached to the event as a json document
-type Handler func(*Event)
+// Handler defines the event handler interface.
+// It is registered with the events framework to be called when an event
+// happens.
+type Handler interface {
+	// Handle is the function that gets called when an event occurs
+	Handle(*Event)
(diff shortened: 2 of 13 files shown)


glusterd2/daemon/events.go
```diff
@@ -0,0 +1,33 @@
...
+	daemonStopped                    = "daemon.stopped"
+	daemonStopFailed                 = "daemon.stopfailed"
+	daemonStartingAll                = "daemon.startingall"
+	daemonStartedAll                 = "daemon.startedall"
+	daemonStartAllFailed             = "daemon.startallfailed"
+)
+
+// newEvent returns an event of given type with daemon data filled
+func newEvent(d Daemon, e daemonEvent, pid int) *events.Event {
+	data := make(map[string]string)
```
purpleidea: You can replace this sort of pattern with:

```
	data := map[string]string{
		"foo": "hey",
		"bar": "hey2",
	}
```

OR

just use a struct, which is probably more appropriate.

HTH

kshlm: Having a unique struct per event for additional data, would work out if the events were localized. It wouldn't work for global events, as the global recievers wouldn't know what struct was set to properly deserialize.

kshlm: I'll change to the suggested coding pattern though.

kshlm:
GlobalEvents don't apparently get recieved at remote ends. I'm trying to debug this at the moment. Don't merge yet.

kshlm:
> GlobalEvents don't apparently get recieved at remote ends. I'm trying to debug this at the moment. Don't merge yet.

This happens because the global listeners on the new node were either still trying to listen to the old store or were stopped, after the old store was stopped on peer probe. I need stop/start the events listener on peer probe.

Now I need to figure out why gd2 isn't building for me. Something about linking failing.

kshlm:
> Now I need to figure out why gd2 isn't building for me. Something about linking failing.
Fixed this. Problem with go packages I was trying to use.

But anyways, even after stopping/stopping the global listener on probe/detach, still couldn't get global events to appear.

Figured out we have a bigger problem causing this. We haven't namespaced correctly! We namespaced `clientv3.Client.KV` only, but left out `clientv3.Client.Lease` and `clientv3.Client.Watcher`.
Also when using elasticetcd, we namespaced the KV used by it. Affecting it's performance.

Need to clean this up.

prashanthpai:
> Figured out we have a bigger problem causing this. We haven't namespaced correctly! We namespaced clientv3.Client.KV only, but left out clientv3.Client.Lease and clientv3.Client.Watcher.

I intentionally left out namespacing elasticetcd access of store.

kshlm:
Fixed namespace problems with #486. This PR now depends on that being merged first.

glusterd2/events/eventhandler.go
```diff
@@ -0,0 +1,129 @@
...
+
+func handleEvents(in <-chan *Event, h Handler, events ...string) {
+	var wg sync.WaitGroup
+
+	events = normalizeEvents(events)
+
+	for e := range in {
+		if interested(e, events) {
+			go func() {
+				wg.Add(1)
```
prashanthpai: AFAIK, `wg.Add()` must be called before spawning the go-routine.

```go
wg.Add(1)
go func() {
    defer wg.Done()
    h(e)
}
```

Same comment elsewhere too.

glusterd2/events/eventhandler.go
```diff
@@ -0,0 +1,129 @@
...
+// If no events are specified, handler function is called for all events
+// Handlers need to be thread safe, as they can be called concurrently when
+// multiple events arrive at the same time.
+func Register(h Handler, events ...string) HandlerID {
+	in := make(chan *Event)
+	id := addHandler(in)
+
+	go func() {
+		handlers.wg.Add(1)
+		handleEvents(in, h, events...)
```
prashanthpai: A general comment here on when go routines are spawned will help. One goroutine is spawned per handler and inside that goroutine, a new gorotuine is spawned per event.

glusterd2/events/eventhandler.go
kshlm: 👍

glusterd2/events/events.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"strings"
+
+	"github.com/pborman/uuid"
+)
+
+// Event represents an event in GD2
+type Event struct {
+	// ID is a unique event ID
```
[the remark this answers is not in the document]
kshlm: I agree that this doesn't need to be a part of `Event` itself. But let's take this up a little later, as it isn't essential right now.

glusterd2/events/global.go
```diff
@@ -0,0 +1,108 @@
...
+			if resp.Canceled {
+				return
+			}
+			for _, sev := range resp.Events {
+				var ev Event
+				if err := json.Unmarshal(sev.Kv.Value, &ev); err != nil {
+					log.WithField("event.id", string(sev.Kv.Key)).WithError(err).Error("could not unmarshal global event")
+					continue
+				}
+				if !uuid.Equal(ev.nodeid, gdctx.MyUUID) {
```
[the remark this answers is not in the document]
kshlm: I don't know why I thought this worked. Thanks! Will fix it.

glusterd2/events/logger.go
```diff
@@ -0,0 +1,26 @@
+package events
+
+import (
+	log "github.com/sirupsen/logrus"
+)
+
+var lhID HandlerID
+
+// eventLogger logs all events being generated in the GD2 cluster
```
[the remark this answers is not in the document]
kshlm: :+1: Sounds good. It'll be done.

glusterd2/events/eventhandler.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	delete(handlers.chans, id)
+	return c
+}
+
+// Register a Handler to be called when the given events happen.
+// If no events are specified, handler function is called for all events
+// Handlers need to be thread safe, as they can be called concurrently when
+// multiple events arrive at the same time.
+func Register(h Handler, events ...string) HandlerID {
```
prashanthpai: Now that there's `handler.Events()`, you do not need the events list I suppose. Here and in `handleEvents`

glusterd2/events/events.go
```diff
@@ -0,0 +1,64 @@
...
+	ID uuid.UUID
+	// Name is the the name of the event
+	Name string
+	// Data is any additional data attached to the event.
+	Data map[string]string
+	// global should be set to true to broadcast event to the full GD2 cluster.
```
prashanthpai: We may need a timestamp too. And this timestamp can be part of the etcd key. We can then do a ranged gets from timestamp1 to timestamp2.

glusterd2/events/logger.go
```diff
@@ -0,0 +1,26 @@
...
+// The events are logged in DEBUG level only.
+func eventLogger(e *Event) {
```
prashanthpai: This is polluting the default log with events. May be log to separate file ?

prashanthpai:
LGTM 👍 

It'll be good if @aravindavk also have a glance at this before merging.

aravindavk:
LGTM. Timestamp is required with event data. That can be added later.

PR #480, opened 2017-12-08
add arbiter support

itisravi:
This patch adds partial support for arbiter volume creation and
displaying volume info in the output. brick volgen support is still
pending, wherein we need to insert the arbiter xlator into every 3rd
brick of the replica. For now, even if you create arbiter volume, the
3rd brick becomes a full fledged data brick.

Signed-off-by: Ravishankar N <ravishankar@redhat.com>

### DIFF
--- a/glustercli/cmd/volume.go
+++ b/glustercli/cmd/volume.go
@@ -33,6 +33,7 @@ var (
 	// Create Command Flags
 	flagCreateCmdStripeCount       int
 	flagCreateCmdReplicaCount      int
+	flagCreateCmdArbiterCount      int
 	flagCreateCmdDisperseCount     int
 	flagCreateCmdDisperseDataCount int
 	flagCreateCmdRedundancyCount   int
@@ -50,6 +51,7 @@ func init() {
 	// Volume Create
 	volumeCreateCmd.Flags().IntVarP(&flagCreateCmdStripeCount, "stripe", "", 0, "Stripe Count")
 	volumeCreateCmd.Flags().IntVarP(&flagCreateCmdReplicaCount, "replica", "", 0, "Replica Count")
+	volumeCreateCmd.Flags().IntVarP(&flagCreateCmdArbiterCount, "arbiter", "", 0, "Arbiter Count")
 	volumeCreateCmd.Flags().IntVarP(&flagCreateCmdDisperseCount, "disperse", "", 0, "Disperse Count")
 	volumeCreateCmd.Flags().IntVarP(&flagCreateCmdDisperseDataCount, "disperse-data", "", 0, "Disperse Data Count")
 	volumeCreateCmd.Flags().IntVarP(&flagCreateCmdRedundancyCount, "redundancy", "", 0, "Redundancy Count")
@@ -146,6 +148,7 @@ var volumeCreateCmd = &cobra.Command{
 			Name:    volname,
 			Bricks:  bricks, // string of format <UUID>:<path>
 			Replica: flagCreateCmdReplicaCount,
+			Arbiter: flagCreateCmdArbiterCount,
 			Force:   flagCreateCmdForce,
 		})
 		if err != nil {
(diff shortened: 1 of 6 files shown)


prashanthpai:
add to whitelist

glusterd2/commands/volumes/volume-create.go
```diff
@@ -56,6 +56,14 @@ func createVolinfo(req *api.VolCreateReq) (*volume.Volinfo, error) {
 		v.ReplicaCount = req.Replica
 	}
 
+        if req.Arbiter !=0 {
+                if req.Replica != 3 || req.Arbiter != 1 {
+                        return nil, errors.New("For arbiter configuration, replica count must be 3 and arbiter count must be 1. The 3rd brick of the replica will be the arbiter");
+                }
+                fmt.Println("count=",req.Arbiter)
```
kshlm: I guess this is a debug line. Can be removed.

itisravi:
@kshlm @prashanthpai  Requesting reviews for this PR.

glusterd2/volume/struct.go
```diff
@@ -64,6 +64,7 @@ type Volinfo struct {
 	Transport    string
 	DistCount    int
 	ReplicaCount int
+	ArbiterCount int
```
prashanthpai: This can be a bool - here and elsewhere below and be named `Arbiter`

glustercli/cmd/volume.go
```diff
@@ -50,6 +51,7 @@ func init() {
 	// Volume Create
 	volumeCreateCmd.Flags().IntVarP(&flagCreateCmdStripeCount, "stripe", "", 0, "Stripe Count")
 	volumeCreateCmd.Flags().IntVarP(&flagCreateCmdReplicaCount, "replica", "", 0, "Replica Count")
+	volumeCreateCmd.Flags().IntVarP(&flagCreateCmdArbiterCount, "arbiter", "", 0, "Arbiter Count")
```
prashanthpai: Easier to understand if it's a bool.

itisravi: I think it would be better to retain this as an integer to maintain consistency with how it is used elsewhere in the gluster code (eg. in afr.c). Also, in volumeInfoDisplayNumbricks() we do display it as  a 'number' (i.e. the no. of arbiter bricks in a nx (2+1) volume).

glusterd2/volume/struct.go
itisravi: Same reason as above. I agree that ArbiterCount will always be one (or zero for normal replica volumes)  but it might be better to retain this as an int IMHO.

glusterd2/commands/volumes/volume-create.go
```diff
@@ -57,6 +57,13 @@ func createVolinfo(req *api.VolCreateReq) (*volume.Volinfo, error) {
 		v.ReplicaCount = req.Replica
 	}
 
+	if req.Arbiter != 0 {
```
atinmu: Does it make sense to have all the feature specific validations captured in a function? I'd not like to see the core GD2 functions getting bigger with different validations being added into the code directly. Instead a simple function call would make things more readable IMO.

amarts:
I like the idea of making it a volume type. But the challenge is the order of giving bricks. 

Also, it's high time we fix the naming. We should keep replica N as the N replicas, currently we say `replica 3 arbiter 1`, where data is present only 2 times.

aravindavk:
> Internal representation of Arbiter in code can be an int. However, if it's an int in the user facing public API or CLI, it can be non-intuitive and possibly cause confusion.

I think @itisravi's point is, today arbiter count is always 1. But in future more than one brick may be used as arbiter brick(if required, replica 5 arbiter 2)

I like @amarts idea of specifying `replica 2 arbiter 1 ` (also avoids above confusion)

itisravi:
We did debate using ` replica 2 arbiter 1` vs `replica 3 arbiter 1` while deciding on the CLI originally. We settled on `replica 3 arbiter 1` to keep in line with the (then) already existing EC volume semantics where we specify  redundancy count as a _subset_ of disperse count. For example `disperse 3 redundancy 1` would create a `1 x (2 + 1) = 3` volume.

In all the documentation, admin guide etc. also, we have emphasised that  arbiter is also a type of replica 3 volume where the 3rd replica stores only metadata. So not sure if changing it can cause confusion.

As for the volume type, I think it should still be replicate because it is fundamentally multiple copies (replicas) of the same file (as opposed to EC where data is encoded). Arbiter, thin-arbiter etc. are only strategies to give various degrees of `consistency vs availability`  choice to the user for the replication.

atinmu:
>We did debate using replica 2 arbiter 1 vs replica 3 arbiter 1 while deciding on the CLI originally. We settled on replica 3 arbiter 1 to keep in line with the (then) already existing EC volume semantics where we specify redundancy count as a subset of disperse count. For example disperse 3 redundancy 1 would create a 1 x (2 + 1) = 3 volume. As for the volume type, I think it should still be replicate because it is fundamentally multiple copies (replicas) of the same file (as opposed to EC where data is encoded). Arbiter, thin-arbiter etc. are only strategies to give various degrees of consistency vs availability choice to the user for the replication.

I don't think this comparison is correct. To me replica term should be only used for data bricks and not the ones where meta data or some other form of xattrs are stored. For EC volumes redundancy count 1 doesn't indicate that the node will not store any data (AFAIK). These EC semantics are more w.r.t the policies on how to recover the data back. With Gluster 4.0 this is a good opportunity to correct these semantics as otherwise it does cause confusion to users where some of the users think an arbiter volume is actually a 3 way replica along with 1 more arbiter node.

itisravi:
> With Gluster 4.0 this is a good opportunity to correct these semantics as otherwise it does cause confusion to users where some of the users think an arbiter volume is actually a 3 way replica along with 1 more arbiter node.
Okay, I can send an email on gluster-users to see if anyone has any objection to change the syntax to `replica 2 arbiter 1`. @karthik-us  @pranithk FYI/ comments.

itisravi:
Note: https://github.com/gluster/glusterd2/issues/515 has been opened for changing the syntax to `replica 2 arbiter 1`

PR #481, opened 2017-12-08
Volume Status: Handle nodes being down

### LINKED ISSUE #475: Volume status request should succeed even if few nodes/bricks are down

prashanthpai:
Do not fail the transaction if few of the nodes are down.

Changes to transaction framework:
* Check if nodes are up is now configurable
* Transaction framework users need not explicitly set the list of nodes to be checked for liveness
* Add option to not rollback and ignore failures.

Closes #475

### DIFF
--- a/glusterd2/transaction/transaction.go
+++ b/glusterd2/transaction/transaction.go
@@ -75,10 +75,10 @@ func (t *Txn) checkAlive() error {
 }
 
 // Do runs the transaction on the cluster
-func (t *Txn) Do() (TxnCtx, error) {
+func (t *Txn) Do() error {
 	if t.CheckAlive {
 		if err := t.checkAlive(); err != nil {
-			return nil, err
+			return err
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


glusterd2/transaction/transaction.go
```diff
@@ -26,9 +26,11 @@ type Txn struct {
 	reqID uuid.UUID
 	Ctx   TxnCtx
 	Steps []*Step
-	// Nodes should be a union of the all the TxnStep.Nodes and must be set
-	// before calling Txn.Do(). This is currently only used to determine
-	// liveness of the nodes before running the transaction steps.
+
+	CheckAlive bool
```
kshlm: Instead of `CheckAlive` it would be better to have a `DontCheckAlive`. This way the nil value is always false, and you would only need to set the value when you don't want the check to happen.

It is always nicer to keep the defaults to something most operations would use, and only have those operations that need it different, change it to what they want.

Issue #483, opened 2017-12-11
REST APIs end points for Gluster Metrics


aravindavk:
Today we have many alternatives/experiments to collect the metrics
from Gluster Cluster. But it is very difficult to consume these Metrics
since each to be collected through its own mechanism.

Metrics Glusterd2 plugin provides Uniform APIs for each type of
metrics(More metrics can be added when required). All the APIs will be
available in two forms, Cluster API and Local API.

While collecting Cluster data, Metrics plugin will make sure to reach
only those nodes which are required for collecting the data.

The Local Metrics API is limited to collecting data from the node where
Glusterd is running. For example, if a REST call is made to node 1's
glusterd then it returns local metrics from that node. Glusterd2 is
aware of associated metadata for each local metric, it adds the
required metadata to the API response.

## Metrics as API endpoints
Each metric will have two APIs, one to get Cluster's data and other
one to get Local data. Local Metrics APIs are mostly consumed by
`node_exporter`(Prometheus) or `node_agent`(Collectd based).

All of the metrics APIs provided by Glusterd2 are independent of
Monitoring tools. Any monitoring tools can be used to consume these
APIs. In case of Prometheus, the exporter can call Glusterd2's local
metrics APIs and export it. The Local REST call is similar to RPC and
it is not expensive. If Metrics are collected via Collectd, then
Collectd Gluster plugin can call local/cluster APIs and push metrics
to Graphite of central system.

### Local Metrics
Many monitoring application prefers collecting local data instead of
collecting all nodes data from one node.

Glusterd2 adds the relationship data to the collected local metrics in
API response. For example, Brick utilization. Instead of returning
Just brick path and Utilization data, API includes relationship data
as well. (POC is already available for Glusterd1
https://github.com/gluster/glustercli-python/tree/master/gluster/metrics)

### Cluster Metrics(Etcd data)
Many Cluster data is stored in etcd itself. Geting these data is not
expensive even though it is returning Cluster wide data. For example
Volume info, Number of Bricks per node etc

### Cluster Metrics(Non Etcd data)
Real time metrics of Cluster. To get these metrics Glusterd2 has to
reach out many nodes and aggregate the results. These APIs are not
suitable for continuous monitoring(less frequent poll)

All these metrics will have uniform APIs as follows. If Monitoring
application is capable of aggregating the metrics collected from
Glusterd2 then it can use `local` APIs, else it can consume `cluster`
APIs. Local APIs will not flood Network since it will not reach other
[...]

PR #484, opened 2017-12-11
Add size information to volume status and brick status

### LINKED ISSUE #329: gluster volume status should display size of the volume
It'd be good to display the total size and current utilization of the volume in volume status output.

prashanthpai:
Closes #329, #485


prashanthpai:
Sample output:

```json
$ curl -s -X GET http://127.0.0.1:24007/v1/volumes/test/status | python -m json.tool
{
    "bricks": [
        {
            "info": {
                "host": "127.0.0.1",
                "id": "ef1f8526-13a4-4058-a0bf-d37fde77d18f",
                "node-id": "ac2ae80d-a26e-43e9-911e-ff6df1d04b50",
                "path": "/export/brick1/data",
                "volume-id": "e3542345-1d9a-45b6-b9dc-620dc9c815e5",
                "volume-name": "test"
            },
            "online": true,
            "pid": 4556,
            "port": 49153
        }
    ],
    "size": {
        "capacity": 1047379968,
        "free": 902447104,
        "used": 144932864
    }
}
```

prashanthpai:
> One suggestion to keep all the sizes reported from gd2 to be always be in single unit (say Kb or something). Our experience from gd1 has been like sometimes the output is in MBs and sometimes just no of bytes.

We're sticking to bytes only for now. The CLI or tendrl can then interpret or display it however it likes :)

glusterd2/commands/volumes/disk_usage.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, err
+	}
+	defer syscall.Unmount(tempDir, syscall.MNT_FORCE)
+
+	var fstat syscall.Statfs_t
+	if err := syscall.Statfs(tempDir, &fstat); err != nil {
+		return nil, err
+	}
+
+	if fstat.Type != fuseSuperMagic {
```
atinmu: The function itself does a fuse mount.  Any specific reason why do we need to double confirm it?

glusterd2/commands/volumes/volume-status.go
```diff
@@ -111,31 +114,45 @@ func volumeStatusHandler(w http.ResponseWriter, r *http.Request) {
...
 		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err.Error(), api.ErrCodeDefault)
 		return
 	}
 
-	result, err := createVolumeStatusResp(txn.Ctx, vol)
+	var size *api.SizeInfo
+	select {
+	case size = <-ch:
+		// should be ready by the time the txn is over
+	case <-time.After(time.Second * 2):
```
atinmu: I am not very sure about what exactly are we trying to achieve here. If we're defining a particular time out here, then with more number of bricks in a volume would a mount be able to succeed and provide back the statfs details?

glusterd2/commands/volumes/disk_usage.go
prashanthpai: This is mostly redundant. This check reduces the likelihood of us returning statfs info of local filesystem instead of gluster volume's statfs. 

The `cmd.Wait()` on line 49 returns when the child (first `fork()`) exits. That's the first check to see if spawning child worked fine. If the fuse mount exits after the second `fork()` on failure scenarios, then the mountpoint dir points to local filesystem. We need to return correct data or no data.

glusterd2/commands/volumes/volume-status.go
prashanthpai: If glusterd2 isn't able to mount and get size information within certain period of time, we quit waiting. This is intended to handle scenarios where mount can be stuck or is taking too long to respond.

atinmu:
> This sort of information, which requires a costly operation shouldn't be available as the default response. This should be behind a 'detail' flag in the request. This would match volume status in GD1, where brick size information is only provided for volume status detail

+1, I agree.

glusterd2/commands/volumes/disk_usage.go
atinmu: Ack.

glusterd2/commands/volumes/volume-status.go
atinmu: There's still an open question. With a higher number of bricks, the mount might genuinely take some additional time. We may need to actually decide wisely on the timeout value we choose here (IMHO).

prashanthpai:
> This sort of information, which requires a costly operation shouldn't be available as the default response. This should be behind a 'detail' flag in the request. This would match volume status in GD1, where brick size information is only provided for volume status detail

Fair enough. I need some more clarity before implementing this.
* The `detail` flag - will that be a HTTP query param ?
* Or should this be a separate API ?
* If it's the same API, should the API response vary based on presence or absence of this flag ?

FWIW, volume status txn shall communicate with all nodes which make up the volume to see if the bricks are up. So does the mount process (a few more back and forth RPCs and process spawn overhead). However, both the transaction and the mount happens concurrently and not sequentially.

sankarshanmukhopadhyay:
I'd probably lean towards a separate API (in context of https://github.com/gluster/glusterd2/pull/484#issuecomment-351021402) thus completely separately serving the consumers who require this bit of detail. While considering testing the "scale" of this operations, it would be required to note both API response times as well as transaction hops for the aggregation of the data across nodes.

atinmu:
>I'd probably lean towards a separate API (in context of #484 (comment)) thus completely separately serving the consumers who require this bit of detail. While considering testing the "scale" of this operations, it would be required to note both API response times as well as transaction hops for the aggregation of the data across nodes.

With GD1, we already have `gluster volume status <volname> detail` command which provides some additional details about the brick, device, filesystem than what `gluster volume status <volname>` offers. If the current set of additional information is something which can be consumed separately (or they may not be much relevant from volume status perspective) having a different API is not a bad choice IMHO.

aravindavk:
> Or should this be a separate API ?

Separate API may be required for providing Uniform metrics APIs https://github.com/gluster/glusterd2/issues/483

> FWIW, volume status txn shall communicate with all nodes which make up the volume to see if the bricks are up. So does the mount process but with a few more back and forth RPCs and process spawn overhead. However, both the transaction and the mount happens concurrently and not sequentially.

+1 We can also improve this by mounting the volume in multiple bricks(For each volume mount in random(Volume bricks))

prashanthpai:
I've thought a little more about this.

Users do not consume bricks directly. Volume is the smallest unit of consumable storage. Returning brick statuses (pid, port, online) doesn't make much sense for an API or operation named as "volume status".

I'm leaning towards naming `/volumes/<volname>/bricks` as the API which shall return brick statuses (pid, port, online) along with regular static brick info.

 The `/volume/<volname>/status` can return more information like size etc pertaining to volume as a single entity.

shtripat:
Aren't we still moving to wards API like `/volumes/{vol-id}/bricks` ?

On volume status front I do agree with @prashanthpai but I understand that volume to brick relation would for sure be maintained under volume listing API in that case.

aravindavk:
> I've thought a little more about this.

> Users do not consume bricks directly. Volume is the smallest unit of consumable storage. Returning brick statuses (pid, port, online) doesn't make much sense for an API or operation named as "volume status".

> I'm leaning towards naming /volumes/<volname>/bricks as the API which shall return brick statuses (pid, port, online) along with regular static brick info.

We still need summary of bricks status in Volume API. (Volume Online/Offline instead of brick Online/Offline)

If we are introducing brick APIs, volname to be used as optional at the end. (/volumes/bricks/<volname>)

> The /volume/<volname>/status can return more information like size etc pertaining to volume as a single entity.

prashanthpai:
> We still need summary of bricks status in Volume API. (Volume Online/Offline instead of brick Online/Offline)

This can be provided. Volume online/offline can be known when we do a mount to get size info.

> If we are introducing brick APIs, volname to be used as optional at the end. (/volumes/bricks/)

We're not introducing a separate brick API yet.

prashanthpai:
**Current state of things:**

```json
$ curl -s -X GET http://127.0.0.1:24007/v1/volumes/test/bricks | python -m json.tool
[
    {
        "info": {
            "host": "127.0.0.1",
            "id": "2a2e4600-d945-4ddf-829a-485b8061fbf3",
            "node-id": "84f234ad-637c-4b55-8517-3e704c064a2c",
            "path": "/export/brick1/data",
            "volume-id": "7047efc0-9348-4ea2-8f78-8266d6e4fea8",
            "volume-name": "test"
        },
        "online": true,
        "pid": 18238,
        "port": 49154
    },
    {
        "info": {
            "host": "127.0.0.1",
            "id": "33a47760-f5fa-4086-b016-dd9f39a5d89e",
            "node-id": "f2fbf205-a598-42a4-8999-5e6e43498dc3",
            "path": "/export/brick2/data",
            "volume-id": "7047efc0-9348-4ea2-8f78-8266d6e4fea8",
            "volume-name": "test"
        },
        "online": true,
        "pid": 18287,
        "port": 49158
    },
    {
        "info": {
            "host": "127.0.0.1",
            "id": "a90f13db-ba77-4cb2-a082-7e2706294381",
            "node-id": "84f234ad-637c-4b55-8517-3e704c064a2c",
            "path": "/export/brick3/data",
            "volume-id": "7047efc0-9348-4ea2-8f78-8266d6e4fea8",
            "volume-name": "test"
        },
        "online": true,
        "pid": 18266,
        "port": 49159
    },
    {
        "info": {
            "host": "127.0.0.1",
            "id": "57eb3920-78fe-4b5c-a062-e993e83bc1a5",
            "node-id": "f2fbf205-a598-42a4-8999-5e6e43498dc3",
            "path": "/export/brick4/data",
            "volume-id": "7047efc0-9348-4ea2-8f78-8266d6e4fea8",
            "volume-name": "test"
        },
        "online": true,
        "pid": 18308,
        "port": 49164
    }
]
```

```json
$ curl -s -X GET http://127.0.0.1:24007/v1/volumes/test/status | python -m json.tool
{
    "info": {
        "arbiter-count": 0,
        "bricks": [
            {
                "host": "127.0.0.1",
                "id": "2a2e4600-d945-4ddf-829a-485b8061fbf3",
                "node-id": "84f234ad-637c-4b55-8517-3e704c064a2c",
                "path": "/export/brick1/data",
                "volume-id": "7047efc0-9348-4ea2-8f78-8266d6e4fea8",
                "volume-name": "test"
            },
            {
                "host": "127.0.0.1",
                "id": "33a47760-f5fa-4086-b016-dd9f39a5d89e",
                "node-id": "f2fbf205-a598-42a4-8999-5e6e43498dc3",
                "path": "/export/brick2/data",
                "volume-id": "7047efc0-9348-4ea2-8f78-8266d6e4fea8",
                "volume-name": "test"
            },
            {
                "host": "127.0.0.1",
                "id": "a90f13db-ba77-4cb2-a082-7e2706294381",
                "node-id": "84f234ad-637c-4b55-8517-3e704c064a2c",
                "path": "/export/brick3/data",
                "volume-id": "7047efc0-9348-4ea2-8f78-8266d6e4fea8",
                "volume-name": "test"
            },
            {
                "host": "127.0.0.1",
                "id": "57eb3920-78fe-4b5c-a062-e993e83bc1a5",
                "node-id": "f2fbf205-a598-42a4-8999-5e6e43498dc3",
                "path": "/export/brick4/data",
                "volume-id": "7047efc0-9348-4ea2-8f78-8266d6e4fea8",
                "volume-name": "test"
            }
        ],
        "distribute-count": 2,
        "id": "7047efc0-9348-4ea2-8f78-8266d6e4fea8",
        "name": "test",
        "options": {},
        "replica-count": 2,
        "state": "Started",
        "transport": "tcp",
        "type": "Distributed-Replicate"
    },
    "size": {
        "capacity": 1992949760,
        "free": 1902772224,
        "used": 90177536
    }
}
```

atinmu:
@prashanthpai After all these patches how would `bricks` and `status` API response will look like? Would we not have a different flag in volume status API to fetch the size of the volume? Is volume status API going to provide all the static informations? Based on the output I can ask further questions..

prashanthpai:
> After all these patches how would bricks and status API response will look like? 

There is a sample output in the above comment. I plan to add more fields (mount opts, device) to brick status output.

> Would we not have a different flag in volume status API to fetch the size of the volume?

There are two separate APIs now. One is `/volumes/<volname>/bricks` and `/volumes/<volname>/status`

> Is volume status API going to provide all the static informations?

Volume status will provide dynamic information (currently only size) w.r.t volume as a single entity.

In general, both these APIs return static info in addition to dynamic information.

atinmu:
> I'm leaning towards naming /volumes//bricks as the API which shall return brick statuses (pid, port, online) along with regular static brick info.

> >We still need summary of bricks status in Volume API. (Volume Online/Offline instead of brick Online/Offline)

+1

prashanthpai:
**This is output of the two APIs:**

```json
$ curl -s -X GET http://127.0.0.1:24007/v1/volumes/test/bricks | python -m json.tool
[
    {
        "device": "/dev/loop3",
        "fs-type": "xfs",
        "info": {
            "host": "127.0.0.1",
            "id": "1a6fa8c0-8542-4429-a331-cd475fbaac97",
            "node-id": "2922ffb9-55c8-4c1f-b5dd-def4a84fac81",
            "path": "/export/brick1/data",
            "volume-id": "88c3edbd-d6f8-4088-9e71-7ce1a53a5ca5",
            "volume-name": "test"
        },
        "mount-opts": "rw,seclabel,noatime,nodiratime,attr2,inode64,noquota",
        "online": true,
        "pid": 6684,
        "port": 49154,
        "size": {
            "capacity": 996495360,
            "free": 961363968,
            "used": 35131392
        }
    }
]
```

```json
$ curl -s -X GET http://127.0.0.1:24007/v1/volumes/test/status | python -m json.tool
{
    "info": {
        "arbiter-count": 0,
        "bricks": [
            {
                "host": "127.0.0.1",
                "id": "1a6fa8c0-8542-4429-a331-cd475fbaac97",
                "node-id": "2922ffb9-55c8-4c1f-b5dd-def4a84fac81",
                "path": "/export/brick1/data",
                "volume-id": "88c3edbd-d6f8-4088-9e71-7ce1a53a5ca5",
                "volume-name": "test"
            }
        ],
        "distribute-count": 1,
        "id": "88c3edbd-d6f8-4088-9e71-7ce1a53a5ca5",
        "name": "test",
        "options": {},
        "replica-count": 1,
        "state": "Started",
        "transport": "tcp",
        "type": "Distribute"
    },
    "size": {
        "capacity": 996409344,
        "free": 951320576,
        "used": 45088768
    }
}
```

glusterd2/commands/volumes/fs_utils.go
```diff
@@ -0,0 +1,137 @@
...
+
+	f, err := os.Open("/etc/mtab")
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	var l []*mntent
+
+	scanner := bufio.NewScanner(f)
```
kshlm: It would be better to read the whole file in once, and then parse it line by line. This isn't a particularly large file that we need to stream scan it. And avoids possible issues caued by the file changing in between.

pkg/restclient/volume.go
```diff
@@ -29,9 +29,9 @@ func (c *Client) Volumes(volname string) (api.VolumeListResp, error) {
 }
 
 // VolumeStatus returns the status of a Gluster volume
-func (c *Client) VolumeStatus(volname string) (api.VolumeStatusResp, error) {
-	url := fmt.Sprintf("/v1/volumes/%s/status", volname)
-	var volStatus api.VolumeStatusResp
+func (c *Client) VolumeStatus(volname string) (api.BricksStatusResp, error) {
```
kshlm: Volume status endpoint still exists, and this needs to still match it. Add a new function for brick status.

glusterd2/commands/volumes/fs_utils.go
```diff
@@ -0,0 +1,137 @@
...
+		fsName:  f[0],
+		mntDir:  f[1],
+		mntType: f[2],
+		mntOpts: f[3],
+	}
+}
+
+func readMtabFile() ([]*mntent, error) {
+
+	f, err := os.Open("/etc/mtab")
```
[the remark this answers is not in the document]
prashanthpai: I had to choose between the two files and I chose `/etc/mtab` because [these](https://linux.die.net/man/3/getmntent) glibc wrappers that most of the C programs use read `/etc/mtab`.

I'll change it to `/proc/mounts` 👍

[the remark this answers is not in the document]
aravindavk: glusterd1 is using `getmntent_r` check `glusterd_get_mnt_entry_info` in file `xlators/mgmt/glusterd/src/glusterd-utils.c`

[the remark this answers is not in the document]
prashanthpai: @aravindavk Yes. It's utility function provided by glibc. We can't use that in go directly. Using it via cgo is an overkill as reading `/etc/mtab` or `/proc/mounts` is trivial and easy.

Issue #428, opened 2017-11-15
rebalance commands in GD2


atinmu:
## APIs to be defined:

    rebalance start
    rebalance fix-layout start
    rebalance start [force]
    rebalance stop
    rebalance status

atinmu:
@aravindavk can you please help in defining the ReST APIs for the above?

nbalacha:
Information on the current rebalance behaviour:

**Rebalance [fix-layout] start:**
1. The rebalance process runs on every node of the volume
2. The rebalance process handles a single volume.
3. The rebalance process uses the <volname>-rebalance.vol  volfile which is the same as the Fuse client volfile but only contains xlators from dht and down (No perf xlators, etc). It also seems to have io-stats loaded on top of DHT.
4. Rebalance requires that all child subvols be up as the selfheal which sets the layout of the directories requires all subvols to be available.. The process will terminate if any subvolume goes down.


**Rebalance start force:**
Same as above but skips some space checks.  This is used when the rebalance is started by the remove-brick operation.


**Rebalance stop:**
Stops the ongoing rebalance operation. The process will finish migrating any files it might be in the middle of before terminating so it might take some time to actually stop.

nbalacha:
Rebalance command:

Started running /usr/local/sbin/glusterfs version 4.0dev1     (args: /usr/local/sbin/glusterfs -s localhost --volfile-id rebalance/vol1 --xlator-option *dht.use-readdirp=yes --xlator-option *dht.lookup-unhashed=yes --xlator-option *dht.assert-no-child-down=yes --xlator-option *replicate*.data-self-heal=off --xlator-option *replicate*.metadata-self-heal=off --xlator-option *replicate*.entry-self-heal=off --xlator-option *dht.readdir-optimize=on --process-name rebalance --xlator-option *dht.rebalance-cmd=1 --xlator-option *dht.node-uuid=96f35379-4a82-4207-812f-d91462348cc6 --xlator-option *dht.commit-hash=3502166428 --socket-file /var/run/gluster/gluster-rebalance-55c320ec-8693-4d04-ab26-052f0032ab96.sock --pid-file /var/lib/glusterd/vols/vol1/rebalance/96f35379-4a82-4207-812f-d91462348cc6.pid -l /var/log/glusterfs/vol1-rebalance.log)

atinmu:
## Checklist
- [ ] ReST APIs availibility in wiki
- [ ] CLI commands
- [ ] Sufficient logging
[...]

PR #487, opened 2017-12-12
glusterd2/rebalance : rebalance integration to GD2

Akarsha-rai:
I am still working on the patch.

Please share your inputs on the patch if you have any.

Signed-off-by: Akarsha <akrai@redhat.com>


prashanthpai:
add to whitelist

plugins/rebalance/init.go
```diff
@@ -0,0 +1,62 @@
...
+		route.Route{
+			Name:        "RebalanceStart",
+			Method:      "POST",
+			Pattern:     "/volumes/rebalance/{volname}/start",
+			Version:     1,
+			HandlerFunc: rebalanceStart},
+		route.Route{
+			Name:        "RebalanceStartForce",
+			Method:      "POST",
+			Pattern:     "/volumes/rebalance/{volname}/start/{force}",
```
atinmu: Would a force option need to have a different API? I guess no?

@kshlm @prashanthpai @aravindavk How are we handing the volume start force?

nbalacha: I think it makes sense to have force be an argument to the start API instead a separate API.

aravindavk: force should be included in the HTTP request(Not part of URL)

    {
        "force": true
    }

atinmu:
There are some validation related code pieces which are being commented out in the PR which need to be revived back.

plugins/rebalance/init.go
```diff
@@ -0,0 +1,62 @@
...
+	return nil
+}
+
+// RestRoutes returns list of REST API routes to register with Glusterd
+func (p *Plugin) RestRoutes() route.Routes {
+	return route.Routes{
+		route.Route{
+			Name:        "RebalanceStart",
+			Method:      "POST",
+			Pattern:     "/volumes/rebalance/{volname}/start",
```
kshlm: The path should be "/volumes/{volname}/rebalance...", as these are operations on a volume.

plugins/rebalance/init.go
```diff
@@ -0,0 +1,62 @@
...
+		route.Route{
+			Name:        "RebalanceStatus",
+			Method:      "GET",
+			Pattern:     "/volumes/rebalance/{volname}/status",
+			Version:     1,
+			HandlerFunc: rebalanceStatus},
+		route.Route{
+			Name:        "RebalanceFix-layout",
+			Method:      "POST",
+			Pattern:     "/volumes/rebalance/{volname}/fix-layout/start",
```
kshlm: Similar to the "force" option, "fix-layout" should be an option being passed to "rebalance/start". And you just have a single handler for start.

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,231 @@
...
+import (
+	//"context"
+	//"encoding/json"
+	"fmt"
+
+	"github.com/gluster/glusterd2/glusterd2/gdctx"
+	restutils "github.com/gluster/glusterd2/glusterd2/servers/rest/utils"
+	"github.com/gluster/glusterd2/glusterd2/store"
+	"github.com/gluster/glusterd2/glusterd2/transaction"
+	"github.com/gluster/glusterd2/glusterd2/volume"
```
kshlm: A comment on style. We group packages into blocks based on where they're imported from. The blocks are in the following order,
```go
import (
  // Standard library packages first
  "os"
  "fmt"

  // Packages from GD2 project in the middle

  // External packages last
)
```

Each block is also sorted alphabetically. To make this easier you should set up you editor with plugins that run 'goimports` when you save your file.

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,231 @@
...
+	"github.com/gluster/glusterd2/pkg/api"
+	"github.com/gorilla/mux"
+	"github.com/pborman/uuid"
+	log "github.com/sirupsen/logrus"
+	"net/http"
+)
+
+type Rebelstatus uint64
```
kshlm: s/Rebel/Rebal/

Its rebalance not rebelence. It would be nicer to just call this `RebalanceStatus` instead of shortening.

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,231 @@
...
+
+const (
+	STARTED Rebelstatus = iota
```
kshlm: Try not to use all caps identifiers.

shtripat: Do you really want it to be `Rebelstatus` or `RebalStatus` :)

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	err = txn.Ctx.Set("volname", volname)
+	if err != nil {
+		logger.WithError(err).Error("failed to set request in transaction context")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err.Error(), api.ErrCodeDefault)
+		return
+	}
+
+	reb, err := createRebinfo(req)
```
shtripat: As @kshlm pointed earlier shortening names like this is not corect. Better use `createRebalanceInfo`

atinmu:
@Akarsha-rai Just cross checking if you're stuck here. Would like to hear about the progress.

atinmu:
@Akarsha-rai Please refresh this PR as discussed.

pkg/api/rebalance_req.go
```diff
@@ -0,0 +1,27 @@
...
+type RebalanceStatus uint64
+
+const (
+	// Started should be set only for a volume that has been just started rebalance process
+	Started RebalanceStatus = iota
+	// Inprogress should be set only for a volume that are running rebalance process
+	Inprogress
+	// Failed should be set only for a volume that are failed to run rebalance process
+	Failed
+	// Completed sholud be set only for a volume that the rebalance process is completed
```
nbalacha: Typo: should.

pkg/api/rebalance_req.go
```diff
@@ -0,0 +1,27 @@
+package api
+
+import "github.com/pborman/uuid"
+
+// RebalanceStatus represents Rebalance Status
+type RebalanceStatus uint64
```
nbalacha: This will need more values. See GD1:

enum gf_defrag_status_t {
        GF_DEFRAG_STATUS_NOT_STARTED,
        GF_DEFRAG_STATUS_STARTED,
        GF_DEFRAG_STATUS_STOPPED,
        GF_DEFRAG_STATUS_COMPLETE,
        GF_DEFRAG_STATUS_FAILED,
        GF_DEFRAG_STATUS_LAYOUT_FIX_STARTED,
        GF_DEFRAG_STATUS_LAYOUT_FIX_STOPPED,
        GF_DEFRAG_STATUS_LAYOUT_FIX_COMPLETE,
        GF_DEFRAG_STATUS_LAYOUT_FIX_FAILED,
};

pkg/api/rebalance_req.go
```diff
@@ -0,0 +1,27 @@
...
+	Started RebalanceStatus = iota
+	// Inprogress should be set only for a volume that are running rebalance process
+	Inprogress
+	// Failed should be set only for a volume that are failed to run rebalance process
+	Failed
+	// Completed sholud be set only for a volume that the rebalance process is completed
+	Completed
+)
+
+// RebalanceInfo represents a Rebalance Info Details
```
nbalacha: Please provide more descriptive comments

plugins/rebalance/rebalance.go
```diff
@@ -0,0 +1,109 @@
...
+	buffer.WriteString(fmt.Sprintf(" --xlator-option *dht.use-readdirp=yes"))
+	buffer.WriteString(fmt.Sprintf(" --xlator-option *dht.lookup-unhashed=yes"))
+	buffer.WriteString(fmt.Sprintf(" --xlator-option *dht.assert-no-child-down=yes"))
+	buffer.WriteString(fmt.Sprintf(" --xlator-option replicate.data-self-heal=off"))
+	buffer.WriteString(fmt.Sprintf(" --xlator-option replicate.metadata-self-heal=off"))
+	buffer.WriteString(fmt.Sprintf(" --xlator-option replicate.entry-self-heal=off"))
+	buffer.WriteString(fmt.Sprintf(" --xlator-option *dht.readdir-optimize=on"))
+	buffer.WriteString(fmt.Sprintf(" --process-name rebalance"))
+	buffer.WriteString(fmt.Sprintf(" --xlator-option *dht.rebalance-cmd=%d", cmd))
+	buffer.WriteString(fmt.Sprintf(" --xlator-option *dht.node-uuid=%s", gdctx.MyUUID))
```
nbalacha: I'm not sure what this does but it is expected to be the node-uuid of the node on which the process will be running.

plugins/rebalance/rebalance.go
```diff
@@ -0,0 +1,109 @@
...
+
+	shost, _, _ := net.SplitHostPort(config.GetString("clientaddress"))
+	if shost == "" {
+		shost = "localhost"
+	}
+
+	volFileID := path.Join("rebalance/", r.rebalance.Volname)
+	File := path.Join(config.GetString("logdir"), "glusterfs", r.rebalance.Volname)
+	logFile := fmt.Sprintf("%s-rebalance.log", File)
+	cmd := r.rebalance.Status
```
nbalacha: This is requesting the rebalance status. Is this correct?

plugins/rebalance/rebalance.go
aravindavk: `gdctx.MyUUID` is current node's uuid where Glusterd2 is running

nbalacha: Will this be run on each node of the volume and updated correctly?

pkg/api/rebalance_req.go
nbalacha: Each of these are specific to the rebalance process on a node. So a volume with 3 nodes can have one rebalance process that has failed, one in progress and one that has completed.

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,279 @@
...
+		restutils.SendHTTPError(r.Context(), w, http.StatusInternalServerError, err.Error(), api.ErrCodeDefault)
+		return
+	}
+
+	txn.Nodes = vol.Nodes()
+	txn.Steps = []*transaction.Step{
+		lock,
+		{
+			DoFunc: "rebal-start.Commit",
+			Nodes:  []uuid.UUID{gdctx.MyUUID},
```
aravindavk: This will start rebalance process only in current node. Use `Nodes: txn.Nodes` to start rebalance process in all Volume nodes

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,279 @@
...
+	"github.com/gluster/glusterd2/glusterd2/transaction"
+	"github.com/gluster/glusterd2/glusterd2/volume"
+	"github.com/gluster/glusterd2/pkg/api"
+	"github.com/gorilla/mux"
+	"github.com/pborman/uuid"
+	log "github.com/sirupsen/logrus"
+)
+
+// RebalanceStatus represents Rebalance Status
+type RebalanceStatus uint64
```
nbalacha: Should this be duplicated across files?

plugins/rebalance/rebalance.go
aravindavk: This is part of Daemon interface called from Transaction step. This will be updated correctly when run on each node.

Added a comment in Transaction step to start in all Volume nodes instead of only current node

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,279 @@
...
+	Volname           string
+	Status            RebalanceStatus
+	RebalanceID       uuid.UUID
+	RebalanceFiles    uint64
+	RebalanceData     uint64
+	LookedupFiles     uint64
+	RebalanceFailures uint64
+	RebalanceTime     uint64
+	SkippedFiles      uint64
+	CommitHash        uint64
```
nbalacha: I don't think the CommitHash is saved in this. However, looks like some of the time fields (time_left) are missing.
Please rename RebalanceTime to ElapsedTime.

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func createRebalanceInfo(rebalance *RebalanceInfo) (*RebalanceInfo, error) {
+	v := new(RebalanceInfo)
+	v.Volname = rebalance.Volname
+	v.RebalanceID = uuid.NewRandom()
+	v.Status = rebalance.Status
```
nbalacha: What will this be set to?

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,279 @@
...
+		restutils.SendHTTPError(r.Context(), w, http.StatusNotFound, "Invalid volume", api.ErrCodeDefault)
+		return
+	}
+
+	if vol.State != volume.VolStarted {
+		restutils.SendHTTPError(r.Context(), w, http.StatusBadRequest, "Volume not started", api.ErrCodeDefault)
+		return
+	}
+
+	if vol.DistCount <= 1 {
```
nbalacha: I don't think DistCount can be 0.

plugins/rebalance/rest.go
aravindavk: Answering here since it is generic golang question. Scope of this definition is that directory. All files of a directory belongs to same package in Go, So this will be available for all files

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,279 @@
...
+
+	//Validate rebalance command
+	vol, err := volume.GetVolume(volname)
+	if err != nil {
+		restutils.SendHTTPError(r.Context(), w, http.StatusNotFound, "Invalid volume", api.ErrCodeDefault)
+		return
+	}
+
+	if vol.DistCount <= 1 {
+		restutils.SendHTTPError(r.Context(), w, http.StatusBadRequest, "Volume is not distribute type or contain only 1 brick",
```
nbalacha: Please reword to
Volume is not a distributed volume or contains only one brick

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,279 @@
...
+			api.ErrCodeDefault)
+		return
+	}
+
+	// Check rebalance is already in process
+	if rebalance.Status == Inprogress {
```
nbalacha: Where is this set?

plugins/rebalance/transaction.go
```diff
@@ -0,0 +1,66 @@
...
+			return nil
+		}
+		// Reset all variables
+		glusterdVolinfoResetStats(r)
+		err = daemon.Start(rebalanceProcess, true)
+
+	case "rebalanceStop":
+		// Set rebalance status as stopped
+		var r RebalanceInfo
+		r.Status = Stopped
```
nbalacha: Should this be set only if Daemon.Stop returns success?

plugins/rebalance/transaction.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return err
+	}
+
+	switch cmd {
+
+	case "rebalanceStart":
+		// Set rebalance status as started and store rebalance id
+		var r RebalanceInfo
+		r.Status = Started
```
nbalacha: Should this be set only if Daemon.Start returns success?

plugins/rebalance/utils.go
```diff
@@ -0,0 +1,33 @@
...
+func glusterdVolinfoResetStats(r RebalanceInfo) {
+	r.RebalanceFiles = 0
+	r.RebalanceData = 0
+	r.LookedupFiles = 0
+	r.RebalanceFailures = 0
+	r.RebalanceTime = 0
+	r.SkippedFiles = 0
+}
+
+func setCommitHash(r *RebalanceInfo) {
```
nbalacha: Please copy the comments from gd_set_commit_hash so we know why is being done this way.

plugins/rebalance/rebalance.go
Akarsha-rai: yes. r.rebalance.Status will return Started

plugins/rebalance/rest.go
Akarsha-rai: it will set v.Status as GfDefragStatusStarted

aravindavk:
Updates: #428

pkg/api/rebalance_req.go
Akarsha-rai: I removed duplicates and wrote this in rest.go file

plugins/rebalance/init.go
```diff
@@ -0,0 +1,52 @@
...
+	return nil
+}
+
+// RestRoutes returns list of REST API routes to register with Glusterd
+func (p *Plugin) RestRoutes() route.Routes {
+	return route.Routes{
+		route.Route{
+			Name:        "RebalanceStart",
+			Method:      "POST",
+			Pattern:     "/volumes/{volname}/rebalance/start",
```
nbalacha: How will fix-layout be handled here?

plugins/rebalance/rebalance.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/gluster/glusterd2/glusterd2/gdctx"
+	config "github.com/spf13/viper"
+)
+
+const (
+	sbinDir = "glusterfs"
+)
+
+// RebalanceProcess type represents information required to start rebalance process
+type RebalanceProcess struct {
```
nbalacha: Should this include the logfile path as well?

plugins/rebalance/rebalance.go
```diff
@@ -0,0 +1,108 @@
...
+// Path returns absolute path to the binary of rebalance process
+func (r *RebalanceProcess) Path() string {
+	return r.binarypath
+}
+
+// Args returns arguments to be passed to rebalance process
+func (r *RebalanceProcess) Args() string {
+
+	shost, _, _ := net.SplitHostPort(config.GetString("clientaddress"))
+	if shost == "" {
```
nbalacha: Should this var be called volfileserver (more descriptive) to match gd1?

plugins/rebalance/rebalance.go
```diff
@@ -0,0 +1,108 @@
...
+		shost = "localhost"
+	}
+
+	volFileID := path.Join("rebalance/", r.rebalanceinfo.Volname)
+	File := path.Join(config.GetString("logdir"), "glusterfs", r.rebalanceinfo.Volname)
```
nbalacha: Will config.GetString("logdir") always return only /var/log ?

plugins/rebalance/rebalance.go
```diff
@@ -0,0 +1,108 @@
...
+	logFile := fmt.Sprintf("%s-rebalance.log", File)
+	cmd := r.rebalanceinfo.Status
```
nbalacha: I don't think this looks correct. Status is not the same as cmd.

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,364 @@
...
+
+	err = txn.Ctx.Set("volname", volname)
+	if err != nil {
+		logger.WithError(err).Error("failed to set request in transaction context")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err.Error(), api.ErrCodeDefault)
+		return
+	}
+
+	rebalanceinfo.Volname = volname
+	rebalanceinfo.Status = GfDefragStatusStopped
```
nbalacha: Status and cmd are not the same and it looks like they may be used interchangeably here.

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,364 @@
...
+
+	return &api.RebalanceInfo{
+		Volname:     v.Volname,
+		Status:      api.RebalanceStatus(v.Status),
+		RebalanceID: v.RebalanceID,
+	}
+}
+
+func registerRebalanceStartStepFuncs() {
+	transaction.RegisterStepFunc(StartRebalance, "rebal-start.Commit")
```
aravindavk: You can add all these register calls inside `RegisterStepFuncs()` itself, instead of making as separate funcs

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,364 @@
...
+	transaction.RegisterStepFunc(storeRebalanceDetails, "rebal-start.StoreVolume")
+}
+
+func registerRebalanceStopStepFuncs() {
+	transaction.RegisterStepFunc(StopRebalance, "rebal-stop.Commit")
```
aravindavk: Same as above, refer `plugins/georeplication/init.go`

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,364 @@
...
+}
+
+func registerRebalanceStatusStepFuncs() {
+	transaction.RegisterStepFunc(StatusRebalance, "rebal-status.Commit")
```
aravindavk: Same as above, refer `plugins/georeplication/init.go`

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,364 @@
...
+	logger := gdctx.GetReqLogger(ctx)
+
+	//collect inputs from url
+	volname := mux.Vars(r)["volname"]
+
+	rebalanceinfo := new(RebalanceInfo)
+
+	//Check volname given
+	if volname == "" {
+		restutils.SendHTTPError(r.Context(), w, http.StatusBadRequest, "Volume not found", api.ErrCodeDefault)
```
aravindavk: Redundant check, anyways fail in the next step

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,364 @@
...
+		return
+	}
+
+	if vol.DistCount == 1 {
+		restutils.SendHTTPError(r.Context(), w, http.StatusBadRequest, "Volume is not distributed volume or contain only 1 brick",
+			api.ErrCodeDefault)
+		return
+	}
+
+	/*Check volume is a tier volume or not
```
aravindavk: Remove commented code

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,364 @@
...
+	if vol.Type == volume.Tier {
+	        restutils.SendHTTPError(r.Context(), w, http.StatusBadRequest, "Rebalance operations are not supported on a tier volume",
+	                                api.ErrCodeDefault)
+	        return
+	}*/
+
+	// Check rebalance is already in process
+	if rebalanceinfo.Status == GfDefragStatusStarted {
```
aravindavk: Wrong condition, rebalnceinfo is not fetched from store, it is initialized in Line 95

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,364 @@
...
+	//collect inputs from url
+	volname := mux.Vars(r)["volname"]
+
+	//Validate rebalance command
+	vol, err := volume.GetVolume(volname)
+	if err != nil {
+		restutils.SendHTTPError(r.Context(), w, http.StatusNotFound, "Invalid volume", api.ErrCodeDefault)
+		return
+	}
+
```
aravindavk: Need a condition here to check Rebalance process started or not. That is only possible if Rebalance info is stored in store.

plugins/rebalance/transaction.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	rebalanceProcess, err := NewRebalanceProcess(rinfo)
+	if err != nil {
+		return err
+	}
+
+	switch cmd {
+
+	case "rebalanceStart":
```
aravindavk: Avoid having string based conditions, typo may cause failures. Use constants

plugins/rebalance/transaction.go
```diff
@@ -0,0 +1,53 @@
...
+		var r RebalanceInfo
+		// Reset all variables
+		glusterdvolinforesetstats(r)
```
aravindavk: `r` is not used after this line?

plugins/rebalance/utils.go
```diff
@@ -0,0 +1,133 @@
...
+	   resolution.
+	*/
+
+	hash |= 1 << ((tv.TVusec >> 10) % 3)
+	r.CommitHash = hash
+	return
+}
+
+// GetRebalanceDetails provides the details of rebalance process which is stored
+func GetRebalanceDetails(rebalance *RebalanceInfo) (*RebalanceInfo, error) {
```
aravindavk: Why this function required? input = output

plugins/rebalance/utils.go
```diff
@@ -0,0 +1,133 @@
...
+	if err := AddOrUpdateFunc(&rinfo); err != nil {
+		c.Logger().WithError(err).WithField(
+			"volume", rinfo.Volname).Debug("failed to store rebalance details")
+		return err
+	}
+	return nil
+}
+
+// AddOrUpdateFunc used to add or update the rebalance details
+func AddOrUpdateFunc(r *RebalanceInfo) error {
```
aravindavk: Merge this function with previous function

plugins/rebalance/transaction.go
```diff
@@ -0,0 +1,53 @@
...
+		err = daemon.Start(rebalanceProcess, true)
+
+	case "rebalanceStop":
+		err = daemon.Stop(rebalanceProcess, true)
+
+	case "rebalanceStatus":
+		return nil
```
aravindavk: not implemented?

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,364 @@
...
+	err = txn.Ctx.Set("volname", volname)
+	if err != nil {
+		logger.WithError(err).Error("failed to set request in transaction context")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err.Error(), api.ErrCodeDefault)
+		return
+	}
+	// TODO: list all nodes where the process is running
+	// Currently gives only localhost
+	NodeID := gdctx.MyUUID
+	rebalinfo, err := GetRebalanceDetails(rebal)
```
aravindavk: Is it not getting status from running rebalance process?

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,364 @@
...
+	if err != nil {
+		logger.WithFields(log.Fields{
+			"error":   err.Error(),
+			"volname": volname,
+		}).Error("failed to stop rebalance on volume")
+		restutils.SendHTTPError(r.Context(), w, http.StatusInternalServerError, err.Error(), api.ErrCodeDefault)
+		return
+	}
+
+	e := AddOrUpdateFunc(rebal)
```
aravindavk: Always Update the store as a transaction step with txn.Nodes as current node.

aravindavk:
Also add a test case to create a volume, start volume, start rebalance, stop rebalance and delete volume

Akarsha-rai:
Closing this pull request and the new commit related to rebalance integration to GD2 is in #550

PR #489, opened 2017-12-18
Enable TLS authentication for etcd

shtripat:
Signed-off-by: Shubhendu <shtripat@redhat.com>

### DIFF
--- a/glusterd2/store/config.go
+++ b/glusterd2/store/config.go
@@ -20,6 +20,15 @@ const (
 	etcdCURLsOpt     = "etcdcurls"
 	etcdPURLsOpt     = "etcdpurls"
 	etcdLogFileOpt   = "etcdlogfile"
+	useTLS           = "usetls"
+	srvrCertFile     = "servercertfile"
+	srvrKeyFile      = "serverkeyfile"
+	caFile           = "caFile"
+	trustedCAFile    = "trustedcafile"
+	clntCertFile     = "clientcertfile"
+	clntKeyFile      = "clientkeyfile"
+	peerCertFile     = "peercertfile"
+	peerKeyFile      = "peerkeyfile"
 
 	defaultEtcdLogFile = "etcd.log"
 
@@ -42,9 +51,18 @@ type Config struct {
 	CURLs     []string
 	PURLs     []string
 	NoEmbed   bool
+	UseTLS    bool
 
-	Dir      string
-	ConfFile string
+	Dir           string
+	ConfFile      string
+	SrvrCertFile  string
+	SrvrKeyFile   string
+	CAFile        string
+	TrustedCAFile string
+	ClntCertFile  string
+	ClntKeyFile   string
+	PeerCertFile  string
+	PeerKeyFile   string
 }
 
 // NewConfig returns a new store Config with defaults
@@ -54,8 +72,17 @@ func NewConfig() *Config {
 		[]string{elasticetcd.DefaultCURL},
 		[]string{elasticetcd.DefaultPURL},
 		false,
+		config.GetBool(useTLS),
 		path.Join(config.GetString("localstatedir"), "store"),
 		path.Join(config.GetString("localstatedir"), storeConfFile),
+		config.GetString(srvrCertFile),
+		config.GetString(srvrKeyFile),
+		config.GetString(caFile),
+		config.GetString(trustedCAFile),
+		config.GetString(clntCertFile),
+		config.GetString(clntKeyFile),
+		config.GetString(peerCertFile),
+		config.GetString(peerKeyFile),
 	}
 }
 
@@ -109,9 +136,61 @@ func GetConfig() *Config {
 		saveconf = true
 		conf.PURLs = purls
 	}
+
+	srvrcertfile := config.GetString(srvrCertFile)
+	if len(srvrcertfile) > 0 {
+		saveconf = true
+		conf.SrvrCertFile = srvrcertfile
+	}
+
+	srvrkeyfile := config.GetString(srvrKeyFile)
+	if len(srvrkeyfile) > 0 {
+		saveconf = true
+		conf.SrvrKeyFile = srvrkeyfile
+	}
+
+	cafile := config.GetString(caFile)
+	if len(cafile) > 0 {
+		saveconf = true
+		conf.CAFile = cafile
+	}
+
+	trustedcafile := config.GetString(trustedCAFile)
+	if len(trustedcafile) > 0 {
+		saveconf = true
+		conf.TrustedCAFile = trustedcafile
+	}
+
+	clntcertfile := config.GetString(clntCertFile)
+	if len(clntcertfile) > 0 {
+		saveconf = true
+		conf.ClntCertFile = clntcertfile
+	}
+
+	clntkeyfile := config.GetString(clntKeyFile)
+	if len(clntkeyfile) > 0 {
+		saveconf = true
+		conf.ClntKeyFile = clntkeyfile
+	}
+
+	peercertfile := config.GetString(peerCertFile)
+	if len(peercertfile) > 0 {
+		saveconf = true
+		conf.PeerCertFile = peercertfile
+	}
+
+	peerkeyfile := config.GetString(peerKeyFile)
+	if len(peerkeyfile) > 0 {
+		saveconf = true
+		conf.PeerKeyFile = peerkeyfile
+	}
 	if config.IsSet(noEmbedOpt) {
 		saveconf = true
 		conf.NoEmbed = config.GetBool(noEmbedOpt)
+	}
+	if config.IsSet(useTLS) {
+		saveconf = true
+		conf.UseTLS = config.GetBool(useTLS)
 	}
 
 	if saveconf {
--- a/glusterd2/store/config.go
+++ b/glusterd2/store/config.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path"
 
+	"github.com/gluster/glusterd2/constants"
 	"github.com/gluster/glusterd2/pkg/elasticetcd"
 
 	"github.com/pelletier/go-toml"
@@ -20,12 +21,6 @@ const (
 	etcdCURLsOpt     = "etcdcurls"
 	etcdPURLsOpt     = "etcdpurls"
 	etcdLogFileOpt   = "etcdlogfile"
-	useTLS           = "usetls"
-	certFile         = "cert-file"
-	keyFile          = "key-file"
-	caFile           = "caFile"
-	clntCertFile     = "client-cert-file"
-	clntKeyFile      = "client-key-file"
 
 	defaultEtcdLogFile = "etcd.log"
 
@@ -50,13 +45,13 @@ type Config struct {
 	NoEmbed   bool
 	UseTLS    bool
 
-	Dir           string
-	ConfFile      string
-	CertFile      string
-	KeyFile       string
-	CAFile        string
-	ClntCertFile  string
-	ClntKeyFile   string
+	Dir          string
+	ConfFile     string
+	CertFile     string
+	KeyFile      string
+	CAFile       string
+	ClntCertFile string
+	ClntKeyFile  string
 }
 
 // NewConfig returns a new store Config with defaults
@@ -66,14 +61,14 @@ func NewConfig() *Config {
 		[]string{elasticetcd.DefaultCURL},
 		[]string{elasticetcd.DefaultPURL},
 		false,
-		config.GetBool(useTLS),
+		config.GetBool(constants.UseTLS),
 		path.Join(config.GetString("localstatedir"), "store"),
 		path.Join(config.GetString("localstatedir"), storeConfFile),
-		config.GetString(certFile),
-		config.GetString(keyFile),
-		config.GetString(caFile),
-		config.GetString(clntCertFile),
-		config.GetString(clntKeyFile),
+		config.GetString(constants.CertFile),
+		config.GetString(constants.KeyFile),
+		config.GetString(constants.CAFile),
+		config.GetString(constants.ClntCertFile),
+		config.GetString(constants.ClntKeyFile),
 	}
 }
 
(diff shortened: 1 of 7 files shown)


shtripat:
This is a work in progress PR. Need to verify the changes with certificates in place.
PLEASE DO NOT MERGE.

prashanthpai:
add to whitelist

shtripat:
Verified the changes with below steps

1. Followed https://coreos.com/os/docs/latest/generate-self-signed-certificates.html for generating self-signed certificates for etcd using `cfssl`

2. Added the certificate details in gd2 conf file as below

```
workdir: "/var/lib/gd2"
peeraddress: "{hostname}:24008"
clientaddress: "{hostname}:24007"
etcdcurls: "https://{hostname}:2379"
etcdpurls: "https://{hostname}:2380"
usetls: true
servercertfile: "/var/lib/gd2/certificates/server.pem"
serverkeyfile: "/var/lib/gd2/certificates/server-key.pem"
cafile: "/var/lib/gd2/certificates/ca.pem"
trustedcafile: "/var/lib/gd2/certificates/ca.pem"
clientcertfile: "/var/lib/gd2/certificates/client.pem"
clientkeyfile: "/var/lib/gd2/certificates/client-key.pem"
peercertfile: "/var/lib/gd2/certificates/member1.pem"
peerkeyfile: "/var/lib/gd2/certificates/member1-key.pem"
```

3. Started gd2 using command `./build/glusterd2 --config conf.yaml`

4. Verified etcd connection using CURL command `curl --cacert /var/lib/gd2/certificates/ca.pem --cert /var/lib/gd2/certificates/client.pem --key /var/lib/gd2/certificates/client-key.pem -XGET https://{hostname}:2379/v2/keys` and it shows the below output

```
{"action":"get","node":{"dir":true}}
```

5. Tried running curl without passing certificate details like `curl -XGET https://{hostname}:2379/v2/keys` and the output is as below

```
curl: (60) SSL certificate problem: unable to get local issuer certificate
More details here: https://curl.haxx.se/docs/sslcerts.html

curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option.
HTTPS-proxy has similar options --proxy-cacert and --proxy-insecure.
```

shtripat:
@kshlm @aravindavk @atinmu @prashanthpai please review the changes.

shtripat:
Also tested by disabling TLS as below

```
workdir: "/var/lib/gd2"
peeraddress: "10.70.42.213:24008"
clientaddress: "10.70.42.213:24007"
etcdcurls: "http://10.70.42.213:2379"
etcdpurls: "http://10.70.42.213:2380"
usetls: false
```

and works as expected. Tried now the command `curl -XGET http://10.70.42.213:2379/v2/keys` and shows the below output

```
{"action":"get","node":{"dir":true}}
```

glusterd2/store/remote.go
```diff
@@ -1,19 +1,50 @@
...
 
 func newRemoteStore(conf *Config) (*GDStore, error) {
-
-	c, e := clientv3.New(clientv3.Config{
+	var tlsConfig *tls.Config
+	var c *clientv3.Client
+	var e error
+	if conf.UseTLS {
+		tlsConfig = &tls.Config{
+			MinVersion: tls.VersionTLS10,
```
kshlm: These are the defaults for these. There is no need to set them unless you change to something different. I'd prefer if we just stick to tls1.2, but that's just my opinion.

pkg/elasticetcd/config.go
```diff
@@ -9,12 +9,20 @@ import (
...
-	DefaultName      = "elasticetcd"
-	DefaultIdealSize = 3
-	DefaultDir       = "."
+	DefaultCURL          = "http://0.0.0.0:2379"
+	DefaultPURL          = "http://0.0.0.0:2380"
+	DefaultEndpoint      = "http://localhost:2379"
+	DefaultName          = "elasticetcd"
+	DefaultIdealSize     = 3
+	DefaultDir           = "."
+	DefaultCertFile      = "/var/lib/gd2/certificates/server.crt"
```
kshlm: The elasticetcd package should not contain defaults referencing GD2. It is expected to remain a standalone package, without any ties to the rest of GD2. Please move these defaults to GD2, preferrably to the store package for now.

Also, we shouldn't hardcode paths, as GD2 can use different working/config/state directories. Instead, I'd suggest setting a default directory name for a certs directory, and default filenames for the certs. Then you could build the paths based on the configured directories when you need to pass the paths. For eg.
```go
const (
  DefaultCertDir = "certificates"
  DefaultCertFile = "server.crt"
)

serverCert := path.Join(config.GetString("localstatedir", DefaultCertDir, DefaultCertFile)
```

pkg/elasticetcd/config.go
```diff
@@ -69,18 +77,33 @@ type Config struct {
...
-		PURLs:     defaultPURLs,
-		IdealSize: DefaultIdealSize,
+		Name:          DefaultName,
+		Dir:           DefaultDir,
+		LogDir:        path.Join(DefaultDir, "log"),
+		Endpoints:     defaultEndpoints,
+		CURLs:         defaultCURLs,
+		PURLs:         defaultPURLs,
+		IdealSize:     DefaultIdealSize,
+		CertFile:      DefaultCertFile,
```
kshlm: It's not necessary for the cert files to be set in the default config, as they are not necessary for the operation of elasticetcd.

shtripat:
>> You should definitely move the defaults that have been set in elasticetcd out, and into GD2.

Ack. 

>> Also, I'd like to understand what all the different certs are for. And if possible, reduce the number.

The no of certificates is as per the guidelines from etcd documents. For sure server and client certificates are to be different. It like using CA we create server certificates which is private to server. The client certificates are to be given to clients which connect to the server. Yes the client certificates could be same for all. Still I strongly feel the peer certificates should be different from client-server combo. We should not mix client-to-server communication with peer-to-peer communication.

So in total we need at least below files I feel

```
ca.pem
server.pem, server-key.pem
client.pem, client-key.pem
peer.pem, peer-client.pem
```

comments??

glusterd2/store/config.go
```diff
@@ -42,9 +51,18 @@ type Config struct {
...
 	PURLs     []string
 	NoEmbed   bool
-
-	Dir      string
-	ConfFile string
+	UseTLS    bool
+
+	Dir           string
+	ConfFile      string
+	SrvrCertFile  string
```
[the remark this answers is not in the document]
shtripat: Server has its own certificate and key and server should be started that set.
Clients get a public key and certificate which they use for getting authenticated at server.
Similarly peer-to-peer communication is also authenticated using a pair of certificate and key.

Same server certificates and keys it possible as we need below json input for `cfssl`

```
{
    "CN": "My own CA",
    "hosts": [
        "{host1}",
        "{host2}",
        ..........
    ],
    "key": {
        "algo": "ecdsa",
        "size": 256
    },
    "names": [
        {
            "C": "US",
            "ST": "CA",
            "L": "San Francisco"
        }
    ]
}
```

I think similar thing is possible for clients as well with below json input

```
{
    "CN": "My own CA",
    "hosts": ["{clnt1}", "clnt2", ......],
    "key": {
        "algo": "ecdsa",
        "size": 256
    },
    "names": [
        {
            "C": "US",
            "ST": "CA",
            "L": "San Francisco"
        }
    ]
}
```

glusterd2/store/remote.go
shtripat: Ack

pkg/elasticetcd/config.go
shtripat: Agree with dirname and filename thing. Will change accordingly.

pkg/elasticetcd/config.go
shtripat: Ack. Will remove these.

glusterd2/store/remote.go
shtripat: @kshlm do you want min version to be set as 1.2?

pkg/elasticetcd/config.go
shtripat: Now its removed from EE config and whatever value is set in config file would be considered as full path if the certificates. User can choose to keep the certs in any path.

kshlm:
The recommendations in the etcd document are for general etcd deployments, where you have an etcd cluster, and external consumers. In that case it makes sense to split TLS for internal etcd cluster interface (peer cert), and external client API (server cert and client cert).

In GD2 with elasticetcd, we are a single entity. We are the etcd cluster doing the internal operations, and serving etcd kv store with ourselves as clients. So, it is okay (IMO) for us to be using a single certificate which identifies us, for all cases.

If we really need to, we could split up TLS configuration for internal and external traffic. A cert and CA for internal traffic (etcd peer traffic and GD2 gRPC traffic). And a seperate cert and CA for external facing interfaces (GD2 ReST, SunRPC, etcd client API). External clients (GD2 or etcd) get their own certs, signed using the extrenal interface CA.

shtripat:
I agree to the suggestion of two sets one for internal traffic and one for external traffic.
Should we include gRPC, SunRPC and GD2 REST as well as part of this PR only?

I feel its better we rename the certs and key configuration names accordingly in this PR. And we have separate PRs for enabling TLS for gRPC, SunRPC and GD2 REST using the same cert configurations.

In that case I suggest below configurations to be maintained

CAFile: The pem file for certifying authority
GD2IntCertFile: Certificate for internal communication (etcd peer-to-peer and gRPC)
GD2IntCertFile: Key for internal communication (etcd peer-to-peer and gRPC)
GD2ExtCertFile: Certificate for external communication (GD2 REST, SunRPC and etcd client)
GD2ExtKeyFile: Key for external communication (GD2 REST, SunRPC and etcd client)

Suggestions? @prashanthpai @aravindavk @amarts @atinmu 
@kshlm are we ok with the names of configurations? or may be we can have name suggestions.

kshlm:
Let it be 'Client*' and 'Peer*' for external and internal respectively. We already call the addresses we listen on as ClientAddress and PeerAddress.

shtripat:
Now we have client and server certs only. Instead of calling peer certs its just `CertFile` and `KeyFile` for server.

shtripat:
Can others also please review the PR?

atinmu:
@shtripat What's pending on this patch? Are you yet to address comments from @kshlm ? I see this patch needs a rebase.

shtripat:
@atinmu comments from Kaushal taken care already. Will rebase and push again. @kshlm needs to have a look once more.

glusterd2/store/config.go
```diff
@@ -20,6 +20,12 @@ const (
 	etcdCURLsOpt     = "etcdcurls"
 	etcdPURLsOpt     = "etcdpurls"
 	etcdLogFileOpt   = "etcdlogfile"
+	useTLS           = "usetls"
+	certFile         = "cert-file"
+	keyFile          = "key-file"
+	caFile           = "caFile"
+	clntCertFile     = "client-cert-file"
```
kshlm: It would be preferrable to add this to `glusterd2/config.go`. I'd want to use the same options for tls certs for other servers.

shtripat: done. will update the PR accordingly.

shtripat: Moving the constants to `glusterd2/config.go` causes cyclic dependencies as below

```
package github.com/gluster/glusterd2/glusterd2
	imports github.com/gluster/glusterd2/glusterd2/daemon
	imports github.com/gluster/glusterd2/glusterd2/events
	imports github.com/gluster/glusterd2/glusterd2/store
	imports github.com/gluster/glusterd2/glusterd2
```

How to go about this?

shtripat: I feel rather we can have a `commons` package with all the common stuff defined which could be used across modules like `glusterd2`, `glustercli` etc.

@kshlm thoughts??

kshlm:
@shtripat Still failing lint checks on [centos-ci][1].

> 10:06:21 ./pkg/elasticetcd/client.go does not conform to gofmt rules. Run: gofmt -s -w ./pkg/elasticetcd/client.go
> 10:06:21 ./pkg/elasticetcd/config.go does not conform to gofmt rules. Run: gofmt -s -w ./pkg/elasticetcd/config.go
> 10:06:22 ./commons/constants.go:4:2: exported const UseTLS should have comment (or a comment on this block) or be unexported

Also, when fixing this. We keep common packages under `pkg/`.

[1]: https://ci.centos.org/job/gluster_glusterd2/820/console

shtripat:
Ah I ran only ran `gofmt -d {dir}`. Will do the needed and update.
Also is it fine if I move `commons/constants.go` to `pkg/commons/constants.go` ?

pkg/commons/constants.go
```diff
@@ -0,0 +1,11 @@
+package commons
```
prashanthpai: We have tried to avoid doing something like if the reason was just to circumvent cyclic dependency. I'm okay duplicating this if multiple packages require just these constants.

shtripat: @kshlm @aravindavk what is your thought on this?

kshlm: I'm actually okay with doing this, as I don't want to duplicate code.

shtripat: @prashanthpai also its not just about avoiding the cyclic dependency. Rather its more to do with reuse of common constants. I personally agree with @kshlm here. @aravindavk ?

aravindavk: commons package looks good to me, otherwise we will end up editing in multiple places if we have to change.

prashanthpai: Okay. Fair enough.

I'm afraid that we might end up putting lot more stuff (like struct definitions) under `common` [because of its name](https://blog.golang.org/package-names). I want to discourage that practice. IIUC, the `pkg` directory is meant to be home for reusable code like libraries which can be used by other projects too.

What are you thoughts about naming this package as `constants` because that's what it provides ?

shtripat: You a separate package `constants` parallel to `pkg`? I am fine with that.

PR #490, opened 2017-12-18
quota plugin stubs

sanoj-unnikrishnan:
Initial plugin with stubs required for quota.
Sending this so as to divide development task
and code in parallel.

Updates: #421
Signed-off-by: Sanoj Unnikrishnan <sunnikri@redhat.com>

### DIFF
--- a/plugins/quota/api/req.go
+++ b/plugins/quota/api/req.go
@@ -1,0 +1,13 @@
+package quota
+
+// SetLimitReq represents REST API request to Limit Usage/objects of a directory
+type SetLimitReq struct {
+	Path             string `json:"path"`
+	SizeUsageLimit   int    `json:"size-usage-limit,omitempty"`
+	ObjectCountLimit int    `json:"object-count-limit,omitempty"`
+}
+
+// RemoveLimitReq represents REST API request to Remove Usage/objects of a directory
+type RemoveLimitReq struct {
+	Path string `json:"path"`
+}
--- a/plugins/quota/api/resp.go
+++ b/plugins/quota/api/resp.go
@@ -1,0 +1,31 @@
+package quota
+
+type crawlInfo struct {
+	CrawlPid int `json:"crawl-pid"`
+	MountPid int `json:"crawl-mount-pid"`
+}
+
+type list struct {
+	Path      string `json:"path"`
+	HardLimit int64  `json:"hard-limit"`
+	SoftLimit int64  `json:"soft-limit"`
+	Used      int64  `json:"used"`
+	Available int64  `json:"available"`
+	/*
+	 * TODO / Review: "Merge soft limit exceeded and hard limit
+	 * exceeded" flags into a single field called status.
+	 */
+	SoftLimitExceeded bool `json:"soft-limit-exceeded"`
+	HardLimitExceeded bool `json:"hard-limit-exceeded"`
+	// LimitType indicates directory quota/inode-quota/...
+	LimitType int32 `json:"limit-type"`
+}
+
+//ListResp is an array of structs representing individual limits.
+type ListResp []list
+
+//DisableResp gives the information of disable crawler on success
+type DisableResp crawlInfo
+
+//EnableResp gives the information of enable crawler on success
+type EnableResp crawlInfo
--- a/plugins/quota/init.go
+++ b/plugins/quota/init.go
@@ -1,0 +1,62 @@
+package quota
+
+import (
+	"github.com/gluster/glusterd2/glusterd2/servers/rest/route"
+	"github.com/prashanthpai/sunrpc"
+)
+
+// Plugin is a structure which implements GlusterdPlugin interface
+type Plugin struct {
+}
+
+// Name returns name of plugin
+func (p *Plugin) Name() string {
+	return "quota"
+}
+
+// SunRPCProgram returns sunrpc program to register with Glusterd
+func (p *Plugin) SunRPCProgram() sunrpc.Program {
+	return nil
+}
+
+// RestRoutes returns list of REST API routes to register with Glusterd
+func (p *Plugin) RestRoutes() route.Routes {
+	return route.Routes{
+		route.Route{
+			Name:        "QuotaEnable",
+			Method:      "POST",
+			Pattern:     "/quota/{volname}",
+			Version:     1,
+			HandlerFunc: quotaEnableHandler},
+		route.Route{
+			Name:        "QuotaDisable",
+			Method:      "DELETE",
+			Pattern:     "/quota/{volname}",
+			Version:     1,
+			HandlerFunc: quotaDisableHandler},
+		route.Route{
+			Name:        "QuotaList",
+			Method:      "GET",
+			Pattern:     "/quota/{volname}/limit",
+			Version:     1,
+			HandlerFunc: quotaListHandler},
+		route.Route{
+			Name:        "QuotaLimit",
+			Method:      "POST",
+			Pattern:     "/quota/{volname}/limit",
+			Version:     1,
+			HandlerFunc: quotaLimitHandler},
+		route.Route{
+			Name:        "QuotaRemove",
+			Method:      "DELETE",
+			Pattern:     "/quota/{volname}/limit",
+			Version:     1,
+			HandlerFunc: quotaRemoveHandler},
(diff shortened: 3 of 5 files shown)


prashanthpai:
add to whitelist

atinmu:
From the centos-ci 

```./glusterd2/plugin/plugins.go does not conform to gofmt rules. Run: gofmt -s -w ./glusterd2/plugin/plugins.go```

File: plugins/quota/init.go [hunk 1]
```diff
@@ -0,0 +1,110 @@
...
+		route.Route{
+			Name:        "QuotaRemoveObjects",
+			Method:      "POST",
+			Pattern:     "/quota/{volname}/remove-objects",
+			Version:     1,
+			HandlerFunc: quotaRemoveObjectsHandler},
+		route.Route{
+			Name:        "QuotaDefaultSoftLimit",
+			Method:      "POST",
+			Pattern:     "/quota/{volname}/default-soft-limit",
```
aravindavk: We can have single route for all the configurations

    POST /quota/{volname}/config

Request body:

    {
        "default-soft-limit": 70
    }

limit-usage, default-soft-limit, alert-time, soft-timeout, hard-timeout

and reset API

    DELETE /quota/{volname}/config

prashanthpai: All options that are [part of quota xlator](https://github.com/gluster/glusterfs/blob/master/xlators/features/quota/src/quota.c#L5263) can be a normal volume set option.

sanoj-unnikrishnan: @aravindavk , @prashanthpai, 

I don't see an issue with suggestion by @prashanthpai as long as the necessary validations can be done in volume set path (volume is started , quota is enabled).

The glustercli would have to be mapped to the volume set route to maintain backward compatibility. So as far as glustercli is concerned, the same volume option would be visible in 2 ways. 
eg:
> gluster_cli quota <volname> soft-timeout <value>
> gluster volume set <volname> quota.soft-timeout <value>
Is this ok ?  I think yes!

If we agree, will also have to refresh the patch to make the options settable again
https://review.gluster.org/#/c/18743/4/xlators/features/quota/src/quota.c

atinmu: We need to have an answer to https://github.com/gluster/glusterd2/issues/501 sooner than later to get to the decision on the above.

sanoj-unnikrishnan: Discussion notes:
with @aravindavk , @harigowtham

Decided to go with options route for setting quota config (as suggested by @prashanthpai).
#501 would be addressed to have validators which quota can also make use of.

sanoj-unnikrishnan:
update pull request with review comments.
@aravindavk , @prashanthpai  please review.

plugins/quota/api/req.go
```diff
@@ -0,0 +1,13 @@
+package quota
+
+// QuotaLimitReq represents REST API request to Limit Usage/objects of a directory
+type QuotaLimitReq struct {
+	Path        string `json:"path"`
+	UsageLimit  int    `json:"UsageLimit,omitempty"`
```
aravindavk: Use small case for json names "usage_limit", "object_limit" etc

plugins/quota/init.go
```diff
@@ -0,0 +1,86 @@
...
+
+// RestRoutes returns list of REST API routes to register with Glusterd
+func (p *Plugin) RestRoutes() route.Routes {
+	return route.Routes{
+		route.Route{
+			Name:        "QuotaHelp",
+			Method:      "GET",
+			Pattern:     "/quota/help",
+			Version:     1,
+			HandlerFunc: quotaHelpHandler},
```
aravindavk: This Route can be removed, generated from the helper script

plugins/quota/api/req.go
```diff
@@ -0,0 +1,13 @@
+package quota
+
+// LimitReq represents REST API request to Limit Usage/objects of a directory
+type LimitReq struct {
+	Path        string `json:"path"`
+	UsageLimit  int    `json:"usage-limit,omitempty"`
```
prashanthpai: I understand that these names come from how these features were implemented. However, we have a chance to change these legacy names for better ones. How about `UsageLimitSize` and `UsageLimitInode`. The names are intuitive.

plugins/quota/api/req.go
```diff
@@ -0,0 +1,13 @@
+package quota
+
+// LimitReq represents REST API request to Limit Usage/objects of a directory
+type LimitReq struct {
```
prashanthpai: You can name this as `SetLimitReq` and the other one as `RemoveLimitReq`

plugins/quota/api/resp.go
```diff
@@ -0,0 +1,25 @@
+package quota
+
+type crawlInfo struct {
+	FindPid  int `json:"crawl-pid"`
```
prashanthpai: We need better names for `FindPid`

plugins/quota/init.go
```diff
@@ -0,0 +1,80 @@
...
+	return route.Routes{
+		route.Route{
+			Name:        "QuotaEnable",
+			Method:      "POST",
+			Pattern:     "/quota/{volname}",
+			Version:     1,
+			HandlerFunc: quotaEnableHandler},
+		route.Route{
+			Name:        "QuotaDisable",
+			Method:      "DELETE",
```
prashanthpai: This can be `POST` instead as no resource is being deleted. And the enable and disable APIs be `/quota/{volname}/enable` and ``/quota/{volname}/disable``

plugins/quota/init.go
```diff
@@ -0,0 +1,80 @@
...
+			Pattern:     "/quota/{volname}",
+			Version:     1,
+			HandlerFunc: quotaDisableHandler},
+		route.Route{
+			Name:        "QuotaList",
```
prashanthpai: You dont need many APIs. You can merge both size and inode quota as a single API - one to set, one to list, one to remove.

Is it more expensive to get both information (size and inode quota) from backend ?

plugins/quota/init.go
harigowtham: with quota disable, the config file and a few xttars are deleted.Similarly with enable we create/set a few xttars. So Aravinda commented that its better to make use of the methods in the API for this instead of a separate API itself. and thus this change was done. And as a few things are removed I feel it makes sense to have it as DELETE method instead for the quota API instead of new API itself. The same goes for POST as enable. @aravindavk if you have any thoughts do let us know.

sanoj-unnikrishnan:
.

plugins/quota/api/req.go
sanoj-unnikrishnan: Agreed!

plugins/quota/api/req.go
sanoj-unnikrishnan: felt that usage would intutively mean disk space usage..
I find UsageLimitSize / UsageLimitInode a little odd.
How About:
SizeUsageLimit
ObjectCountLimit

plugins/quota/api/resp.go
sanoj-unnikrishnan: Agreed. will change to CrawlPid.

plugins/quota/init.go
sanoj-unnikrishnan: Delete does delete all the limit that has been set previously.
the quota.conf (file / data from etcd) will be removed.

plugins/quota/api/req.go
prashanthpai: True. My bad. That sounds better.

plugins/quota/init.go
sanoj-unnikrishnan: There is no additional cost in setting / getting / removing it.
I think it is a good idea to merge the two, though directory quota alone is the more stable and widely used option.
@aravindavk  Do u agree in it be merged into single API.

aravindavk:
retest this please

atinmu:
@prashanthpai any comments not yet incorporated?

plugins/quota/init.go
aravindavk: +1

prashanthpai:
> @prashanthpai any comments not yet incorporated?

@atinmu Sanoj is refreshing this with quota list API update. I'll merge this right after that.

Issue #404, opened 2017-11-07
Option validation for volume set


kshlm:
Right now volume set doesn't validate the options being set. But it needs to. Fixing this take two parts.

- [ ] Add validation functions for the available option types
- [ ] Call validation functions from volume set

kshlm:
@devyanikota If you want to pick this up, please assign it to your self.

devyanikota:
@kshlm I think I don't have the rights to assign this to myself(no gear sign againt the 'assignees').
Can you please do the same. Thank you.

kshlm:
I've fixed permissions, you should be able to assign yourself to issues now. I've assigned this to you in any case.

kshlm:
@devyanikota I've updated the issue summary with what I would have done. I think its clear enough, but let me know if you have any doubts.

atinmu:
@devyanikota what's pending for this issue?

atinmu:
@devyanikota Could you update us on the pending things for this issue?

PR #494, opened 2017-12-19
Updates #404: Add validations for Internet Address.

devyanikota:

### DIFF
--- a/glusterd2/xlator/options/options.go
+++ b/glusterd2/xlator/options/options.go
@@ -2,6 +2,7 @@ package options
 
 import (
 	"errors"
+	"net"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -186,12 +187,29 @@ func ValidateInt(o *Option, val string) error {
 
 // ValidateInternetAddress validates the Internet Address
 func ValidateInternetAddress(o *Option, val string) error {
-	return ErrInvalidArg
+	_, _, err := net.ParseCIDR(val)
+	if err != nil {
+		return ErrInvalidArg
+	} else if validate.IsIP(val) != true {
+		return ErrInvalidArg
+	} else if validate.IsHost(val) != true {
+		return ErrInvalidArg
+	} else if strings.ContainsAny(val, "* & # & ? & ^") == true {
+		return ErrInvalidArg
+	}
+	return nil
 }
 
 // ValidateInternetAddressList validates the Internet Address List
 func ValidateInternetAddressList(o *Option, val string) error {
-	return ErrInvalidArg
+	iplist := strings.Split(val, ",")
+	for _, ip := range iplist {
+		err := ValidateInternetAddress(o, ip)
+		if err != nil {
+			return ErrInvalidArg
+		}
+	}
+	return nil
 }
 
 // ValidatePath validates if the option is a valid path


glusterd2/xlator/options/options.go
```diff
@@ -186,12 +187,29 @@ func ValidateInt(o *Option, val string) error {
 
 // ValidateInternetAddress validates the Internet Address
 func ValidateInternetAddress(o *Option, val string) error {
-	return ErrInvalidArg
+	_, _, err := net.ParseCIDR(val)
+	if err != nil {
+		return ErrInvalidArg
+	} else if validate.IsIP(val) != true {
```
atinmu: Is `validate.IsIP` taking consideration of validating both Ipv4 and IPv6?

atinmu: Shouldn't validateInternetAddress also consider a valid hostname to be a correct val passed in?

devyanikota: @atinmu yes, validate.IsIP validates IPv4 and IPv6 addresses too.

devyanikota: @atinmu validate.IsHost() takes care of that.

devyanikota:
@kshlm @atinmu while discussing with @aravindavk, we realized that since  GF_OPTION_TYPE_INTERNET_ADDRESS  doesn't seem to have the _SETTABLE_ flag, so we might not need the validations?
Please give your opinion. Thank you.

glusterd2/xlator/options/options.go
atinmu: My bad, sorry.

atinmu:
> @kshlm @atinmu while discussing with @aravindavk, we realized that since GF_OPTION_TYPE_INTERNET_ADDRESS doesn't seem to have the SETTABLE flag, so we might not need the validations?

@devyanikota I may not have the complete background of the intention of the PR, but if we don't need the validations then why are these functions in place? Wouldn't a peer probe code path will hit this validation?

aravindavk:
> @devyanikota I may not have the complete background of the intention of the PR, but if we don't need the validations then why are these functions in place? Wouldn't a peer probe code path will hit this validation?

I think we don't need validation if SETTABLE is not specified.

There are some internal options, which will gets added in volfile but user will not set those options. Mostly picked from peer info and added during volfile generation. Even if we add these validations, it will never gets used since none of these options are set by API users.

I don't think peer probe will use this validation. Others can confirm.

atinmu:
> I think we don't need validation if SETTABLE is not specified.There are some internal options, which will gets added in volfile but user will not set those options. Mostly picked from peer info and added during volfile generation. Even if we add these validations, it will never gets used since none of these options are set by API users. I don't think peer probe will use this validation. Others can confirm.

If I can map this functionality back in GD1 & CLI I see `valid_internet_address` function is being called in `cli_cmd_peer_probe_cbk` .

kshlm:
While INTERNET_ADDRESS doesn't have any settable options, we do have settable INTERNET_ADDRESS_LIST (auth.allow/reject) which basically needs the same checks. As @devyanikota has implemented `ValidateInternetAddressList` on top of `ValidateInternetAddress`, I don't see a problem in keeping this as is.

As for the other comments about validating arguments in CLI, they could technically call the validation functions in the options package to perform the validation. The only thing that sort of prevents this right now is that the `options` package is under `glusterd2` directory, but we can move it to be under `pkg`.

atinmu:
So this is more from a option specific validation function than a generic
one then. @devyanikota would you take care of the peer probe validations as
highlighted in a separate PR then?

PR #505, opened 2017-12-25
Volume group profile functionality

kmlmhnn:


glusterd2/commands/volumes/options.go
```diff
@@ -0,0 +1,13 @@
+package volumecommands
+
+type option struct {
+	OptionName string
+	OnValue    string
+	OffValue   string
+}
+
+// GroupOptions maps from a profile name to a set of options
+var GroupOptions = map[string][]option{
```
atinmu: Does this mean we 'd need to have all the group profile options  of each profile dumped into this structure? If that's the case, then this is not flexible. Consider someone wants to create his/her own custom profile with different tunables. How can we achieve it?

kmlmhnn: GroupOptions map is loaded once into the store during initialization and it is fetched from the store each time a group profile option is set/unset.
So if we allow the user to modify the in-store state of GroupOptions, that would allow the creation of custom profiles.

atinmu:
As we discussed, we should provide APIs to allow users to define their own custom profiles. For all the default and tested profiles like virt, md-cache I don't think it makes much sense to allow users to tweak in the parameters but for custom profiles it does. Also along with that, an API to unset such profiles would be very good to have. Additionally if one can use an API to view what all tunables are part of a custom profile it'd be ideal.

glusterd2/main.go
```diff
@@ -97,6 +100,15 @@ func main() {
 		log.WithError(err).Fatal("Could not add self details into etcd")
 	}
 
+	// Load the group option map into the store
```
prashanthpai: You can move this to a separate function (in volumes package) and call that function.

glusterd2/commands/volumes/options.go
```diff
@@ -0,0 +1,13 @@
+package volumecommands
```
prashanthpai: Please name this file as `grouped-options` or something else.

glusterd2/commands/volumes/options.go
prashanthpai: Please name this as `DefaultGroupOptions`

glusterd2/commands/volumes/options.go
```diff
@@ -0,0 +1,13 @@
...
+
+type option struct {
+	OptionName string
+	OnValue    string
+	OffValue   string
+}
+
+// GroupOptions maps from a profile name to a set of options
+var GroupOptions = map[string][]option{
+	"profile.test": {{"afr.eager-lock", "on", "off"},
```
prashanthpai: You can actually start with real profiles here as supported by glusterd1.

e2e/volume_ops_test.go
```diff
@@ -237,6 +237,41 @@ func TestVolumeOptions(t *testing.T) {
...
+		_, err = client.VolumeCreate(createReq)
+		r.Nil(err)
+
+		err = client.VolumeDelete(volname)
+		r.Nil(err)
+	}
+
+	profileCreateReq := api.ProfileCreateReq{
+		ProfileName: "profile.test2",
+		Options: []api.Option{{"opt1", "on", "off"},
```
prashanthpai: Shouldn't this fail as the option names are invalid ?

glusterd2/commands/volumes/commands.go
```diff
@@ -32,6 +32,18 @@ func (c *Command) Routes() route.Routes {
 			Pattern:     "/volumes/{volname}/options",
 			Version:     1,
 			HandlerFunc: volumeOptionsHandler},
+		route.Route{
+			Name:        "VolumeProfileTunables",
```
prashanthpai: Users can confuse this name with the [volume profile](http://docs.gluster.org/en/latest/Administrator%20Guide/Monitoring%20Workload/) command which records the fops and collects I/O metrics.

How about `/volumes/group-options` endpoint ? Avoid using profile elsewhere too.

@kshlm @aravindavk Thoughts ?

glusterd2/commands/volumes/commands.go
```diff
@@ -32,6 +32,18 @@ func (c *Command) Routes() route.Routes {
...
+			Method:      "GET",
+			Pattern:     "/volumes/options/{profilename}",
```
prashanthpai: As it's inexpensive to `GET` all group options at once from etcd, you can simply provide that API which returns everything instead of just one.

glusterd2/commands/volumes/profile-create.go
```diff
@@ -0,0 +1,57 @@
...
+		return
+	}
+
+	var groupOptions map[string][]api.Option
+	if err := json.Unmarshal(resp.Kvs[0].Value, &groupOptions); err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err.Error(), api.ErrCodeDefault)
+		return
+	}
+
+	if _, ok := groupOptions[req.ProfileName]; ok {
```
prashanthpai: A user should be allowed to replace an existing grouped option.

And an API to delete too ? You can disallow deleting default/built-in grouped option.

pkg/api/volume_resp.go
```diff
@@ -82,3 +82,8 @@ type VolumeExpandResp VolumeInfo
 
 // VolumeListResp is the response sent for a volume list request.
 type VolumeListResp []VolumeGetResp
+
```
prashanthpai: How about this ?

```go
 type VolumeOption struct {
 	Name        string `json:"name"`
 	OnValue     string `json:"onvalue"`
 	OffValue    string `json:"offvalue"`
 }

type GroupOption struct {
        Name         string       `json:"name"`
        Options     []VolumeOption `json:"options"`
}

type GroupOptionReq GroupOption

type GroupOptionListResp []GroupOption

Comment (kshlm) [reply pkg/api/volume_resp.go]: s/GroupOption/OptionGroup/ sounds better.

**Comment by atinmu** [comment]
@kmlmhnn What is pending work in this PR apart from rebasing it? Looks like some of the comments need to be addressed as well. Can we please complete it?

**Comment by kmlmhnn** [comment]
@atinmu Working on it right now.

**Comment by prashanthpai** [comment]
@kmlmhnn Can you rebase and fix merge conflicts ? I merged your xlator validation func PR first.

File: glusterd2/commands/volumes/optiongroup-delete.go [hunk 10]
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/gluster/glusterd2/glusterd2/store"
+	"github.com/gluster/glusterd2/pkg/api"
+
+	"github.com/gorilla/mux"
+)
+
+func optionGroupDeleteHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	groupName := mux.Vars(r)["groupname"]
+
```
prashanthpai: We may need a lock between `Store.Get` here and `Store.Put` below. For now, this is fine.

Issue #467, opened 2017-11-30
CLI: Add tab (bash) completion for glustercli


prashanthpai:
Although this may seem not so important, it can go a long way in improving usability.

PR #508, opened 2017-12-26
cli: Add Bash completion support

aravindavk:
Fixes: #467


glustercli/bash_completion/main.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"os"
+
+	"github.com/gluster/glusterd2/glustercli/cmd"
+)
+
+func main() {
+	filename := os.Args[1]
+	cmd.RootCmd.GenBashCompletionFile(filename)
```
prashanthpai: It'll be good to check (and log/print - you can use go's standard log/fmt package) returned error when `GenBashCompletionFile` is called.

For example, running it as `$ build/bash_completion /etc/bash_completion.d/glustercli.sh` fails silently as only root can write to `/etc/bash_completion.d/`

glustercli/bash_completion/main.go
```diff
@@ -0,0 +1,12 @@
+package main
+
+import (
+	"os"
+
+	"github.com/gluster/glusterd2/glustercli/cmd"
+)
+
+func main() {
+	filename := os.Args[1]
```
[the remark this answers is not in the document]
aravindavk: Skipped validation since it is internal tool used only by Makefile. I will add the check

glustercli/bash_completion/main.go
aravindavk: Ok, I will add the log

aravindavk:
@prashanthpai addressed review comments, please review

prashanthpai:
retest this please

PR #512, opened 2017-12-28
Updates #404: Add validations for Client Auth Address.

devyanikota:
Validating if the "address"
* is (anonymous) OR
* is valid FQDN or valid IPv4/6 address OR
* contains wildcard chars e.g.'*' or '#' or '?' or '^'
* is valid ipv4 subnet pattern (xx.xx.xx.xx/n)

### DIFF
--- a/glusterd2/xlator/options/options.go
+++ b/glusterd2/xlator/options/options.go
@@ -2,7 +2,6 @@ package options
 
 import (
 	"errors"
-	"net"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -94,7 +93,7 @@ func (o *Option) Validate(val string) error {
 	case OptionTypeBool:
 		err = ValidateBool(o, val)
 	case OptionTypeClientAuthAddr:
-		err = ValidateClientAuthAddr(o, val)
+		err = ValidateInternetAddressList(o, val)
 	case OptionTypeDouble:
 		err = ValidateDouble(o, val)
 	case OptionTypeInt:
@@ -158,11 +157,6 @@ func ValidateBool(o *Option, val string) error {
 	}
 }
 
-// ValidateClientAuthAddr validates mount auth address
-func ValidateClientAuthAddr(o *Option, val string) error {
-	return ErrInvalidArg
-}
-
 //ValidateDouble validates if the option is of type double
 func ValidateDouble(o *Option, val string) error {
 	var err error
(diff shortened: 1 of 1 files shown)


devyanikota:
@kshlm since validateInternetAddress() handles 3 of the above validations, I only added the validation of ipv4 subnet pattern in validateClientAuthAddress()
Can you please review the same. Thanks.

glusterd2/xlator/options/options.go
```diff
@@ -160,7 +160,23 @@ func ValidateBool(o *Option, val string) error {
...
 // ValidateClientAuthAddr validates mount auth address
 func ValidateClientAuthAddr(o *Option, val string) error {
-	return ErrInvalidArg
+	if len(val) == 0 {
+		return ErrInvalidArg
+	}
+	err := ValidateInternetAddress(o, val)
+	if err != nil {
+		i := strings.Index(val, "/")
+		if i != -1 && validate.IsIP(val[:i]) {
```
kshlm: You don't need custom validation functions for the subnet patterns, or the CIDR notation, as they're properly called. You should use `net.ParseCIDR`. Or better still use govalidators `IsCIDR`.

devyanikota: @kshlm ok. So since ValidateInternetAddress() already uses net.ParseCIDR to check for subnet patterns, so just calling the same function to handle this validation as well.
Please review. thanks.

glusterd2/xlator/options/options.go
```diff
@@ -160,7 +160,8 @@ func ValidateBool(o *Option, val string) error {
 
 // ValidateClientAuthAddr validates mount auth address
 func ValidateClientAuthAddr(o *Option, val string) error {
-	return ErrInvalidArg
+	err := ValidateInternetAddress(o, val)
```
kshlm: I missed this earlier. `ClientAuthAddr` is supposed to be a comma-seperated list. So you'll need to iterate through the list. Since this is the same as what you do for `ValidInternetAddressList`, just call `ValidateInternetAddressList` for `case OptionTypeClientAuthAddr` in `(*Option)Validate()`.

glusterd2/xlator/options/options.go
```diff
@@ -187,14 +188,17 @@ func ValidateInt(o *Option, val string) error {
 
 // ValidateInternetAddress validates the Internet Address
 func ValidateInternetAddress(o *Option, val string) error {
+	if len(val) == 0 {
+		return ErrInvalidArg
+	}
 	_, _, err := net.ParseCIDR(val)
```
kshlm: govalidator has a function `IsCIDR`. It will be better to use that to keep validators uniform.

glusterd2/xlator/options/options.go
devyanikota: @kshlm rectified the same as suggested. Please review the same. Thank you.

PR #514, opened 2018-01-02
Add validations for the pairs of elements.

devyanikota:
The name of the function ValidatePriorityorSize() is misleading.
I tried replicating the same logic as in the validation of priority in options.c in GD1.

### DIFF
--- a/glusterd2/xlator/options/options.go
+++ b/glusterd2/xlator/options/options.go
@@ -251,7 +251,19 @@ func ValidatePercentOrSize(o *Option, val string) error {
 
 // ValidatePriorityOrSize validates either priority or size
 func ValidatePriorityOrSize(o *Option, val string) error {
-	return ErrInvalidArg
+	pairs := strings.Split(val, ',')
+	if len(pairs)%2 != 0 {
+		return ErrInvalidArg
+	}
+	// TODO: Validating the keys.
+	// Validating the values
+	for i := range pairs {
+		_, err := strconv.ParseInt(pairs[i+1], 10, 64)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 // ValidateSizeList validates if the option is a valid size list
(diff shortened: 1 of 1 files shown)


devyanikota:
@kshlm I am calling ValidateSizeList() inside the loop here in ValPriority() to validate a string. Inside that function, splitting at ',' gives us a list with a single element. So we make #len(pairs) number of calls.

2nd: I tried thinking of creating a list of the key's values and then sending it to ValidateSizeList(),
but the creation of that list using Buffer itself leaves us with O(N) complexity.
What do you suggest?

glusterd2/xlator/options/options.go
```diff
@@ -250,8 +250,16 @@ func ValidatePercentOrSize(o *Option, val string) error {
...
 
 // ValidatePriorityOrSize validates either priority or size
+// Example: val := "k1:1024KB, k2:512MB, k3:512GB"
+// It is verified next if 1024KB is a valid size.
 func ValidatePriorityOrSize(o *Option, val string) error {
-	return ErrInvalidArg
+	pairs := strings.Split(val, ",")
+	for _, pair := range pairs {
+		kv := strings.Split(pair, ":")
+		err := ValidateSizeList(o, kv[1])
```
kshlm: Depending on the suffix, you'll need to either validate size or just ensure the value is a number. ValidateSizeList validates that the value is a number, but it also tries to validate that number is a multiple of 512, so it doesn't work.

Just that one addition.

kshlm:
> @kshlm I am calling ValidateSizeList() inside the loop here in ValPriority() to validate a string. Inside that function, splitting at ',' gives us a list with a single element. So we make #len(pairs) number of calls.
>
> 2nd: I tried thinking of creating a list of the key's values and then sending it to ValidateSizeList(),
but the creation of that list using Buffer itself leaves us with O(N) complexity.
What do you suggest?

Don't worry too much about optimization here. Try to keep the code as understandable as possible.

glusterd2/xlator/options/options.go
devyanikota: @kshlm I have added validation checks for both size and number.
And for the second comment, I don't think our efforts to break ValidateSizeList() into two functions(say validateSizeType) would help, since we'll anyway have to check if the string has a 'B' suffix to call ValidateSizeType() in both the caller functions.
Please review if we wish to implement this^^ Thanks.

Issue #434, opened 2017-11-15
NFS-Ganesha integration with GD2


atinmu:
Placeholder github issue to track the NFS-Ganesha integration work required. Further details to be filled in.

kalebskeithley:
As a starting point, I see three modes of operation (three use cases): 1) no NFS, 2) simple NFS, 3) clustered NFS using storhaug.

Note: gNFS is disabled by default when a volume is selected.

If the admin enables NFS and gNFS is enabled, start the glusterfs NFS servers
If the admin enabled NFS and gNFS is disabled, start nfs-ganesha (systemctl start nfs-ganesha). Need support to specifiy which node in the trusted pool to run ganesha on.
If the admin wants clustered nfs, they use the storhaug API to setup and start, and teardown a pacemaker-based HA cluster with nfs-ganesha (and samba)

nixpanic:
Instead of hook scripts, the suggestion has been to handle events. GD2 does not provide events yet, but dbus events could be exposed. In that case, a process can listen for events, create/update the export configuration and inform nfs-ganesha over its dbus interface.

Potentially the event handler for (re)configuring nfs-ganesha can be the same process as the one for Samba (#433)?

aravindavk:
Do we need pre hook or event for Volume Start or Stop(Or any other event)?

Events are asynchronous, so main action will not wait for the response of hooks scripts or events consumers output. For example, Volume start command will not wait for hooks scripts or events consumers response. Volume start command will be successful even if a hook script or events consumer failed. 

If we are using pre scripts currently, is the expectation to stop the action if pre script is not successful?

thotz:
There are two assumptions to be made : 
* Storhaug will be ready and available, so that all the nfs-ganesha ha/export related operations will be handled by it.
* Shared storage will be used to store ganesha related configuration files
 (Please note changes need here does not dependent much storhaug, it effects only the running ganesha process)

Following volume commands need to be considered

- Must one
Stop  -- volume needed to be unexported admin tries to stop it
Start --  if volume is started again after a stop, then it needed to re-exported
For above commands information need to send to running ganesha process via dbus
- Good to have 
Delete -- better to inform that configuration file for that volume can removed from shared storage
Snapshot -- better to save the configuration file used for a volume during snapshot
This information need to given storhaug , may be with help of events framework

- Not very much important 
[...]

Issue #518, opened 2018-01-09
Implement GF_HNDSK_GET_VOLUME_INFO procedure of handshake RPC program


prashanthpai:
Currently glusterd2 only implements `GF_HNDSK_GETSPEC` RPC procedure. There are other RPC procedures that needs to be implemented:

```c
rpcsvc_actor_t gluster_handshake_actors[GF_HNDSK_MAXVALUE] = {
        [GF_HNDSK_NULL]         = {"NULL",        GF_HNDSK_NULL,         NULL,                NULL, 0, DRC_NA},
        [GF_HNDSK_GETSPEC]      = {"GETSPEC",     GF_HNDSK_GETSPEC,      server_getspec,      NULL, 0, DRC_NA},
        [GF_HNDSK_EVENT_NOTIFY] = {"EVENTNOTIFY", GF_HNDSK_EVENT_NOTIFY, server_event_notify, NULL, 0, DRC_NA},
        [GF_HNDSK_GET_VOLUME_INFO] = {"GETVOLUMEINFO", GF_HNDSK_GET_VOLUME_INFO, server_get_volume_info, NULL, 0, DRC_NA},
        [GF_HNDSK_GET_SNAPSHOT_INFO] = {"GETSNAPINFO", GF_HNDSK_GET_SNAPSHOT_INFO, server_get_snap_info, NULL, 0, DRC_NA},
};
```

@thotz was trying to use NFS Ganesha and the gfapi initialisation failed as follows (in gd2 logs):

```log
[2018-01-09 16:37:59.806831]  INFO Remote server has no such procedure: {ProgramNumber:14398633 ProgramVersion:2 ProcedureNumber:6}
``` 

NFS ganesha relies on `GF_HNDSK_GET_VOLUME_INFO` procedure.

prashanthpai:
- [ ] GF_HNDSK_NULL
- [ ] GF_HNDSK_EVENT_NOTIFY
- [ ] GF_HNDSK_GET_VOLUME_INFO

`GF_HNDSK_GET_SNAPSHOT_INFO` shall be implemented separately later.

PR #519, opened 2018-01-10
NFS-Ganesha integration : Sending dbus signals for start and stop

thotz:
If volume is already exported via Ganesha, then during stop the volume
need to be unexport and again re-export during volume start. The old glusterd
does this with help of dbus signals called via wrapper script "dbus-send.sh"
(available in glusterfs 3.10)

TODO :
* need to add validations before calling the script like
	i.) ganesha is running or not
	ii.) volume is exported or not
* Send a patch to include dbus-send.sh in master branch

Required to fix following issue/bug to work ganesha with GD2
* https://bugzilla.redhat.com/show_bug.cgi?id=1532238
* https://github.com/gluster/glusterd2/issues/518

Thanks ppai for all the help

Fixes: #434
Signed-off-by: Jiffin Tony Thottan <jthottan@redhat.com>

### DIFF
--- a/glusterd2/events/nfs-ganesha.go
+++ b/glusterd2/events/nfs-ganesha.go
@@ -1,0 +1,41 @@
+package events
+
+import (
+	"fmt"
+	"os/exec"
+
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	eventVolumeStarted = "volume.started"
+	eventVolumeStopped = "volume.stopped"
+)
+
+type ganesha struct{}
+
+func (g *ganesha) Handle(e *Event) {
+	var option string
+	if e.Name == eventVolumeStarted {
+		option = "on"
+	} else if e.Name == eventVolumeStopped {
+		option = "off"
+	}
+	dbuscmdStr := fmt.Sprintf("/usr/libexec/ganesha/dbus-send.sh /etc/ganesha/ %s %s", option, e.Data["volume.name"])
+	ganeshaCmd := exec.Command("/bin/sh", "-c", dbuscmdStr)
+	err := ganeshaCmd.Run()
+	if err != nil {
+		log.WithError(err).WithField("command", dbuscmdStr).Error("Failed to execute command")
+	} else {
+		log.WithField("command", dbuscmdStr).Debug("Command succeeded")
+	}
+}
+
+func (g *ganesha) Events() []string {
+	return []string{eventVolumeStarted, eventVolumeStopped}
+}
+
+func registerGaneshaHandler() {
+	g := new(ganesha)
+	Register(g)
+}
--- a/glusterd2/events/events.go
+++ b/glusterd2/events/events.go
@@ -50,7 +50,7 @@ func Broadcast(e *Event) error {
 func Start() error {
 	StartGlobal()
 	startEventLogger()
-
+	registerGaneshaHandler()
 	return nil
 }
 


prashanthpai:
add to whitelist

prashanthpai:
@thotz Please rebase against master (contains the RPC fix) and fix the linting issue (check the CI test output)

glusterd2/events/nfs-ganesha.go
```diff
@@ -0,0 +1,37 @@
+package events
+
+import (
+	log "github.com/sirupsen/logrus"
```
prashanthpai: The general convention followed for imports is as follows:
```
standard golang packages

gd2 packages

third-party packages
```

In this case:
```go
import (
    "fmt"
    "os/exec"

)
```

glusterd2/events/nfs-ganesha.go
```diff
@@ -0,0 +1,37 @@
...
+		err := ganeshaCmd.Run()
+		if err != nil {
+			log.Warn(dbuscmdStr, " got failed with ", err)
+		} else {
+			log.Debug(dbuscmdStr, " got succeed")
+		}
+}
+
+func (g *ganesha) Events() []string {
+	        return []string{"volume.started", "volume.stopped"}
```
prashanthpai: As these are used at multiple places in the same file, you can simply define them as constants instead:

```go
const (
    eventVolumeStarted = "volume.started"
    eventVolumeStopped = "volume.stopped"
)
```
You can reuse the constants in `Events()` and in `Handle()`

In the future, we may have these events exported from a common package.

glusterd2/events/nfs-ganesha.go
```diff
@@ -0,0 +1,37 @@
...
+		if (e.Name == "volume.started") {
+			option = "on"
+		} else if (e.Name == "volume.stopped") {
+			option = "off"
+		}
+		dbuscmdStr := fmt.Sprintf("/usr/libexec/ganesha/dbus-send.sh /etc/ganesha/ %s %s", option, e.Data["volume.name"])
+		ganeshaCmd := exec.Command("/bin/sh", "-c", dbuscmdStr)
+		err := ganeshaCmd.Run()
+		if err != nil {
+			log.Warn(dbuscmdStr, " got failed with ", err)
```
prashanthpai: You can log as follows:
```go
log.WithError(err).WithField("command", dbuscmdStr).Error("Failed to execute command")
...
log.WithField("command", dbuscmdStr).Debug("Command succeeded")
```

glusterd2/events/nfs-ganesha.go
thotz: Done

glusterd2/events/nfs-ganesha.go
thotz: Done

glusterd2/events/nfs-ganesha.go
thotz: Done

thotz:
What does GD1 with nfs-ganesha ?
==========================
-- Setting up HA, creating/managing conf files for ganesha and volumes, exporting/unexporting volume
-- Uses shared storage for saving configuration files
-- Also while taking snapshot, it saves the export related conf files
Most taken care by gluster commands like start, stop, snapshot, set, reset, ganesha specfic(nfs-ganesha and ganesha.enable)

NFS-Ganesha integration with GD2
============================
Storhaug does the following
-- sets up ganesha cluster
-- export/unexport the volume
-- uses shared storage for storing configuration file

Current plans for integration
-----------------------------
Approach 1 : just make storhaug will take care of everything.

GD2 will handle following scenarios with help of running ganesha process/shared storage
1.) for start/stop : sends the dbus signal to local running ganesha process with help of shared storage
2.) enabling gNFS(if it is supported with GD2??) : ideally it should fail if there running ganesha process, again we use shared storage to ensure it
3.) Snapshots : saves the configuration file while taking snapshot and restores the ganesha conf whenever needed from shared storage.

Approach 2 :
Retains everything like GD1 does, but GD2 calls storhaug cli's instead of calling helper scripts

Enhancements can be done for new approach (in storhaug and semantics to cli)
----------------------------------------------------------------------------------------------------------
1.) Ganesha related
-- have different modes : HA, without HA, pnfs, remote

2.) Export related
-- more user friendly approach for exporting multiple subdirs for a single volume

3.) Configuration related
-- creates directory for each volume, save configurataion file as directory specific
-- change the dependency on share storage() 

4.) exporting snapshots(like mounting snapshots via fuse)

5.) add options for managing ganesha cluster as well in GD2

6.) fetching client related details from ganesha cluster

PR #520, opened 2018-01-11
doc: quick start guide updated with new volume request format

aravindavk:


aravindavk:
This is not required in quick start guide if #523 merges.

aravindavk:
@prashanthpai please review this patch if the patch #523 is not required.

Issue #521, opened 2018-01-11
Add support for bricks list without explicit subvols for Volume Create


aravindavk:
With the introduction of subvols, Volume Create request becomes lengthy for plain distribute Volumes and for Volumes with homogeneous sub volumes.

Volume Request API can also accept just list of bricks as before and split into sub volumes before storing in etcd as Volinfo.

For example, following request with list of bricks will be converted as list of subvols

    {
                "name": "testvol",
                "replica" : 2,
                "bricks": [
            	    "<uuid1>:/export/brick1/data",
            	    "<uuid2>:/export/brick2/data",
            	    "<uuid1>:/export/brick3/data",
            	    "<uuid2>:/export/brick4/data"
                ],
                "force": true
    }

Converted Volume request

    {
            "name": "testvol",
            "subvols": [
                {
                    "type": "replicate",
                    "bricks": [
                        {"nodeid": "<uuid1>", "path": "/export/brick1/data"},
                        {"nodeid": "<uuid2>", "path": "/export/brick2/data"}
                    ],
                    "replica": 2
                },
                {
                    "type": "replicate",
                    "bricks": [
                        {"nodeid": "<uuid1>", "path": "/export/brick3/data"},
                        {"nodeid": "<uuid2>", "path": "/export/brick4/data"}
                    ],
                    "replica": 2
                }
            ],
            "force": true
    }

kshlm:
Why!!
I'd rather not support this. Let's leave the old format juggling to CLI. Anyone who uses to use the ReST API, should use the new format. There is no need to support any sort of backwards compatibility here, as any user of this API will be a new user.

aravindavk:
> I'd rather not support this. Let's leave the old format juggling to CLI. Anyone who uses to use the ReST API, should use the new format. There is no need to support any sort of backwards compatibility here, as any user of this API will be a new user.

This is not for backward compatibility, added this for Convenience. API request format is repeatitive if  a Volume has only one subvolume or homogeneous subvolumes.

prashanthpai:
When the overall directive is to move away from normal users (not power users) having to understand terminologies such as `bricks`, understanding `subvol` terminology is additional burden IMO. However, I'd still want users to be able to specify replica/redundancy sets clearly.

atinmu:
I kind of agree with Aravinda. Having user to have an easier way to
construct the API request would add value from an user perspective.
Thoughts?

aravindavk:
[...]

PR #523, opened 2018-01-12
volume-create: Add support of old format with just list of bricks

aravindavk:
Fixes: #521

### DIFF
--- a/glusterd2/commands/volumes/volume-create.go
+++ b/glusterd2/commands/volumes/volume-create.go
@@ -26,6 +26,14 @@ func unmarshalVolCreateRequest(msg *api.VolCreateReq, r *http.Request) (int, err
 		return http.StatusBadRequest, gderrors.ErrEmptyVolName
 	}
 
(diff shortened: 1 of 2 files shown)


prashanthpai:
My thoughts on this:
* The API shouldn't be implicit when it comes to choosing bricks and replica sets. Being explicit is better and dis-ambiguous. Further, having a single format for core volume create API is better.
* However, users now will have to understand the term `Subvol` which is slight additional burden IMO.
* The API should also allow for a brick to be explicitly marked/tagged as arbiter.

I'm cleaning up the PR review backlog and I'm inclining towards abandoning this based on @kshlm's inputs [here](https://github.com/gluster/glusterd2/issues/521#issuecomment-357215213) and [here](https://botbot.me/freenode/gluster-dev/search/?q=523) unless @aravindavk @amarts @atinmu have other thoughts ?

aravindavk:
Sure, we can work on this API if required in future. I will close this issue.

PR #526, opened 2018-01-14
Heal info - granular entry optio - review

vpandey-RH:
**DO NOT MERGE** 

This PR is just to retrieve reviews on code quality or any other issues. It deals with heal info APIs and granular heal option enable/disable APIs.
This PR can only be considered for merging only once [PR#471]( https://github.com/gluster/glusterd2/pull/471) is ready until then reviewers are requested for suggestions to improve the code.


vpandey-RH:
add to whitelist.

atinmu:
@vpandey-RH What's the update on this PR?

vpandey-RH:
@atinmu  Waiting for reviews.

prashanthpai:
@vpandey-RH Please rebase to master and cleanup the commits from unintended changes

vpandey-RH:
@prashanthpai  Working on that.

vpandey-RH:
@prashanthpai  Changed the SocketFile() method in glustershd.go file because the heal index and  heal full API includes sending a RPC call to glustershd and the GetRPCClient() calls SocketFile() method of a daemon.

glusterd2/volume/struct.go
```diff
@@ -96,19 +96,20 @@ type Subvol struct {
...
+	DistCount                int
+	Options                  map[string]string
+	State                    VolState
+	Checksum                 uint64
+	Version                  uint64
+	Subvols                  []Subvol
+	Auth                     VolAuth // TODO: should not be returned to client
+	GraphMap                 map[string]string
+	HealEnabled              bool
+	GranularHealEntryEnabled bool
```
prashanthpai: Why can't this be a volume set ?

vpandey-RH: Because before setting the option I need to run glfsheal binary with granular-entry-heal-op first and check if there are any heals pending.  I don't think volume set has the support for this.

vpandey-RH:
@prashanthpai  @aravindavk  Please review the PR.

vpandey-RH:
@atinmu  @prashanthpai @aravindavk The tests for now will fail because the glfsheal binary is looking for socket file with path DATADIR + '/run/glusterd2.socket' as mentioned in glusterfs.h . While running tests, the testsuite make its own glusterd2.socket in WORKDIR + '/run/glusterd2.socket' , so that's glfsheal binary fails to find glusterd2.socket

atinmu:
> The tests for now will fail because the glfsheal binary is looking for socket file with path DATADIR + '/run/glusterd2.socket' as mentioned in glusterfs.h . While running tests, the testsuite make its own glusterd2.socket in WORKDIR + '/run/glusterd2.socket' , so that's glfsheal binary fails to find glusterd2.socket. One of the ways for overcoming this is to add an extra argument in glfsheal binary(glfs-heal.c) to take path of the socket file as well.

So this basically means that DATADIR and WORKDIR paths are different. Is this true even with existing GD1 and .t runs? If so how that's handled? @aravindavk @prashanthpai

aravindavk:
Fix is required to glfsheal binary, it should accept `--glusterd-sock` similar to Glusterd1 Gluster CLI.

@vpandey-RH Please work on the `glusterfs` patch first, and pass that flag while calling from Glusterd2

vpandey-RH:
@aravindavk Sure, will work on that. Can anyone open an issue on bugzilla regarding this ?

prashanthpai:
retest this please

prashanthpai:
@vpandey-RH  Wait till your [dependent change](https://review.gluster.org/#/c/19745/) ends up in glusterfs nightly RPM.

plugins/glustershd/glustershd.go
```diff
@@ -62,7 +60,12 @@ func (shd *Glustershd) Args() string {
 
 // SocketFile returns path to the socket file used for IPC.
 func (shd *Glustershd) SocketFile() string {
-	return ""
+	if shd.socketfilepath != "" {
```
prashanthpai: `SocketFile()` is for location for socket file that self-heal daemon **listens on**. Is self-heal daemon listening on a socket file ? Does glusterd2 communicates with self heal daemon over socket file.

I think you're getting confused between listening on socket file and connecting to glusrerd's socket file.

plugins/glustershd/init.go
```diff
@@ -35,6 +35,30 @@ func (p *Plugin) RestRoutes() route.Routes {
 			Pattern:     "/volumes/{name}/heal/disable",
 			Version:     1,
 			HandlerFunc: glustershDisableHandler},
+		route.Route{
+			Name:        "GlustershInfo",
+			Method:      "GET",
+			Pattern:     "/volumes/{name}/{opts}/heal-info",
+			Version:     1,
```
prashanthpai: Please fill request and response struct fields for these rest routes wherever applicable. See [this issue](https://github.com/gluster/glusterd2/issues/584) for example and reference.

plugins/glustershd/init.go
```diff
@@ -44,4 +68,7 @@ func (p *Plugin) RegisterStepFuncs() {
 	transaction.RegisterStepFunc(txnSelfHealStart, "selfheal-start")
 	transaction.RegisterStepFunc(txnSelfHealdUndo, "selfheald-undo")
 	transaction.RegisterStepFunc(txnSelfHealStop, "selfheal-stop")
+	transaction.RegisterStepFunc(txnHealInfo, "heal-info.Commit")
```
prashanthpai: Lose the `Commit` suffix.

plugins/glustershd/init.go
```diff
@@ -44,4 +68,7 @@ func (p *Plugin) RegisterStepFuncs() {
...
+	transaction.RegisterStepFunc(txnGranularEntryHealEnable, "granular-heal.Enable")
```
prashanthpai: Have all heal related step functions start with the same prefix - `selfheal-`

plugins/glustershd/init.go
```diff
@@ -35,6 +35,30 @@ func (p *Plugin) RestRoutes() route.Routes {
 			Pattern:     "/volumes/{name}/heal/disable",
 			Version:     1,
 			HandlerFunc: glustershDisableHandler},
+		route.Route{
+			Name:        "GlustershInfo",
```
prashanthpai: Lost the `Gluster` prefix. Just name it `SelfHealInfo` Same comment elsewhere

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
 
 	restutils.SendHTTPResponse(ctx, w, http.StatusOK, nil)
 }
+
+func glustershInfoHandler(w http.ResponseWriter, r *http.Request) {
```
prashanthpai: The function name is misleading. Lose the gluster prefix and include reference to selfheal in the name.

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+	p := mux.Vars(r)
+	volname := p["name"]
+	option := p["opts"]
+	_ = option
```
prashanthpai: Why ? If you're not using a variable, you should remove it.

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	ctx := r.Context()
+	logger := gdctx.GetReqLogger(ctx)
+
+	//validate volume name
```
prashanthpai: All these validations should be in a step function and be done under a lock.

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+		lock,
+		{
+			DoFunc: "heal-info.Commit",
+			Nodes:  []uuid.UUID{gdctx.MyUUID},
+		},
+		unlock,
+	}
+
+	if err := txn.Ctx.Set("volname", volname); err != nil {
+		logger.WithError(err).Error("failed to set volname in transaction context")
```
prashanthpai: Keep the error message generic.

```go
logger.WithError(err).WithField("key", "volname").Error("failed to set key in transaction context")
```

Follow the same everywhere else.

plugins/glustershd/init.go
```diff
@@ -35,6 +35,30 @@ func (p *Plugin) RestRoutes() route.Routes {
...
+			Method:      "GET",
+			Pattern:     "/volumes/{name}/{opts}/heal-info",
```
prashanthpai: What are possible values for `opts` here ?

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+
+	var options []string
+
+	glusterdSockpath := path.Join(config.GetString("rundir"), "glusterd2.socket")
+	options = append(options, option, "xml", "glusterd-sock", glusterdSockpath)
+	if err := txn.Ctx.Set("option", options); err != nil {
+		logger.WithError(err).Error("failed to set volname in transaction context")
```
prashanthpai: Same comment here about the log entry.

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+
+	err = txn.Do()
+	if err != nil {
+		logger.WithFields(log.Fields{
```
prashanthpai: Use `WithError` instead of putting in error as one of the fields.

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+			"error":   err.Error(),
+			"volname": volname,
+		}).Error("failed to retrieve heal info")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+	var out string
+	err = txn.Ctx.GetNodeResult(gdctx.MyUUID, "stdout", &out)
+	if err != nil {
+		logger.WithFields(log.Fields{
```
prashanthpai: Ditto.

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+			"error": err.Error(),
+		}).Error("Could Not Fetch Stdout", err)
+		return
```
prashanthpai: This will hang forever! Return a HTTP error to the client.

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		logger.WithFields(log.Fields{
+			"error":   err.Error(),
+			"volname": volname,
+		}).Error("failed to retrieve heal info")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+	var out string
+	err = txn.Ctx.GetNodeResult(gdctx.MyUUID, "stdout", &out)
```
prashanthpai: If the stepfunction is run only on this node. Simply use the normal `Get` function.

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+	output := []byte(out)
+
+	var info HealInfo
+	err = xml.Unmarshal(output, &info)
+	if err != nil {
+
+		logger.WithFields(log.Fields{
+			"error": err.Error(),
+		}).Error("Error unmarshalling from XML", err)
+		return
```
prashanthpai: This will hang forever! Return a HTTP error to the client.

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+	output := []byte(out)
+
+	var info HealInfo
+	err = xml.Unmarshal(output, &info)
+	if err != nil {
+
+		logger.WithFields(log.Fields{
+			"error": err.Error(),
```
prashanthpai: Ditto. Same comment about using `WithError`

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+}
+
+func granularHealEnableHandler(w http.ResponseWriter, r *http.Request) {
+	p := mux.Vars(r)
+	volname := p["name"]
+
+	ctx := r.Context()
+	logger := gdctx.GetReqLogger(ctx)
+
+	//validate volume name
```
prashanthpai: Validations must happen in a step function under a lock. So does all read-modify-write operations on the store.

plugins/glustershd/transactions.go
```diff
@@ -69,3 +75,65 @@ func txnSelfHealdUndo(c transaction.TxnCtx) error {
...
+	}
+
+	c.SetNodeResult(gdctx.MyUUID, "stdout", out.String())
+
+	return nil
+}
+
+func txnHealInfo(c transaction.TxnCtx) error {
+	var volname string
+	var option []string
```
prashanthpai: As this is a slice/array, name this as `options` - or `args` shall be better.

plugins/glustershd/transactions.go
```diff
@@ -69,3 +75,65 @@ func txnSelfHealdUndo(c transaction.TxnCtx) error {
...
+	cmd := exec.Command(path, args...)
+	cmd.Stdout = &out
+	err := cmd.Run()
+	if err != nil {
+		fmt.Println("Failed Running Glfsheal Binary")
+
+		return err
+	}
+
+	c.SetNodeResult(gdctx.MyUUID, "stdout", out.String())
```
prashanthpai: As it's run on a single node, simply use the normal `Set`

plugins/glustershd/transactions.go
```diff
@@ -69,3 +75,65 @@ func txnSelfHealdUndo(c transaction.TxnCtx) error {
...
+		fmt.Println("Couldn't Find glfsheal binary")
+
+		return e
+	}
+
+	cmd := exec.Command(path, args...)
+	cmd.Stdout = &out
+	err := cmd.Run()
+	if err != nil {
+		fmt.Println("Failed Running Glfsheal Binary")
```
prashanthpai: Never use `Println`. Log instead. Fix this elsewhere too.

plugins/glustershd/transactions.go
```diff
@@ -69,3 +75,65 @@ func txnSelfHealdUndo(c transaction.TxnCtx) error {
...
+		buffer.WriteString(fmt.Sprintf(" %s", opt))
+	}
+
+	a := buffer.String()
+	args := strings.Fields(a)
+	path, e := exec.LookPath("glfsheal")
+	if e != nil {
+		fmt.Println("Couldn't Find glfsheal binary")
+
+		return e
```
prashanthpai: Consistently use `err` as variable name.

plugins/glustershd/transactions.go
```diff
@@ -69,3 +75,65 @@ func txnSelfHealdUndo(c transaction.TxnCtx) error {
...
+	buffer.WriteString(fmt.Sprintf("%s", volname))
+	for _, opt := range option {
+		buffer.WriteString(fmt.Sprintf(" %s", opt))
+	}
+
+	a := buffer.String()
+	args := strings.Fields(a)
+	path, e := exec.LookPath("glfsheal")
+	if e != nil {
+		fmt.Println("Couldn't Find glfsheal binary")
```
prashanthpai: Never use `Println`. Log instead. Fix this elsewhere too.

plugins/glustershd/transactions.go
```diff
@@ -69,3 +75,65 @@ func txnSelfHealdUndo(c transaction.TxnCtx) error {
...
+func runGlfshealBin(c transaction.TxnCtx, volname string, option []string) error {
+	var out bytes.Buffer
+	var buffer bytes.Buffer
+
+	buffer.WriteString(fmt.Sprintf("%s", volname))
+	for _, opt := range option {
+		buffer.WriteString(fmt.Sprintf(" %s", opt))
+	}
+
+	a := buffer.String()
```
prashanthpai: You can use it directly as shown below:

```go
args := strings.Fields(buffer.String())
```

plugins/glustershd/struct.go
```diff
@@ -0,0 +1,23 @@
+package glustershd
+
+import (
+	"encoding/xml"
+)
+
+// Brick represents brick details for Heal Info command
+type Brick struct {
+	HostID                         string `xml:"hostUuid,attr"json:"hostId"`
```
prashanthpai: Same comment here about naming json strings. And about variable names.

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+		},
+		{
+			DoFunc: "vol-option.UpdateVolinfo",
+			Nodes:  txn.Nodes,
+		},
+		unlock,
+	}
+
+	if err := txn.Ctx.Set("volinfo", v); err != nil {
+		logger.WithError(err).Error("failed to set volinfo in transaction context")
```
prashanthpai: Use a generic error message and use log fields to add into about the keys.

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+
+	if err := txn.Ctx.Set("oldvolinfo", oldvolinfo); err != nil {
+		logger.WithError(err).Error("failed to set oldvolinfo in transaction context")
```
prashanthpai: Use a generic error message and use log fields to add into about the keys.

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+
+	if err := txn.Ctx.Set("volname", volname); err != nil {
+		logger.WithError(err).Error("failed to set volname in transaction context")
```
prashanthpai: Use a generic error message and use log fields to add into about the keys.

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+
+	var options []string
+	glusterdSockpath := path.Join(config.GetString("rundir"), "glusterd2.socket")
+	options = append(options, "granular-entry-heal-op", "glusterd-sock", glusterdSockpath)
+	if err := txn.Ctx.Set("option", options); err != nil {
+		logger.WithError(err).Error("failed to set volname in transaction context")
```
prashanthpai: Use a generic error message and use log fields to add into about the keys.

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+
+	err = txn.Do()
+	if err != nil {
+		logger.WithFields(log.Fields{
+			"error":   err.Error(),
```
prashanthpai: Same comment here.

plugins/glustershd/rest.go
```diff
@@ -187,3 +190,280 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+}
+
+func granularHealDisableHandler(w http.ResponseWriter, r *http.Request) {
+	p := mux.Vars(r)
+	volname := p["name"]
+
+	ctx := r.Context()
+	logger := gdctx.GetReqLogger(ctx)
+
+	//validate volume name
```
prashanthpai: Same comment here - do most of these validations under lock.

pkg/api/healinfo_resp.go
```diff
@@ -0,0 +1,13 @@
+package api
+
+// BrickHealInfo represents brick details for Heal Info command
+type BrickHealInfo struct {
+	HostID                         string `xml:"hostUuid,attr"json:"hostId"`
```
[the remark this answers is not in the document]
vpandey-RH: Yes, I need xml for this purpose only.

plugins/glustershd/rest.go
vpandey-RH: Done.

plugins/glustershd/rest.go
vpandey-RH: Leaving this out for now as discussed.

prashanthpai: No, not this one. I was referring to removing contents of SocketFile().

e2e/glustershd_test.go
```diff
@@ -13,7 +13,7 @@ func TestGlusterShd(t *testing.T) {
 
 	r := require.New(t)
 
-	gds, err := setupCluster("./config/1.toml", "./config/2.toml")
+	gds, err := setupCluster("./config/1.toml", "./config/3.toml")
```
prashanthpai: Unintended change.

plugins/glustershd/glustershd.go
prashanthpai: Move this code from SocketFile() to Args()

plugins/glustershd/rest.go
```diff
@@ -1,17 +1,22 @@
...
 	"net/http"
+	"path"
 
 	"github.com/gluster/glusterd2/glusterd2/gdctx"
 	restutils "github.com/gluster/glusterd2/glusterd2/servers/rest/utils"
 	"github.com/gluster/glusterd2/glusterd2/transaction"
 	"github.com/gluster/glusterd2/glusterd2/volume"
 	"github.com/gluster/glusterd2/pkg/errors"
 
+	glustershdapi "github.com/gluster/glusterd2/plugins/glustershd/api"
```
prashanthpai: This import should be part of the above block as it's also an internal glusterd2 package.

plugins/glustershd/rest.go
prashanthpai: Move this and checks below into a step function. You will lose the ability to return `http.StatusNotFound` (404) but that is okay. The stepfunction can return error if volume not found.

plugins/glustershd/rest.go
```diff
@@ -187,3 +192,261 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+	err = txn.Do()
+	if err != nil {
+		logger.WithError(err).Error("Failed to retrieve heal info.")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+	var out string
+	err = txn.Ctx.Get("stdout", &out)
+	if err != nil {
+		logger.WithError(err).Error("Failed to retrieve heal info.")
```
prashanthpai: Use generic error message i.e `failed to get key from transaction context` and use `WithField`

plugins/glustershd/rest.go
vpandey-RH: But we discussed that for now I should leave the validations part as it is and not include them in a step function.

plugins/glustershd/rest.go
prashanthpai: This comment still stands.

plugins/glustershd/rest.go
```diff
@@ -187,3 +192,261 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+	defer txn.Cleanup()
+
+	//Lock on Volume Name
+	lock, unlock, err := transaction.CreateLockSteps(volname)
+	if err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+
+	v.HealOptions.GranularHealEntryEnabled = true
```
prashanthpai: This should also be in a step function

plugins/glustershd/rest.go
```diff
@@ -187,3 +192,261 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+	txn.Steps = []*transaction.Step{
+		lock,
+		{
+			DoFunc:   "granular-heal.Enable",
+			Nodes:    []uuid.UUID{gdctx.MyUUID},
+			UndoFunc: "granular-heal.Undo",
+		},
+		{
+			DoFunc: "vol-option.UpdateVolinfo",
+			Nodes:  txn.Nodes,
```
prashanthpai: It is sufficient to run this on a single node.

plugins/glustershd/rest.go
```diff
@@ -187,3 +192,261 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+		},
+		unlock,
+	}
+
+	if err := txn.Ctx.Set("volinfo", v); err != nil {
```
prashanthpai: All these transaction set should be done in a stepfunc.

plugins/glustershd/transactions.go
```diff
@@ -69,3 +76,69 @@ func txnSelfHealdUndo(c transaction.TxnCtx) error {
...
+	}
+
+	cmd := exec.Command(path, args...)
+	cmd.Stdout = &out
+	err = cmd.Run()
+	if err != nil {
+		c.Logger().WithError(err).Error("Failed to execute Glfsheal Binary")
+		return err
+	}
+	c.Set("stdout", out.String())
```
prashanthpai: Handle error.

plugins/glustershd/rest.go
prashanthpai: @vpandey-RH It can be moved. There's no technical reasons or dependency to not to move. The only problem with moving is that the stepfunction cannot specify what HTTP status to return to client. I am okay with that.

vpandey-RH:
retest this please

vpandey-RH:
retest this please

vpandey-RH:
@prashanthpai I have limited this API to handle just healinfo and will be uploading a patch for granular heal entry after this PR gets merged. Also, did some improvements based on your reviews. Kindly review it.

pkg/api/healinfo_resp.go
```diff
@@ -0,0 +1,13 @@
+package api
```
prashanthpai: This must be part of `pkg/glustershd/api` package.

plugins/glustershd/init.go
```diff
@@ -35,13 +37,38 @@ func (p *Plugin) RestRoutes() route.Routes {
...
 // Glusterd Transaction framework
 func (p *Plugin) RegisterStepFuncs() {
-	transaction.RegisterStepFunc(txnSelfHealStart, "selfheal-start")
-	transaction.RegisterStepFunc(txnSelfHealdUndo, "selfheald-undo")
-	transaction.RegisterStepFunc(txnSelfHealStop, "selfheal-stop")
+	var sfs = []struct {
+		sf   transaction.StepFunc
+		name string
+	}{
+		{txnSelfHealStart, "selfheal-start"},
```
prashanthpai: The string should come first and them the functions, for better readability.

pkg/api/healinfo_resp.go
```diff
@@ -0,0 +1,13 @@
...
+
```
prashanthpai: The file `glustercli.log` shouldn't be part of the PR.

plugins/glustershd/rest.go
```diff
@@ -113,64 +110,68 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
 		{
 			DoFunc: "vol-option.NotifyVolfileChange",
 			Nodes:  txn.Nodes,
 		},
-		unlock,
 	}
 
-	if err := txn.Ctx.Set("volinfo", v); err != nil {
-		logger.WithError(err).Error("failed to set volinfo in transaction context")
+	if err := txn.Ctx.Set("volfile-property", int(HealEnabled)); err != nil {
```
prashanthpai: What is this for ?

plugins/glustershd/rest.go
```diff
@@ -113,64 +110,68 @@ func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+		logger.WithError(err).WithField("key", "volfile-property").Error("failed to set key in transaction context")
 		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
 		return
 	}
 
-	if err := txn.Ctx.Set("oldvolinfo", oldvolinfo); err != nil {
-		logger.WithError(err).Error("failed to set volinfo in transaction context")
+	if err := txn.Ctx.Set("volfile-property-value", false); err != nil {
```
prashanthpai: And this ?

plugins/glustershd/struct.go
```diff
@@ -0,0 +1,10 @@
+package glustershd
+
+// VolfileOptions is a brick operation
+//go:generate stringer -type=VolfileOptions
+type VolfileOptions uint32
+
+// List of volfile properties that needs to be changed
+const (
+	HealEnabled VolfileOptions = iota // SelfHeal Enable
```
prashanthpai: How and where is this used ?

vpandey-RH:
@aravindavk  Kindly review the PR.

plugins/glustershd/actor.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return err
+	}
+	switch value {
+	case "on":
+		if isHealEnabled(v, key) {
+			err = daemon.Start(glustershDaemon, true)
+			if err == gderrors.ErrProcessAlreadyRunning {
+				return nil
+			}
```
prashanthpai: You should check for nil too. And also return error if the error is other than `gderrors.ErrProcessAlreadyRunning`

plugins/glustershd/actor.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/gluster/glusterd2/glusterd2/xlator"
+	gderrors "github.com/gluster/glusterd2/pkg/errors"
+)
+
+var names = [...]string{"replicate", "afr"}
+
+type shdActor struct{}
+
+func (actor *shdActor) Do(v *volume.Volinfo, key string, value string) error {
+	// validate volume type
```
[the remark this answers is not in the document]
prashanthpai: Wait, you already have this check in `txnValidateVolNameType`

glusterd2/commands/volumes/volume-start.go
```diff
@@ -33,7 +33,6 @@ func startAllBricks(c transaction.TxnCtx) error {
 			return err
 		}
 	}
-
```
prashanthpai: Unintended changes. Try to avoid them

plugins/glustershd/init.go
```diff
@@ -18,24 +20,34 @@ func (p *Plugin) Name() string {
...
 func (p *Plugin) RegisterStepFuncs() {
-	transaction.RegisterStepFunc(txnSelfHealStart, "selfheal-start")
-	transaction.RegisterStepFunc(txnSelfHealdUndo, "selfheald-undo")
-	transaction.RegisterStepFunc(txnSelfHealStop, "selfheal-stop")
+	var sfs = []struct {
+		sf   transaction.StepFunc
+		name string
+	}{
+		{txnHealInfo, "selfheal-info"},
+		{txnValidateVolNameType, "validate-vol-name-type"},
```
prashanthpai: Both validate steps can be merged into one validate step.

Also, prefix step names with `selfheal`

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,95 @@
...
-
-	return false
-}
-
-func glustershEnableHandler(w http.ResponseWriter, r *http.Request) {
-	// Implement the help logic and send response back as below
+func selfhealInfoHandler(w http.ResponseWriter, r *http.Request) {
 	p := mux.Vars(r)
 	volname := p["name"]
+	option := p["opts"]
```
prashanthpai: Where and how is this used ?

I see that you have defined two REST APIs, both of them use the same handler (which is fine).

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,95 @@
...
 
 	ctx := r.Context()
 	logger := gdctx.GetReqLogger(ctx)
 
-	//validate volume name
-	v, err := volume.GetVolume(volname)
-	if err != nil {
-		if err == errors.ErrVolNotFound {
-			restutils.SendHTTPError(ctx, w, http.StatusNotFound, errors.ErrVolNotFound)
+	lock, unlock := transaction.CreateLockFuncs(volname)
```
prashanthpai: If you are taking locks outside, you can do validations in the handler itself instead of step function

plugins/glustershd/transactions.go
```diff
@@ -1,71 +1,132 @@
...
+	var volname string
+
+	if err := c.Get("volname", &volname); err != nil {
 		c.Logger().WithError(err).WithField(
-			"volume", oldvolinfo.Name).Debug("storeVolume: failed to store volume info")
+			"key", "volname").Debug("Failed to get key from transaction context.")
 		return err
 	}
-	if err := volgen.Generate(); err != nil {
+	// Validate volume existence
```
prashanthpai: These need be step functions. You can simply do these validations in the handler right after taking locks.

plugins/glustershd/init.go
prashanthpai: Actually, validations can happen in handler after taking lock.

plugins/glustershd/actor.go
```diff
@@ -0,0 +1,88 @@
...
+
+var names = [...]string{"replicate", "afr"}
+
+type shdActor struct{}
+
+func (actor *shdActor) Do(v *volume.Volinfo, key string, value string) error {
+	glustershDaemon, err := newGlustershd()
+	if err != nil {
+		return err
+	}
```
aravindavk: This actor will be called for all `replicate.*` or `afr.*` options. Add condition to return nil if key is not selfHealKey

plugins/glustershd/actor.go
```diff
@@ -0,0 +1,88 @@
...
+	switch value {
+	case "on":
+		if afr.IsHealEnabled(v) {
```
aravindavk: Is this condition required? value is "on"

plugins/glustershd/actor.go
```diff
@@ -0,0 +1,88 @@
...
+                        err = daemon.Start(glustershDaemon, true)
+			if err == gderrors.ErrProcessAlreadyRunning {
+				return nil
+			} else if err != nil {
+                                return err
+                        }
+		}
+	case "off":
+		if !afr.IsHealEnabled(v) {
```
aravindavk: Same here, value is "off"

plugins/glustershd/actor.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (actor *shdActor) Undo(v *volume.Volinfo, key string, value string) error {
+	glustershDaemon, err := newGlustershd()
+	if err != nil {
+		return err
+	}
+	switch value {
+	case "off":
+		if !afr.IsHealEnabled(v) {
```
aravindavk: Same here, value is "off"

plugins/glustershd/actor.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+	}
+	return nil
+}
+
+func (actor *shdActor) Undo(v *volume.Volinfo, key string, value string) error {
+	glustershDaemon, err := newGlustershd()
+	if err != nil {
+		return err
+	}
```
aravindavk: Add condition to check the key

plugins/glustershd/actor.go
vpandey-RH: Just an additional check to ensure that "on" has been set in volume options. Only when "on" is set in volume options then proceed to start the daemon.

glustercli/cmd/glustershd.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/spf13/cobra"
+)
+
+const (
+	helpSelfHealCmd     = "Gluster Self Heal"
+	helpSelfHealInfoCmd = "Self Heal Info"
+)
+
+const (
+	infoSummary    = "info-summary"
```
prashanthpai: You can put these constants in the same const block. Even better if you get rid of them use it directly.

plugins/afr/validate.go
```diff
@@ -9,12 +9,39 @@ import (
 
 var names = [...]string{"replicate", "afr"}
 
+var shdEnabledKey = [...]string{"afr.self-heal-daemon", "replicate.self-heal-daemon"}
+
+// IsVolReplicate return true if volume is of type replicate, disperse, distreplicate or distdisperse
```
prashanthpai: For now, duplicate (and unexport) this code in each package where it's used. In future we'll merge afr and selfheal packages.

plugins/afr/validate.go
```diff
@@ -9,12 +9,39 @@ import (
...
+func IsVolReplicate(vType volume.VolType) bool {
+	if vType == volume.Replicate || vType == volume.Disperse || vType == volume.DistReplicate || vType == volume.DistDisperse {
+		return true
+	}
+
+	return false
+}
+
+// IsHealEnabled returns true if heal is enabled for the volume otherwise returns false.
+func IsHealEnabled(v *volume.Volinfo) bool {
```
prashanthpai: For now, duplicate (and unexport) this code in each package where it's used. In future we'll merge afr and selfheal packages.

plugins/glustershd/actor.go
```diff
@@ -0,0 +1,102 @@
...
+
+const(
+        selfHealKey = "self-heal-daemon"
+)
+
+type shdActor struct{}
+
+func (actor *shdActor) Do(v *volume.Volinfo, key string, value string) error {
+        glustershDaemon, err := newGlustershd()
+	if err != nil {
```
prashanthpai: Something is off about these indentations. Hmm

plugins/glustershd/actor.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+        if key != selfHealKey {
+                return nil
+        }
+
+	switch value {
+	case "on":
+		if afr.IsHealEnabled(v) {
+                        err = daemon.Start(glustershDaemon, true)
```
prashanthpai:
```go
err = daemon.Start(glustershDaemon, true)
if err != nil || err != gderrors.ErrProcessAlreadyRunning {
    return err
}
```

plugins/glustershd/actor.go
```diff
@@ -0,0 +1,102 @@
...
+                        }
+		}
+	case "on":
+		if afr.IsHealEnabled(v) {
+			isVolRunning, err := volume.AreReplicateVolumesRunning()
+			if err != nil {
+				return err
+			}
+			if !isVolRunning {
+				err = daemon.Stop(glustershDaemon, true)
```
prashanthpai: Just do ```return daemon.Stop(glustershDaemon, true)```

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,122 @@
...
-		},
-		{
-			DoFunc: "selfheal-start",
-			Nodes:  txn.Nodes,
-		},
 		{
-			DoFunc: "vol-option.NotifyVolfileChange",
-			Nodes:  txn.Nodes,
+			DoFunc: "selfheal-info",
+			Nodes:  []uuid.UUID{gdctx.MyUUID},
```
prashanthpai: Why is this run on just this node ?

If it's indeed supposed to run only on this node, you don't need a transaction.

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,129 @@
...
+func getHealInfo(volname string, args []string) error {
+	return runGlfshealBin(volname, args)
 }
 
-func glustershDisableHandler(w http.ResponseWriter, r *http.Request) {
+func selfhealInfoHandler(w http.ResponseWriter, r *http.Request) {
 	p := mux.Vars(r)
 	volname := p["name"]
+	option := p["opts"]
+	_ = option
```
prashanthpai: You don't need this  if you are using `option`

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,129 @@
...
 	ctx := r.Context()
 	logger := gdctx.GetReqLogger(ctx)
 
-	//validate volume name
-	v, err := volume.GetVolume(volname)
-	if err != nil {
-		if err == errors.ErrVolNotFound {
-			restutils.SendHTTPError(ctx, w, http.StatusNotFound, errors.ErrVolNotFound)
+	lock, unlock := transaction.CreateLockFuncs(volname)
+	// Taking a lock outside the txn as volinfo.Nodes() must also
```
prashanthpai: Remove this comment, it's no valid in this case.

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,129 @@
...
 	if err != nil {
-		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
-		return
+		if err == gderrors.ErrVolNotFound {
+			logger.WithError(err).WithField(
+				"volname", volname).Debug("volume not found")
+			restutils.SendHTTPError(ctx, w, http.StatusNotFound, err)
+		} else {
+			logger.WithError(err).WithField(
+				"volname", "volname").Debug("error occured while looking for volume")
```
prashanthpai: `volname` and not `"volname"` in field.

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,129 @@
...
-	lock, unlock, err := transaction.CreateLockSteps(volname)
+	// Validate volume existence
+	volinfo, err := volume.GetVolume(volname)
 	if err != nil {
-		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
-		return
+		if err == gderrors.ErrVolNotFound {
+			logger.WithError(err).WithField(
+				"volname", volname).Debug("volume not found")
+			restutils.SendHTTPError(ctx, w, http.StatusNotFound, err)
```
prashanthpai: Missing return!

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,129 @@
...
+		} else {
+			logger.WithError(err).WithField(
+				"volname", "volname").Debug("error occured while looking for volume")
+			restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
```
prashanthpai: Missing return!

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,129 @@
...
-		},
-		{
-			DoFunc: "vol-option.NotifyVolfileChange",
-			Nodes:  txn.Nodes,
-		},
-		unlock,
+	// Validate volume type
+	if !isVolReplicate(volinfo.Type) {
+		logger.WithError(err).WithField(
+			"volname", volname).Debug("volume type is not supported")
```
prashanthpai: You don't have to log this.

Also return a 400 Bad Request here.

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,129 @@
...
 	}
 
-	if err := txn.Ctx.Set("volinfo", v); err != nil {
-		logger.WithError(err).Error("failed to set volinfo in transaction context")
-		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
-		return
+	// Validate volume state
+	if volinfo.State != volume.VolStarted {
+		logger.WithField("volname", volname).Debug("volume should be in started state")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, gderrors.ErrVolNotStarted)
```
prashanthpai: Log isn't required.

Missing return!

Also return a 400 Bad Request here.

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,129 @@
...
 	}
 
-	if err := txn.Ctx.Set("oldvolinfo", oldvolinfo); err != nil {
-		logger.WithError(err).Error("failed to set volinfo in transaction context")
-		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
-		return
+	var options []string
```
prashanthpai: Move this processing to inside of `getHealInfo` function.

Make `getHealInfo` return glustershdapi.HealInfo.

glustercli/cmd/glustershd.go
```diff
@@ -0,0 +1,96 @@
...
+	helpSelfHealInfoCmd = "Self Heal Info"
+)
+
+var (
+	// Heal Info Flags
+	flagSummaryInfo    bool
+	flagSplitBrainInfo bool
+
+	// Volume Name
+	volname string
```
prashanthpai: This need not be a global variable as it's not used to store value of a flag. You can already access volume name using the variable `volname := args[0]`

glustercli/cmd/glustershd.go
```diff
@@ -0,0 +1,96 @@
...
+		if err != nil {
+			if verbose {
+				log.WithFields(log.Fields{
+					"volume": volname,
+					"error":  err.Error(),
+				}).Error("failed to get heal info")
+			}
+			failure(fmt.Sprintf("Failed to get heal info for volume %s\n", volname), err, 1)
+		}
+
```
prashanthpai: Can you explain what the following code does ? I'm seeing a call to `json.Marshal` immediately followed by a call to `json.Unmarshal`

glusterd2/volume/struct.go
```diff
@@ -80,20 +80,19 @@ type Subvol struct {
...
+	Name      string
+	Type      VolType
+	Transport string
+	DistCount int
+	Options   map[string]string
+	State     VolState
+	Checksum  uint64
+	Version   uint64
+	Subvols   []Subvol
+	Auth      VolAuth // TODO: should not be returned to client
```
prashanthpai: Unrelated to your PR: The TODO comment is no longer valid. As you will be refreshing this PR anyway to address other comments, you can remove it too :)

pkg/restclient/glustershd.go
```diff
@@ -3,20 +3,19 @@ package restclient
...
-	err := c.post(url, nil, http.StatusOK, nil)
-	return err
+// SelfHealInfo sends request to heal-info API
+func (c *Client) SelfHealInfo(params ...string) ([]api.BrickHealInfo, error) {
+	var url string
+	if len(params) == 1 {
+		url = fmt.Sprintf("/v1/volumes/%s/heal-info", params[0])
+	} else {
+		url = fmt.Sprintf("/v1/volumes/%s/%s/heal-info", params[0], params[1])
+	}
```
prashanthpai: You should also handle other cases, such as:

* no params passed
* more than 2 params passed

....in which cases, the caller should be returned an error.

plugins/afr/validate.go
```diff
@@ -9,12 +9,28 @@ import (
 
 var names = [...]string{"replicate", "afr"}
 
+var shdEnabledKey = [...]string{"afr.self-heal-daemon", "replicate.self-heal-daemon"}
```
prashanthpai: `shdKeys` would a better name. The key name itself doesn't enable self heal, the value is what is taken into consideration.

prashanthpai: Wait, is `isVolReplicate` even used in `afr` package ? If not, please remove it.

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,129 @@
...
-
-func glustershEnableHandler(w http.ResponseWriter, r *http.Request) {
-	// Implement the help logic and send response back as below
-	p := mux.Vars(r)
-	volname := p["name"]
-
-	ctx := r.Context()
-	logger := gdctx.GetReqLogger(ctx)
+var (
+	healInfoOutput string
```
prashanthpai: It's not good to use global variables in REST handlers. Why is a global variable needed ?

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,129 @@
...
-			"error":   err.Error(),
-			"volname": volname,
-		}).Error("failed to stop self heal daemon")
+		logger.WithError(err).Error("Error unmarshalling from XML")
 		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
 		return
 	}
 
-	restutils.SendHTTPResponse(ctx, w, http.StatusOK, nil)
+	restutils.SendHTTPResponse(ctx, w, http.StatusOK, &info.Bricks)
```
prashanthpai: While defining rest routes, you have defined `glustershdapi.HealInfo` as the response type. That's the one that should be returned to the client.

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,129 @@
...
-			Nodes:  txn.Nodes,
-		},
-		{
-			DoFunc: "vol-option.NotifyVolfileChange",
-			Nodes:  txn.Nodes,
-		},
-		unlock,
+	// Validate volume type
+	if !isVolReplicate(volinfo.Type) {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "volume type is not supported")
```
prashanthpai: A better error is - `Invalid operation for this volume type`

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,129 @@
...
+	// Validate volume state
+	if volinfo.State != volume.VolStarted {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, gderrors.ErrVolNotStarted)
 		return
 	}
 
-	if err := txn.Ctx.Set("oldvolinfo", oldvolinfo); err != nil {
-		logger.WithError(err).Error("failed to set volinfo in transaction context")
-		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+	if getHealInfo(volname, option) != nil {
```
prashanthpai: The convention followed is as follows:

```go
if err := getHealInfo(volname, option); err != nil {
// use logger.WithError(err).WithField("volname", volname).Debug(....)
}
```

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,129 @@
...
+		logger.WithField("volname", volname).Debug("heal info operation failed")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, "heal info operation failed")
 		return
 	}
 
-	err = txn.Do()
+	output := []byte(healInfoOutput)
```
prashanthpai: Don't use a global variable

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,129 @@
...
+
+	var info glustershdapi.HealInfo
+	err = xml.Unmarshal(output, &info)
 	if err != nil {
-		logger.WithFields(log.Fields{
-			"error":   err.Error(),
-			"volname": volname,
-		}).Error("failed to stop self heal daemon")
+		logger.WithError(err).Error("Error unmarshalling from XML")
```
prashanthpai: `Error unmarshalling XML output from xxxxx command`

plugins/glustershd/rest.go
```diff
@@ -1,197 +1,129 @@
...
-			DoFunc: "selfheal-start",
-			Nodes:  txn.Nodes,
-		},
-		{
-			DoFunc: "vol-option.NotifyVolfileChange",
-			Nodes:  txn.Nodes,
-		},
-		unlock,
-	}
+	healInfoOutput = out.String()
```
prashanthpai: Do not use global variable.

plugins/afr/validate.go
vpandey-RH: isVolReplicate is used in validateOptions() function in afr package

vpandey-RH:
retest this please

glustercli/cmd/glustershd.go
```diff
@@ -0,0 +1,80 @@
...
+		for index := range selfHealInfo {
```
prashanthpai: You can use a table I guess.

plugins/glustershd/init.go
```diff
@@ -18,24 +19,23 @@ func (p *Plugin) Name() string {
...
-			Name:        "GlustershEnable",
-			Method:      "POST",
-			Pattern:     "/volumes/{name}/heal/enable",
-			Version:     1,
-			HandlerFunc: glustershEnableHandler},
+			Name:         "SelfHealInfo",
+			Method:       "GET",
+			Pattern:      "/volumes/{name}/{opts}/heal-info",
+			Version:      1,
+			ResponseType: utils.GetTypeString((*glustershdapi.Brick)(nil)),
```
prashanthpai: You have to re-generate API documentation.

plugins/glustershd/rest.go
prashanthpai: This hasn't been addressed.

plugins/afr/validate.go
```diff
@@ -9,12 +9,26 @@ import (
 
 var names = [...]string{"replicate", "afr"}
 
+// isVolReplicate return true if volume is of type replicate, disperse, distreplicate or distdisperse
```
prashanthpai: s/return/returns

vpandey-RH:
The issue with return type of the API has been handled. I have changed the response type in init.go since I only need heal info for bricks.

plugins/glustershd/init.go
prashanthpai: In both these cases, in the handler, what you are actually returning is a list/slice of `glustershd.Brick` type. Here it's a single instance. Please fix it later. I'll merge this as this incorrect value here doesn't affect functionality.

Issue #431, opened 2017-11-15
bitrot management and API support in GD2


atinmu:
# Things to be done
  Bitrot management daemon
  
# ReST APIs to be defined:

volume bitrot <VOLNAME> {enable|disable}
volume bitrot <volname> scrub-throttle {lazy|normal|aggressive}
volume bitrot <volname> scrub-frequency {hourly|daily|weekly|biweekly|monthly}
volume bitrot <volname> scrub {pause|resume|status|ondemand}

atinmu:
## Checklist
- [ ] ReST APIs availibility in wiki
- [ ] CLI commands
- [ ] Sufficient logging
- [ ] Availibility of metrics data if applicable
- [ ] Xlator options table upto date?
- [ ] Test cases & test coverage of the functionality

atinmu:
@kotreshhr Please cross check the check list for the completion of the items and start ticking out the ones which are being completed here.

kotreshhr:
The checked ones are working via REST API
 
- [ ] volume bitrot {enable|disable}
- [ ] volume bitrot scrub-throttle {lazy|normal|aggressive}
- [ ] volume bitrot scrub-frequency {hourly|daily|weekly|biweekly|monthly}
- [ ] volume bitrot scrub {pause|resume|ondemand}
- [ ] volume bitrot scrub {status}\
- [ ] CLI support for all above

kotreshhr:
This is mostly done. Need to go through if there any pending TODOs and add tests

PR #529, opened 2018-01-17
volgen2: Provide option to ignore options

aravindavk:
Some components like bit-rot uses same xlator for both `bitd` and
`scrubd`, if option `scrubber=on` present in volfile then the daemon
becomes `scrubber` else it becomes `bitd`. While generating `bitd`
volfile `scrubber` option should not be added.

Updates: #431

### DIFF
--- a/glusterd2/volgen2/volgen.go
+++ b/glusterd2/volgen2/volgen.go
@@ -29,6 +29,10 @@ type Entry struct {
 	// option remote-subvolume gfproxyd-<volname>
 	ExtraData  map[string]string
 	SubEntries []Entry
+	// IgnoreOptions represents list of options which should not be added in the
+	// generated Volfile. For example, bitd and scrubd both uses same xlator, if
+	// scrubber = on, then it becomes scrub daemon else it becomes bitd.
+	IgnoreOptions map[string]bool
 }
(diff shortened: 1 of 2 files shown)


glusterd2/volgen2/volgen.go
```diff
@@ -29,6 +29,10 @@ type Entry struct {
 	// option remote-subvolume gfproxyd-<volname>
 	ExtraData  map[string]string
 	SubEntries []Entry
+	// IgnoreOptions represents list of options which should not be added in the
+	// generated Volfile. For example, bitd and scrubd both uses same xlator, if
+	// scrubber = on, then it becomes scrub daemon else it becomes bitd.
+	IgnoreOptions map[string]int
```
prashanthpai: Suggestion: As values are not consumed, this map can be `map[string]bool` or be a set (using map with [emptry struct](https://dave.cheney.net/2014/03/25/the-empty-struct) as value) like `map[string]struct{}`

aravindavk: makes sense. I will update the code

aravindavk:
@prashanthpai Changed the value to bool as suggested. Please review

PR #530, opened 2018-01-19
Add support for validator functions of volume options

kmlmhnn:


glusterd2/commands/volumes/common.go
```diff
@@ -27,6 +28,25 @@ func validateOptions(opts map[string]string) error {
 	return nil
 }
 
+func validateXlatorOptions(opts map[string]string, volinfo *volume.Volinfo) error {
+	for k, v := range opts {
+		_, xl, key, err := options.SplitKey(k)
+		if err != nil {
+			return err
+		}
+		xlator, err := xlator.Find(xl)
```
aravindavk: Use different variable name. Imported library name and this variable name is same

PR #535, opened 2018-01-24
Add support to set global options

samikshan:
Global options can be set using volume name as 'all'.

No support is there for validation of global option values so far.

Signed-off-by: Samikshan Bairagya <samikshan@gmail.com>

### DIFF
--- a/glusterd2/cluster/cluster.go
+++ b/glusterd2/cluster/cluster.go
@@ -12,6 +12,7 @@ import (
 )
 
 const (
+	// ClusterPrefix represents the etcd end-point for cluster wide attributes
 	ClusterPrefix string = "cluster/"
 )
 
@@ -38,8 +39,8 @@ type GlobalOption struct {
 	Type         options.OptionType
 }
 
-// AddClusterAtributes updates etcd with cluster attributes if not already present
-func LoadClusterAtributes() error {
+// LoadClusterAttributes updates etcd with cluster attributes if not already present
+func LoadClusterAttributes() error {
 	var clstr Cluster
 	resp, err := store.Store.Get(context.TODO(), ClusterPrefix)
 	if err != nil {
@@ -49,7 +50,7 @@ func LoadClusterAtributes() error {
 	if resp.Count == 0 {
 		// If cluster instance isn't available add it to etcd
 		clstr.Options = make(map[string]string)
-		for k, _ := range GlobalOptMap {
+		for k := range GlobalOptMap {
 			clstr.Options[k] = GlobalOptMap[k].DefaultValue
 		}
 
@@ -60,6 +61,5 @@ func LoadClusterAtributes() error {
 	}
 
 	// If cluster instance is already available on etcd just return
-
 	return nil
 }
--- a/glusterd2/cluster/cluster.go
+++ b/glusterd2/cluster/cluster.go
@@ -1,19 +1,10 @@
 package cluster
 
 import (
-	"context"
-	"encoding/json"
-
 	"github.com/gluster/glusterd2/glusterd2/gdctx"
-	"github.com/gluster/glusterd2/glusterd2/store"
 	"github.com/gluster/glusterd2/glusterd2/xlator/options"
 
 	"strconv"
-)
-
-const (
-	// ClusterPrefix represents the etcd end-point for cluster wide attributes
-	ClusterPrefix string = "cluster/"
 )
 
 // GlobalOptMap contains list of supported cluster-wide options, default values and value types
@@ -37,29 +28,4 @@ type GlobalOption struct {
 	Key          string
 	DefaultValue string
 	Type         options.OptionType
-}
-
-// LoadClusterAttributes updates etcd with cluster attributes if not already present
-func LoadClusterAttributes() error {
-	var clstr Cluster
-	resp, err := store.Store.Get(context.TODO(), ClusterPrefix)
-	if err != nil {
-		return err
-	}
-
-	if resp.Count == 0 {
-		// If cluster instance isn't available add it to etcd
-		clstr.Options = make(map[string]string)
-		for k := range GlobalOptMap {
-			clstr.Options[k] = GlobalOptMap[k].DefaultValue
-		}
-
-		data, _ := json.Marshal(clstr)
-		if _, err := store.Store.Put(context.TODO(), ClusterPrefix, string(data)); err != nil {
-			return err
-		}
-	}
-
-	// If cluster instance is already available on etcd just return
-	return nil
 }
--- a/glusterd2/cluster/cluster.go
+++ b/glusterd2/cluster/cluster.go
@@ -1,0 +1,65 @@
+package cluster
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/gluster/glusterd2/glusterd2/gdctx"
+	"github.com/gluster/glusterd2/glusterd2/store"
+	"github.com/gluster/glusterd2/glusterd2/xlator/options"
+
+	"strconv"
+)
+
+const (
+	ClusterPrefix string = "cluster/"
+)
+
+// GlobalOptMap contains list of supported cluster-wide options, default values and value types
+var GlobalOptMap = map[string]GlobalOption{
+	"cluster.server-quorum-ratio":    {"cluster.server-quorum-ratio", "51", options.OptionTypePercent},
+	"cluster.shared-storage":         {"cluster.shared-storage", "disable", options.OptionTypeBool},
+	"cluster.op-version":             {"cluster.op-version", strconv.Itoa(gdctx.OpVersion), options.OptionTypeInt},
+	"cluster.max-op-version":         {"cluster.max-op-version", strconv.Itoa(gdctx.OpVersion), options.OptionTypeInt},
+	"cluster.brick-multiplex":        {"cluster.brick-multiplex", "disable", options.OptionTypeBool},
+	"cluster.max-bricks-per-process": {"cluster.max-bricks-per-process", "0", options.OptionTypeInt},
+	"cluster.localtime-logging":      {"cluster.localtime-logging", "disable", options.OptionTypeBool},
+}
+
+// Cluster contains cluster-wide attributes
+type Cluster struct {
+	Options map[string]string
+}
+
+// GlobalOption reperesents cluster wide options
+type GlobalOption struct {
+	Key          string
+	DefaultValue string
+	Type         options.OptionType
+}
+
+// AddClusterAtributes updates etcd with cluster attributes if not already present
+func LoadClusterAtributes() error {
+	var clstr Cluster
+	resp, err := store.Store.Get(context.TODO(), ClusterPrefix)
+	if err != nil {
+		return err
+	}
+
+	if resp.Count == 0 {
+		// If cluster instance isn't available add it to etcd
+		clstr.Options = make(map[string]string)
+		for k, _ := range GlobalOptMap {
+			clstr.Options[k] = GlobalOptMap[k].DefaultValue
+		}
+
+		data, _ := json.Marshal(clstr)
+		if _, err := store.Store.Put(context.TODO(), ClusterPrefix, string(data)); err != nil {
+			return err
+		}
+	}
+
+	// If cluster instance is already available on etcd just return
+
+	return nil
+}
--- a/glusterd2/commands/global/commands.go
+++ b/glusterd2/commands/global/commands.go
@@ -1,0 +1,35 @@
+// Package globaloptions implements the commands to get and set cluster level options
+package globalcommands
+
+import (
+	"github.com/gluster/glusterd2/glusterd2/servers/rest/route"
+)
+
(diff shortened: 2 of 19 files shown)


atinmu:
22:24:18 ./glusterd2/xlator/options/options.go:87:6: exported type GlobalOption should have comment or be unexported
22:24:18 Found 1 lint suggestions; failing.
22:24:18 ./glusterd2/xlator/global.go:20:2: exported var GlobalOptMap should have comment or be unexported
22:24:18 ./glusterd2/xlator/global.go:64:1: exported function LoadGlobalOptions should have comment or be unexported
22:24:18 ./glusterd2/xlator/global.go:65:9: should omit 2nd value from range; this loop is equivalent to `for k := range ...`
22:24:18 Found 3 lint suggestions; failing.
22:24:18 ./glusterd2/xlator/global.go does not conform to gofmt rules. Run: gofmt -s -w ./glusterd2/xlator/global.go
22:24:18 ./glusterd2/xlator/utils.go:10:2: exported const GlobalOptsPrefix should have comment (or a comment on this block) or be unexported
22:24:18 Found 1 lint suggestions; failing.
22:24:19 ./pkg/api/volume_resp.go:74:1: comment on exported type VolumeOptionsGetResp should be of the form "VolumeOptionsGetResp ..." (with optional leading article)
22:24:19 Found 1 lint suggestions; failing.

^^ lint failures.  @samikshan please fix this.

kshlm:
Let's try not to shoehorn global options (the ones that work on volume "all") into the normal volume set API/command. The options being set here don't just operate on volumes, but instead act on the full cluster (eg. cluster.op-version).

I'd really like this to be a seperate API and a seperate command. Ideally, all these would be set on a cluster-object that is persisted in the store.

amarts:
the 'global' option possible with https://review.gluster.org/#/c/18315/ is meant to be more of per volume option, and not part of 'all'. This translator is making glusterfs's 'ctx' specific options as a possibility.

The one with 'all' is as mentioned by @kshlm are global for the cluster (and may be for glusterd itself)

atinmu:
For all global cluster specific options like op-version, brick-multiplex et all are only limited to glusterd. Not sure if having a different ReST API would make much sense here as of users it's still to be done through a volume set interface, but I agree that we shouldn't try to put the logic of this segregation in the main volume set function. Would it make much sense to have something like:

gluster volume set cluster <global option> <value> for the ones which are related to cluster level tuning?

aravindavk:
We need to clearly differentiate between the global options per volume(via global xlator) and Cluster wide options. 

As discussed with @prashanthpai and @amarts, Global options per Volume can be set using the same interface available today for other Volume options. (May need some changes to parse global.so file). For example, `glustercli volume set <volname> global.<opt-name> <value>`. This needs to be added to all generated Volfiles.

Setting Cluster wide options: Who will use these options? will it be added to volfile? or If it is default option which can be configured before any Volume Create, so that all new Volume create will have these options set by default. @atinmu @amarts

atinmu:
> Setting Cluster wide options: Who will use these options? will it be added to volfile? or If it is default option which can be configured before any Volume Create, so that all new Volume create will have these options set by default. @atinmu @amarts

These options will not be written to any volfiles as the attributes are applicable at cluster level and they are purely a cluster wide tuning, they don't apply to a volume level and limited to glusterd only. As suggested by Kaushal, we just need to have these values stored in etcd.

samikshan:
Right now https://github.com/gluster/glusterfs/blob/e7ee4d8949b015a1421d4d97b763ccb4dcfc6eb1/cli/src/cli-cmd-parser.c#L410 contains the list of volume names that aren't allowed currently. Implementing something like `volume set cluster` to set cluster level options, would add "cluster" to that list of volume names. Are we okay with that kind of a change?

The other option which is to have separate APIs/commands for setting/getting these options could be implemented. However, I am unsure about what data points should be present in the cluster object other than these options. Peer list seems like one of the data points but it would be redundant given that the peer list is already maintained.

glusterd2/commands/global/commands.go
```diff
@@ -0,0 +1,35 @@
...
+			Name:        "SetGlobalOptions",
+			Method:      "POST",
+			Pattern:     "/cluster/options",
+			Version:     1,
+			HandlerFunc: setGlobalOptionsHandler,
+		},
+		route.Route{
+			Name:        "GetGlobalOptions",
+			Method:      "GET",
+			Pattern:     "/cluster/options/{optname}",
```
aravindavk: Return all options at once, CLI can filter if required. (optname can be removed)

glusterd2/commands/global/getglobaloptions.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/gorilla/mux"
+)
+
+func getGlobalOptionsHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	optname := mux.Vars(r)["optname"]
+
+	var c cluster.Cluster
+	resp, err := store.Store.Get(context.TODO(), cluster.ClusterPrefix)
```
aravindavk: Create this as function in separate file `store-utils.go`. So that all store related functions will be in one place.

glusterd2/commands/global/setglobaloptions.go
```diff
@@ -0,0 +1,58 @@
...
+	ctx := r.Context()
+
+	var req api.GlobalOptionReq
+	if err := restutils.UnmarshalRequest(r, &req); err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusUnprocessableEntity, errors.ErrJSONParsingFailed.Error(), api.ErrCodeDefault)
+		return
+	}
+
+	var c cluster.Cluster
+	resp, err := store.Store.Get(context.TODO(), cluster.ClusterPrefix)
```
aravindavk: Same as above comment, keep store related things in a separate file.

glusterd2/commands/global/setglobaloptions.go
```diff
@@ -0,0 +1,58 @@
...
+
+			c.Options[k] = v
+		} else {
+			restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, fmt.Sprintf("Invalid global option: %s", k), api.ErrCodeDefault)
+			continue
+		}
+	}
+
+	data, _ := json.Marshal(c)
+	if _, err := store.Store.Put(context.TODO(), cluster.ClusterPrefix, string(data)); err != nil {
```
aravindavk: same

glusterd2/commands/global/setglobaloptions.go
```diff
@@ -0,0 +1,58 @@
...
+		return
+	}
+
+	if err = json.Unmarshal(resp.Kvs[0].Value, &c); err != nil {
+		return
+	}
+
+	for k, v := range req.Options {
+		if _, found := cluster.GlobalOptMap[k]; found {
+			// TODO validate the value type for global option
```
aravindavk: Is it possible to utilize the volume options validation functions here?

glusterd2/commands/volumes/common.go
```diff
@@ -18,7 +18,6 @@ import (
 // validateOptions validates if the options and their values are valid and can
 // be set on a volume.
 func validateOptions(opts map[string]string) error {
-
```
aravindavk: nit: change not related to this patch

glusterd2/main.go
```diff
@@ -108,6 +109,10 @@ func main() {
 		log.WithError(err).Fatal("Failed to generate local auth token")
 	}
 
+	if err := cluster.LoadClusterAttributes(); err != nil {
```
aravindavk: Not really required. since all the default values and other details available with Glusterd2 itself.

glusterd2/xlator/global.go
```diff
@@ -14,7 +14,6 @@ var (
 
 // Load load all available xlators and intializes the xlators and options maps
 func Load() (err error) {
-
```
aravindavk: nit: unrelated

glusterd2/xlator/global.go
```diff
@@ -37,6 +36,7 @@ func Xlators() map[string]*Xlator {
 
 // loadOptions loads all available options into the options.Options map,
 // indexed as <xlator-id>.<option-key> for all available option keys
+// Also initialise the globalOptMap with defauls/value from etcd server
```
aravindavk: Is this comment belong to previous patchset?

pkg/restclient/volume.go
```diff
@@ -69,6 +69,13 @@ func (c *Client) VolumeSet(volname string, req api.VolOptionReq) error {
 	return err
 }
 
+// GlobalOptionSet sets cluster level options
+func (c *Client) GlobalOptionSet(req api.GlobalOptionReq) error {
+	url := fmt.Sprintf("/v1/cluster/options")
+	err := c.post(url, req, http.StatusOK, nil)
```
aravindavk: skip err variable.

    return c.post(url, req, http.StatusOK, nil)

glusterd2/cluster/cluster.go
```diff
@@ -0,0 +1,65 @@
...
+
+// GlobalOption reperesents cluster wide options
+type GlobalOption struct {
+	Key          string
+	DefaultValue string
+	Type         options.OptionType
+}
+
+// LoadClusterAttributes updates etcd with cluster attributes if not already present
+func LoadClusterAttributes() error {
```
aravindavk: This function is not required.  Store only modified/set global options.

glusterd2/commands/global/getglobaloptions.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/gluster/glusterd2/glusterd2/cluster"
+	restutils "github.com/gluster/glusterd2/glusterd2/servers/rest/utils"
+	"github.com/gluster/glusterd2/glusterd2/store"
+	"github.com/gluster/glusterd2/pkg/api"
+	"github.com/gluster/glusterd2/pkg/errors"
+	"github.com/gorilla/mux"
+)
+
+func getGlobalOptionsHandler(w http.ResponseWriter, r *http.Request) {
```
aravindavk: Read default options to temp opts, then get from store and update temp opts and return. Good to return the following details.

     Name
     Value
     Default Value
     ModifiedFlag

pkg/restclient/volume.go
```diff
@@ -69,6 +69,13 @@ func (c *Client) VolumeSet(volname string, req api.VolOptionReq) error {
...
+	return err
+}
+
```
aravindavk: Also a func required to get Global options

glusterd2/commands/global/setglobaloptions.go
samikshan: The validation functions for volume options will work only on the separately defined option type in xlator/options/options.go. In fact Validate() is a method and works only on volume options for now. I do feel that they should be made reusable for cluster wide option validation.

glusterd2/commands/global/getglobaloptions.go
samikshan: Sure.

glusterd2/xlator/global.go
samikshan: Yes it is! Will remove.

glusterd2/commands/global/getglobaloptions.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/gluster/glusterd2/glusterd2/cluster"
+	restutils "github.com/gluster/glusterd2/glusterd2/servers/rest/utils"
+	"github.com/gluster/glusterd2/pkg/api"
+)
+
+func getGlobalOptionsHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	c, _ := cluster.GetCluster()
```
aravindavk: Don't ignore error. 

Ignore only specific error if required. Error may be due to etcd connectivity.

pkg/restclient/volume.go
```diff
@@ -69,6 +69,12 @@ func (c *Client) VolumeSet(volname string, req api.VolOptionReq) error {
 	return err
 }
 
+// GlobalOptionSet sets cluster level options
+func (c *Client) GlobalOptionSet(req api.GlobalOptionReq) error {
+	url := fmt.Sprintf("/v1/cluster/options")
+	return c.post(url, req, http.StatusOK, nil)
+}
+
```
aravindavk: Also need `GlobalOptions()`

samikshan: If you mean GlobalOptionsGet() do you think it makes sense to add this part after the patch on volume get support is merged?

glusterd2/commands/global/getglobaloptions.go
samikshan: Yes. Missed the error handling here.

pkg/restclient/volume.go
aravindavk: Sure, makes sense.

samikshan:
Global options reset can definitely be done. I would prefer to work on it in a later patch though.

PR #536, opened 2018-01-25
bitrot: Add scrub ondemand support

kotreshhr:
Bitrot scrubber daemon supports ondemand scrub by
providing brick-op service. Glusterd invokes rpc
to avail the service.

Updates: #431

### DIFF
--- a/plugins/bitrot/rest.go
+++ b/plugins/bitrot/rest.go
@@ -37,8 +37,7 @@ func bitrotEnableHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Check if bitrot is already enabled
-	val, exists := volinfo.Options[volume.VkeyFeaturesBitrot]
-	if exists && val == "on" {
+	if volume.IsBitrotEnabled(volinfo) {
 		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, errors.ErrBitrotAlreadyEnabled.Error(), api.ErrCodeDefault)
 		return
 	}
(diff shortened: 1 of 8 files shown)


plugins/bitrot/transactions.go
```diff
@@ -97,3 +98,41 @@ error:
...
+			"volume", volname).Error("failed to connect to scrubd")
+		return err
+	}
+
+	req := &brick.GfBrickOpReq{
+		Name: volname,
+		Op:   brick.OpNodeBitrot,
+	}
+	var rsp brick.GfBrickOpRsp
+	err = client.Call("BrickOp", req, &rsp)
```
prashanthpai: PR #537 fixes the RPC issue and is in master now. Please rebase. This call will now look like:

```go
client.Call("Brick.OpNodeBitrot", req, &rsp)
```

kotreshhr: dict unserialize is failing at scrub daemon end. Let me retest.

[2018-01-25 10:19:23.190595] W [dict.c:2835:dict_unserialize] (-->/usr/local/lib/libglusterfs.so.0(+0x72ab3) [0x7f28cd911ab3] -->/usr/local/sbin/glusterfs(glusterfs_handle_bitrot+0x1d4) [0x40d1a1] -->/usr/local/lib/libglusterfs.so.0(dict_unserialize+0xa7) [0x7f28cd8bed65] ) 0-dict: buf is null! [Invalid argument]

kotreshhr: Fixed! It was using dict and had serialize. But DictSerialize was in sunrpc package which was causing cycling dependency. Hence moved to package dict

plugins/bitrot/rest.go
```diff
@@ -177,5 +177,67 @@ func bitrotDisableHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	// Check if volume is started
+	if volinfo.State != volume.VolStarted {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, errors.ErrVolNotStarted.Error(), api.ErrCodeDefault)
+		return
+	}
+
+	// Check if bitrot is disabled
+	val, exists := volinfo.Options[volume.VkeyFeaturesBitrot]
+	if exists && val == "off" {
```
aravindavk: This condition is wrong when bitrot is not enabled. (exists will be false). Better make this as function in volume as `volume.IsBitrotEnabled`. since it is used in multiple places now.

kotreshhr: Done

aravindavk:
Looks good. CI is failing due to a golint error.

kotreshhr:
Fixed the golint errors

Issue #482, opened 2017-12-11
Add support to set arbitrary metadata on peer and volume object 


prashanthpai:
...glusterd2 shall not interpret these metadata (K/V pairs) set by users (or programs) in anyway. Glusterd2 may however impose a size restriction.

This additional metadata can be set and consumed by tools that needs to tag peers for say rack awareness or in future tag volumes or even devices with a "class".

A pre-requisite for implementing this is providing option to edit peer object after creation. Refer issue #462 and #356

rishubhjain:
@prashanthpai So one of the additional metadata can be group/zone the peer belongs to which can be saved along with Peer information and this metadata can be used by  the heketi plugin (after integration). Does this sound right?

prashanthpai:
@rishubhjain Yes. Correct. Glusterd2 will not interpret these metadata in any way. They are key-value pairs.

rishubhjain:
@atinmu @prashanthpai @kmlmhnn I am working on this issue, User will be able to group the nodes for stack awareness by providing the group number while attaching peer . This is partial implementation of this issue, but it will be used for dynamic volume provisioning.

Example post request:

{
    "addresses": [
        "host1name1",
        "host1name2"
    ],
    "group": "1"
}

prashanthpai:
@rishubhjain Consider not introducing a new member to `Peer` struct just for group. You can create `Metadata` as the member which can contain many key value pairs with "group" being one of them,

amarts:
After this is available:

*  We need a mechanism to add further logic in volume create based on these tags (may be like hooks, provide filters? (or consume it in IVP when available).
* Similar for volume tags, it should be available for plugins, so they can decide further options. For example, I will consume the volumes which are set with a tag 'Block' as the backend volume for block files, and nothing else. (ie, block cli itself fails if we provide other volume) and things like this.

None of the core functionality should depend on these tags, but the higher level functionalities can depend on it.

prashanthpai:
Listing (filtering) based on some key and/or value has been done for volume listing (https://github.com/gluster/glusterd2/issues/462) by @vpandey-RH 

@rishubhjain or @vpandey-RH Any takers to do the same for peer listing as well ?

vpandey-RH:
@prashanthpai  I can start working on this if @rishubhjain  is not looking to work on this.

rishubhjain:
@vpandey-RH Go ahead

prashanthpai:
[...]

Issue #416, opened 2017-11-13
node grouping option 


amarts:
## Problem Statement

There is no awareness to glusterd about where each node is, and there are no grouping of these nodes. One of the major usecase of grouping few nodes into different groups is rack awareness for the config layer. Also it can be extended to group them as different group of machines with different warranty expiry dates etc.

### Why do we need it?

Rack awareness, mainly because a true high availability in a local cluster be achieved when you can place your data on different racks, and not in adjacent machines.
Warranty period so that you can avoid sudden collapse of storage in one (or few) set of replica pairs. 

### What improvements can be immediately done?

* Provide option to `--group <group_name>` during `peer probe $peer`
* Provide an option to edit 'peer' details, where existing `group` can be changed. (There can be more properties for a given peer, IP/hostname being few of them).

atinmu:
IMO, we'd need check how heketi has handled this so far and have the respective APIs defined in glusterd2.

prashanthpai:
> Provide option to --group <group_name> during peer probe $peer

We had planned this - ability to set custom key/value pairs (metadata) for peers. The higher layers (dynamic provisioning bits) can set and interpret them in whatever way they would like. 

> Provide an option to edit 'peer' details, where existing group can be changed. (There can be more properties for a given peer, IP/hostname being few of them).

This is also planned. Phase 1 to get this done was to use `UUID` to represent bricks internally (unilke gd1 ehere it was `host:path`). So when the IP of a peer changes, we just have to change the peer list. That change has been completed.

aravindavk:
+1

We also need to tag/group the IPs of a peer, so that we can specify a IP which will be used only for Internal traffic, management traffic and client traffic etc. (Currently we are accepting hostname/IP as list)

atinmu:
## Checklist
- [ ] ReST APIs availibility in wiki
- [ ] CLI commands
- [ ] Sufficient logging
- [ ] Availibility of metrics data if applicable
- [ ] Xlator options table upto date?
- [ ] Test cases & test coverage of the functionality

rishubhjain:
@atinmu The grouping of nodes(for stack awareness) in heketi is done via mentioning the zone they(node) belong to  which is done when node is added (Please refer to https://github.com/heketi/heketi/blob/master/doc/api/api.md#add-node).
We can introduce similar concept in GD2 by requiring user to input the zone/group (whichever name is appropriate) while attaching peer.
[...]

PR #541, opened 2018-01-30
WIP: Node Grouping

rishubhjain:
For each node one can mention group that the particular node belongs to in the conf file.

for example:
$ cat conf.yaml
workdir: "/var/lib/gd2"
peeraddress: "192.168.56.101:24008"
clientaddress: "192.168.56.101:24007"
etcdcurls: "http://192.168.56.101:2379"
etcdpurls: "http://192.168.56.101:2380"
group: 1

During attaching new peer, user will have to mention the group of the new node in the post request.
Example post request:

{
"addresses": [
"host1name1",
"host1name2"
],
"group": "1"
}

Please refer : https://github.com/gluster/glusterd2/issues/482 , https://github.com/gluster/glusterd2/issues/416

### DIFF
--- a/glusterd2/commands/peers/addpeer.go
+++ b/glusterd2/commands/peers/addpeer.go
@@ -15,11 +15,11 @@ import (
 
 type peerAddReq struct {
 	Addresses []string
-        Group     int
+	Group     int
 }
 
 func addPeerHandler(w http.ResponseWriter, r *http.Request) {
-        fmt.Printf("$$$$$$$$$$$$ In addPeerHandler")
+
 	ctx := r.Context()
 	logger := gdctx.GetReqLogger(ctx)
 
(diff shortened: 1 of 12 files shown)


rishubhjain:
@prashanthpai Please review. For now I have just added another parameter (group) instead of metadata as you suggested, I will like to  get the method approved before incorporating those suggestions.

aravindavk:
Adding group/metadata to config is not looking right to me. Maintaining different config for each node is not easy to maintain when we scale.

You can accept group info during peer probe or a separate command/API to do that. metadata approach looks good to me since we can have many logical groups based on the key value pairs, not just group.

@prashanthpai let me know your thoughts

prashanthpai:
@rishubhjain +1 to Aravinda's comment. 

The `Metadata` (key-value pairs) field in `Peer` object must be settable during peer add. And later we should introduce an API to change existing value. There can be many more keys than just the `Group`. The core of glusterd2 will not read or interpret these keys and values in any way.

rishubhjain:
Please refer to PR : https://github.com/gluster/glusterd2/pull/549

PR #542, opened 2018-01-30
Update spec to prepare for RC0

kshlm:


extras/rpms/glusterd2.spec
```diff
@@ -10,15 +10,15 @@
...
+Version: 4.0rc0
+Release: 1
 Summary: The GlusterFS management daemon (preview)
 License: GPLv2 or LGPLv3+
 URL: https://%{provider_prefix}
 # Use vendored tarball instead of plain git archive
 Source0: https://%{provider_prefix}/releases/download/v%{version}-%{release}/%{name}-v%{version}-%{release}-vendor.tar.xz
 
-ExclusiveArch: x86_64
+ExclusiveArch: %{go_arches}
```
prashanthpai: I'm not very familiar with the RPM packaging. Can you check if the package providing this macro needs to be added explicitly as build time requirement ?
`BuildRequires:  go-srpm-macros`

prashanthpai: It's in the minimal build root apparently :+1:

PR #547, opened 2018-02-02
Separate Peer and Client Addresses in Peer object

prashanthpai:
Peer address is used by glusterd2 to communicate with each other. Client address is used by clients (CLI, mount, curl) to communicate with glusterd2. A single `Addresses` currently showed in response and CLI is unusable by clients as it's address used for internal communication

### DIFF
--- a/glustercli/cmd/peer.go
+++ b/glustercli/cmd/peer.go
@@ -60,8 +60,8 @@ var peerProbeCmd = &cobra.Command{
 		}
 		fmt.Println("Peer probe success")
 		table := tablewriter.NewWriter(os.Stdout)
-		table.SetHeader([]string{"ID", "Name", "Addresses"})
-		table.Append([]string{peer.ID.String(), peer.Name, strings.Join(peer.Addresses, ",")})
+		table.SetHeader([]string{"ID", "Name", "PeerAddresses"})
+		table.Append([]string{peer.ID.String(), peer.Name, strings.Join(peer.PeerAddresses, ",")})
 		table.Render()
 	},
 }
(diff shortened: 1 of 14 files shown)


prashanthpai:
This is required for getspec PR and also will be good to get in before peer metadata PR that Rishub is working on.

aravindavk:
Any change expected during peer probe? Description about ClientAddress vs Peer address would be helpful.

prashanthpai:
> Any change expected during peer probe

No changes there.

> Description about ClientAddress vs Peer address would be helpful.

It's [there](https://github.com/gluster/glusterd2/blob/master/doc/network.md) but not easily accessible. I'll edit the REST API wiki to add documentation.

aravindavk:
Looks good to me. 

Just a thought! How about maintaing extra info along with address itself? So that many grouping can be easily done later(Now only limited to Peer IP and Client IP). For example, Client IPs, Volume IPs, Mgmt IPs etc. 

    type PeerIPGroup int

    const (
        DefaultIP PeerIPGroup = iota
        ClientIP
        PeerIP
        VolumeIP
    )

    type PeerAddress struct {
        Hostname string
        Tags []PeerIPGroup
    }

prashanthpai:
@aravindavk That's a good idea! But we can incorporate it later when the usecase comes. One use-case where we might need is routing rebalance/self-heal traffic through a particular separate network interface. For now the naming of peer and client address is modelled after how etcd does it. Etcd has a peerurls for communication between etcd nodes and a clienturls for clients to talk to etcd.

PR #549, opened 2018-02-02
Creating meta data for peer

rishubhjain:
Issue https://github.com/gluster/glusterd2/issues/482


rishubhjain:
@prashanthpai @aravindavk Can you review this PR and check if its as per comments on https://github.com/gluster/glusterd2/pull/541

aravindavk:
How about adding MetaData after peer is successfully joined? After JoinCluster, you will have peerid, Use that peerid and update the metadata in store(All those extra field in RPC can be removed)

rishubhjain:
@aravindavk thanks for the review, yes it can be done, @prashanthpai and I discussed this idea.
@prashanthpai Should I go ahead and make the changes as per aravinda's suggestion?

prashanthpai:
> How about adding MetaData after peer is successfully joined? After JoinCluster, you will have peerid, Use that peerid and update the metadata in store(All those extra field in RPC can be removed)

Sounds good to me. @rishubhjain Please go ahead

rishubhjain:
@aravindavk @prashanthpai I have updated the PR as per your suggestions, Please review.

PR #551, opened 2018-02-02
Setup dep usage in make to allow env, like proxy settings

sheenobu:
Use like:

   make DEPENV=http_proxy=...


prashanthpai:
add to whitelist

PR #552, opened 2018-02-03
xlator: report proper error message when xlator dir is empty

sheenobu:
Before

> Failed to load xlator options error=stat : no such file or directory

After:

> Failed to load xlator options error=No xlators dir foun


prashanthpai:
add to whitelist

PR #556, opened 2018-02-05
bitrot: Add cli support

kotreshhr:
Usage:

    glustercli bitrot <volname> {enable|disable}
    glustercli bitrot  scrub-throttle <volname> {lazy|normal|aggressive}
    glustercli bitrot scrub-frequency <volname> {hourly|daily|weekly|biweekly|monthly}
    glustercli bitrot scrub <volname> {pause|resume|status|ondemand}

NOTE: This patch doesn't support scrub status. It will be done in a separate patch

Issue: #431


glustercli/cmd/bitrot.go
```diff
@@ -0,0 +1,169 @@
...
+
+var bitrotScrubFrequencyCmd = &cobra.Command{
+	Use:   "scrub-freq <volname> {hourly|daily|weekly|biweekly|monthly}",
+	Short: helpBitrotScrubFrequencyCmd,
+	Args:  cobra.ExactArgs(2),
+	Run: func(cmd *cobra.Command, args []string) {
+		volname := args[0]
+		var option []string
+		option = append(option, "bit-rot.scrub-freq")
+		option = append(option, args[1])
```
aravindavk: One clarification: Are these Volume options validated on Volume Set?

pkg/restclient/bitrot.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"fmt"
+	"net/http"
+)
+
+// BitrotEnable enables bitrot for a volume
+func (c *Client) BitrotEnable(volname string) error {
+	url := fmt.Sprintf("/v1/volumes/%s/bitrot/enable", volname)
+	err := c.post(url, nil, http.StatusOK, nil)
```
aravindavk: Return directly without err variable

    return c.post(url, nil, http.StatusOK, nil)

pkg/restclient/bitrot.go
```diff
@@ -0,0 +1,27 @@
...
+	return err
+}
+
+// BitrotDisable disables bitrot for a volume
+func (c *Client) BitrotDisable(volname string) error {
+	url := fmt.Sprintf("/v1/volumes/%s/bitrot/disable", volname)
+	err := c.post(url, nil, http.StatusOK, nil)
```
aravindavk: same as above

pkg/restclient/bitrot.go
```diff
@@ -0,0 +1,27 @@
...
+	return err
+}
+
+// BitrotScrubOndemand starts bitrot scrubber on demand for a volume
+func (c *Client) BitrotScrubOndemand(volname string) error {
+	url := fmt.Sprintf("/v1/volumes/%s/bitrot/scrubondemand", volname)
+	err := c.post(url, nil, http.StatusOK, nil)
```
aravindavk: Same as above

glustercli/cmd/bitrot.go
kotreshhr: Nope, I will add validation

pkg/restclient/bitrot.go
kotreshhr: Done

pkg/restclient/bitrot.go
kotreshhr: Done

pkg/restclient/bitrot.go
kotreshhr: Done

kotreshhr:
Addressed review comments. @aravindavk, please review

glusterd2/xlator/validation.go
```diff
@@ -19,6 +19,30 @@ func validateReplica(v *volume.Volinfo, key string, value string) error {
...
 }
 
+func validateBitrot(v *volume.Volinfo, key string, value string) error {
+	switch key {
+	case "scrub-throttle":
+		acceptedThrottleValues := []string{"lazy", "normal", "aggressive"}
+		if Contains(value, acceptedThrottleValues) {
+			return nil
+		}
+		return errors.New("Invalid option value")
```
prashanthpai: Suggestion: it'll be nicer for the errors to contain some more detail. w.r.t to which key is invalid and which values are valid ones. You can propagate the same back to  CLI too.

```go
var err error
...
	err := fmt.Errorf("Invalid value specified for option '%s'. Possible values: %s", key, strings.Join(acceptedThrottleValues, ", "))
...
return err
```

glustercli/cmd/bitrot.go
```diff
@@ -0,0 +1,203 @@
...
+				"volume": volname,
+				"error":  err.Error(),
+			}).Error("failed to disable bitrot")
+			failure(fmt.Sprintf("Failed to disable bitrot for volume %s\n", volname), 1)
+		}
+		fmt.Printf("Bitrot disabled successfully for volume '%s'\n", volname)
+	},
+}
+
+func bitrotValidScrubThrottleValue(value string) bool {
```
aravindavk: These validator functions not required in CLI since API will validate the values.  In future Volume Set API can return different error codes for Invalid Key, Invalid Value etc that can be used in CLI. For now, we can remove the check from here.

kotreshhr: Done

glusterd2/xlator/validation.go
kotreshhr: Done

kotreshhr:
@aravindavk @prashanthpai Addressed review comments. Also added support for scrub status as well in the same patch. Please review.

PR #562, opened 2018-02-09
Switch to github.com/cockroachdb/cmux

kshlm:
This is part of the attempt to consolidate our dependencies.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -26,13 +26,83 @@
   version = "v1.0.0"
 
 [[projects]]
+  branch = "master"
   name = "github.com/cockroachdb/cmux"
   packages = ["."]
-  revision = "112f0506e7743d64a6eb8fedbcff13d9979bbf92"
+  revision = "30d10be492927e2dcae0089c374c455d42414fcb"
 
 [[projects]]
   name = "github.com/coreos/etcd"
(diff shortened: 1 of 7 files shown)


prashanthpai:
The library changes look good. 

However, TLS settings were explicitly used only for REST sub-listener intentionally.  This was for the possibility of having an option to turn on TLS just for the management API separately and not for sunrpc clients.

kshlm:
@prashanthpai I'd like to keep the number of supportable options down as much as possible. Adding seperate configs for each client server would be going in the other direction. As of right now, we already have two different configs for peer connections and client connections. This IMO is a good enough split.

Also, `cockroachdb/cmux` doesn't support the `cmux.TLS` matcher.

aravindavk:
I agree with @prashanthpai, we should be able to control TLS selectively between sunrpc and REST.

prashanthpai:
> @prashanthpai I'd like to keep the number of supportable options down as much as possible.

Me too. However, in this scenario, it's a real use case because the same port is used for supporting management operations, I/O client operations and even brick communications (for example: signin). The admin or operator may want to enable TLS based authentication for management operations only and not for clients. Because we use the same port, firewall based crude filtering is also ruled out.

> Also, cockroachdb/cmux doesn't support the cmux.TLS matcher.

I see. Is it possible to import (if the license is compatible) the original cmux library into `pkg` package ?

kshlm:
> Me too. However, in this scenario, it's a real use case because the same port is used for supporting management operations, I/O client operations and even brick communications (for example: signin). The admin or operator may want to enable TLS based authentication for management operations only and not for clients. Because we use the same port, firewall based crude filtering is also ruled out.

Not really. Even though we do use sunrpc, the stuff it's used for is still part of the management domain. But I can believe that a user might want to selectively enable TLS (but I'd believe that to be very rare).

But, it will also be easier later to expand the supported options; than it will be to add extra options now and remove them later.

> I see. Is it possible to import (if the license is compatible) the original cmux library into pkg package ?
Nope that would be considered the same as bundling/vendoring. One thing we could do about the `cmux.TLS` matcher, would be to implement our own matcher for TLS.

aravindavk:
> Not really. Even though we do use sunrpc, the stuff it's used for is still part of the management domain. But I can believe that a user might want to selectively enable TLS (but I'd believe that to be very rare).

REST APIs will be directly used in many integrations(Mobile/web apps, Remote CLI etc) but sunrpc will be used for internal communications and limited number of external machines.  If TLS is not selective then we end up disabling it for REST apis if it is disabled for sunrpc clients.

kshlm:
@prashanthpai @aravindavk I've added a custom TLS matcher and gone back to having TLS just for ReST.

Issue #311, opened 2017-05-27
Implement /info or /endpoints endpoint


prashanthpai:
Implement `/info` or `/endpoints` for service or endpoint discovery.

PR #563, opened 2018-02-13
Implement /endpoints API for listing endpoints

prashanthpai:
...and add support to generate (very primitive) REST API doc listing.
This is only a stopgap solution.

Closes: https://github.com/gluster/glusterd2/issues/311

What's nicer would be to have support for generation of swagger/openapi
spec from the code comments and annotations. It's a tedious task to add
them to existing APIs and shall be done later.
See: https://goswagger.io/generate/spec.html

Usage:
```sh
$ curl -o endpoints.json -s -X GET http://127.0.0.1:24007/endpoints
$ go build pkg/tools/generate-doc.go
$ ./generate-doc
```


glusterd2/plugin/plugins.go
```diff
@@ -7,13 +7,12 @@ import (
...
 	"github.com/gluster/glusterd2/plugins/georeplication"
 	"github.com/gluster/glusterd2/plugins/glustershd"
-	"github.com/gluster/glusterd2/plugins/hello"
 	"github.com/gluster/glusterd2/plugins/quota"
 )
 
 // PluginsList is a list of plugins which implements GlusterdPlugin interface
 var PluginsList = []GlusterdPlugin{
-	&hello.Plugin{},
+	// &hello.Plugin{},
```
aravindavk: This plugin can be removed completely(Including directory). It was added as example.

prashanthpai:
> Is it possible to add more fields to "Route" so that it can be picked up here?

I did think of that. And yes it's possible. We currently iterate over [mux.Route](https://godoc.org/github.com/gorilla/mux#Route) which we can't add fields to. However, we can add it to [our version of Route](https://github.com/gluster/glusterd2/blob/master/glusterd2/servers/rest/route/route.go#L11) and iterate over that instead.

aravindavk:
@prashanthpai manual rebase required.

pkg/tools/generate-doc.go
```diff
@@ -0,0 +1,64 @@
...
+	"strings"
+	"text/template"
+
+	"github.com/gluster/glusterd2/pkg/api"
+
+	log "github.com/sirupsen/logrus"
+)
+
+const endpointsTable = `# API Endpoints
+
```
prashanthpai: TODO: I'll have to learn to nest templates here. I'll do it later. The generated markdown doc as lot of links but they aren't rendered, so it's not a problem.

prashanthpai:
@aravindavk Updated the PR and... 

* Removed hello plugin
* Added new fields to Route
* Added links to request and response struct in godoc

You can [see it rendered](https://github.com/prashanthpai/glusterd2/blob/21e3be2033d00dfded7af9b73b5bb56a491d8f84/endpoints.md)

I have added request and response structs for a few APIs and not all.

PR #570, opened 2018-02-15
Selective fast builds

kshlm:
This is required for building unbundled RPMs as an unprivileged user.
If not, the golang build process tries to install the built deps into the
system.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -42,8 +42,15 @@ else
     echo "Plugins Disabled"
 fi
 
+if [ "$FASTBUILD" == "yes" ];then
+  # Enable the `go build -i` flag to install dependencies during build and
+  # allow faster rebuilds of GD2.
+  INSTALLFLAG="-i"
+fi
+
+
 echo "Building $BIN $VERSION"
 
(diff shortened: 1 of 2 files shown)


scripts/build.sh
```diff
@@ -42,8 +42,15 @@ else
     echo "Plugins Disabled"
 fi
 
+if [ "$FASTBUILD" == "yes" ];then
+  # Enable the `go build -i` flag to install dependencies during build and
```
aravindavk: Is it allowed during build? My understanding is no network calls during build

kshlm: This flag installs the intermediate dependencies that `go build` builds into the `$GOPATH/pkg`. This doesn't actually download any packages.

Without this flag, on each build `go build` builds all intermediate packages, and discards them once the main package is built. This slows down subsequent builds which makes developement slower.

But when building RPMs, you build only once in a clean environment. So there is nothing to gain by trying to install the intermediate dependencies. Also, when building using system packaged dependencies, this flag will cause `go build` to try to install the built packages into system paths (`/usr/`), which fails when building as an unprivileged user.

aravindavk: got it.

PR #571, opened 2018-02-16
Fix few pending stuff in REST API doc generation

prashanthpai:
* REST API doc now resides in `doc/` dir and is mentioned in `README.md`
* Added request and response structs to peer APIs
* The `/statedump` and `/endpoints` now follow the normal convention of registering a mux route.
* Added some useful comments about further pending TODOs
* Add generated doc to `.gitattributes` to suppress GitHub showing diff.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+doc/endpoints.md linguist-generated=true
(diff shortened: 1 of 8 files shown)


prashanthpai:
The horizontal scroll bar in the rendered markdown page is a little irritating though. But a table is convenient for now. When the `.Description` field of routes get populated, we can get rid of the table and list APIs one after the other like we do in wiki.

PR #517, opened 2018-01-08
Adds event listener and related APIs to glusterd

nnDarshan:
Issue: https://github.com/gluster/glusterd2/issues/418
Signed-off-by: Darshan N <darshan.n.2024@gmail.com>

PR #577, opened 2018-02-21
Listen on UDP port 24007 for incoming events

prashanthpai:

### DIFF
--- a/glusterd2/servers/eventlistener/handler.go
+++ b/glusterd2/servers/eventlistener/handler.go
@@ -2,15 +2,17 @@ package eventlistener
 
 import (
 	"encoding/json"
-	"github.com/gluster/glusterd2/glusterd2/gdctx"
-	events "github.com/gluster/glusterd2/plugins/events"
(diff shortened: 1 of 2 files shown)


aravindavk:
We need to make that port configurable, I don't think disabling it is good approach.

prashanthpai:
> We need to make that port configurable, I don't think disabling it is good approach.

I'm not really in favour of introducing another config option. May be I need to understand the usecase better. Can we make it use port 24007 if possible ? IIRC, that was something we considered earlier.

I'm surprised that this wasn't caught or seen in e2e tests. I see stacktraces because of port clash but the process doesn't crash or error out gracefully. This PR is fix for just that, not for the original problem. Hence, called out as "temporary".

There are other leftovers to fix for events listener that was merged in https://github.com/gluster/glusterd2/pull/517 and this can be taken care whenever that's fixed.

prashanthpai:
@aravindavk Event listener will now listen on the configured `clientaddress` which defaults to `:24007`

PR #582, opened 2018-02-22
Default configuration for glusterd2 when run using `systemctl`

aravindavk:
systemctl enable glusterd2
    systemctl start glusterd2
    systemctl status glusterd2


kshlm:
When starting from systemd, GD2 should automatically pickup configuration from `/etc/glusterd2/glusterd2.toml`. `make install` automatically installs this file, and it is included in built rpms. So GD2 should just pick it up and start. No need to supply command line arguments.

This file is generated using the [`scripts/gen-gd2conf.sh`](https://github.com/gluster/glusterd2/blob/master/scripts/gen-gd2conf.sh) script. It currently contains just these 3 config options too. So if you need to change any config options, you just need to edit that.

brainfunked:
> When starting from systemd, GD2 should automatically pickup configuration from `/etc/glusterd2/glusterd2.toml`.

The [quick start user guide](https://github.com/gluster/glusterd2/blob/master/doc/quick-start-user-guide.md) presently states that the file name should be `glusterd.{yaml,toml,json}`, which seems to be a typo. GD2 automatically picked up my `glusterd2.yaml` file.

On a side note, if the file is present with more than one of the three formats, is there any order of priority?

aravindavk:
> This file is generated using the scripts/gen-gd2conf.sh script. It currently contains just these 3 config options too. So if you need to change any config options, you just need to edit that.

Ok, I will modify that file if required.

aravindavk:
> The quick start user guide presently states that the file name should be glusterd.{yaml,toml,json}, which seems to be a typo. GD2 automatically picked up my glusterd2.yaml file.

> On a side note, if the file is present with more than one of the three formats, is there any order of priority?

Thanks, we will work on it to fix Quick start guide and other documentation.

@kshlm, @prashanthpai based on our usage till now, yaml format is the one which is mostly used. We can limit the config file to this format to avoid confusions to users.

prashanthpai:
> @kshlm, @prashanthpai based on our usage till now, yaml format is the one which is mostly used

@kshlm had a preference for toml as yaml is sensitive to whitespace. Our support for the 3 formats come from support provided by viper.

aravindavk:
> @kshlm had a preference for toml as yaml is sensitive to whitespace. Our support for the 3 formats come from support provided by viper.

Ok, we will use toml itself. We can separately add config file name validation, before giving it to viper.

prashanthpai:
@aravindavk Wait! There's been some misunderstanding. I don't think anyone requested to remove support for other two file formats. They worked just fine. Why do we need to limit it to one format ? Almost all the code to support them are done by library. The issue reported by Mrugesh was about name of the conf file in the quick start guide.

aravindavk:
> @aravindavk Wait! There's been some misunderstanding. I don't think anyone requested to remove support for other two file formats. They worked just fine. Why do we need to limit it to one format ? Almost all the code to support them are done by library. The issue reported by Mrugesh was about name of the conf file in the quick start guide.

It is not about technical capability, limited for better usability. Having single config file type has many benefits.

- No confusion to users
- All documentation can refer to single format and documentation will be uniform.
- Workaround to users or debugging will be easy since only one supported format

Also as reported by Mrugesh, if we have "glusterd2.yaml", "glusterd2.toml" and "glusterd2.json" in the same directory which one will take the precedence? We can find this out by looking at viper's code base and document it. But I feel, it is adding more confusion than the benefit.

kshlm:
I'm with @aravindavk on this one. It's time we standardized on a single format. Keeps it simple for users.

scripts/gen-gd2conf.sh
```diff
@@ -19,4 +20,5 @@ cat >$OUTPUT <<EOF
 localstatedir = "$GD2STATEDIR"
 logdir = "$GD2LOGDIR"
 logfile = "$GD2.log"
+rundir = "$RUNDIR"
```
kshlm: The rundir needs to be GD2 specific, ie. `$RUNDIR/$GD2`. Better to setup $GD2RUNDIR and use it here.

aravindavk: Done

aravindavk:
@kshlm please review and merge

PR #583, opened 2018-02-22
Quota: to set quota options

harigowtham:
Various quota options we set as quota set command in glusterd one.
From glusterd2 onwards they are maintained through volume set commands.
this patch enables setting quota options like soft and hard timeout,
deemstatfs, alert-time and default-soft-limit.

The parsing for time during volume set was bug, have changed it to
accomodate the backward compatiblity with glusterd1's validation.

Updates: #421

### DIFF
--- a/glusterd2/xlator/options/options.go
+++ b/glusterd2/xlator/options/options.go
@@ -336,10 +336,46 @@ func ValidateStr(o *Option, val string) error {
 
 // ValidateTime validates if the option is valid time format
 func ValidateTime(o *Option, val string) error {
-	if validate.IsTime(val, "hh:mm:ss") != true {
-		return ErrInvalidArg
+	var time int    // to convert given value from other formates to seconds
+	var tStr string // temp value which has the int without "min" ...
+	multiplier := 1
+	if strings.HasSuffix(val, "sec") {
+		tStr = strings.TrimSuffix(val, "sec")
+	} else if strings.HasSuffix(val, "s") {
+		tStr = strings.TrimSuffix(val, "s")
+	} else if strings.HasSuffix(val, "min") {
+		tStr = strings.TrimSuffix(val, "min")
+		multiplier = 60
+	} else if strings.HasSuffix(val, "m") {
+		tStr = strings.TrimSuffix(val, "m")
+		multiplier = 60
+	} else if strings.HasSuffix(val, "hr") {
+		tStr = strings.TrimSuffix(val, "hr")
+		multiplier = 60 * 60
+	} else if strings.HasSuffix(val, "h") {
+		tStr = strings.TrimSuffix(val, "h")
+		multiplier = 60 * 60
+	} else if strings.HasSuffix(val, "days") {
+		tStr = strings.TrimSuffix(val, "days")
+		multiplier = 60 * 60 * 24
+	} else if strings.HasSuffix(val, "d") {
+		tStr = strings.TrimSuffix(val, "d")
+		multiplier = 60 * 60 * 24
+	} else if strings.HasSuffix(val, "w") {
+		tStr = strings.TrimSuffix(val, "w")
+		multiplier = 60 * 60 * 24 * 7
+	} else if strings.HasSuffix(val, "wk") {
+		tStr = strings.TrimSuffix(val, "wk")
+		multiplier = 60 * 60 * 24 * 7
+	} else {
+		tStr = val
 	}
-	return nil
+	time, err := strconv.Atoi(tStr)
+	if err != nil {
+		return err
+	}
+	time = time * multiplier
+	return ValidateRange(o, strconv.Itoa(time))
 }
 
 // ValidateXlator validates if the option is a valid xlator
--- a/glusterd2/xlator/validation.go
+++ b/glusterd2/xlator/validation.go
@@ -52,6 +52,32 @@ func validateBitrot(v *volume.Volinfo, key string, value string) error {
 	return nil
 }
 
+func validateQuota(v *volume.Volinfo, key string, value string) error {
+
+	// Check if quota is already enabled
+	if volume.IsQuotaEnabled(v) == false {
+		err := fmt.Errorf("Quota not enabled to set this value: '%s'", key)
+		return err
+	}
+
+	switch key {
+	case "deem-statfs":
+		return nil
+	case "hard-timeout":
+		return nil
+	case "soft-timeout":
+		return nil
+	case "alert-time":
+		return nil
+	case "default-soft-limit":
+		return nil
+	default:
+		err := fmt.Errorf("'%s' is not a valid quota option", key)
+		return err
+	}
+	return nil
+}
+
 func registerValidation(xlator string, vf validationFunc) error {
 	xl, err := Find(xlator)
 	if err != nil {
@@ -65,5 +91,11 @@ func registerAllValidations() error {
 	if err := registerValidation("afr", validateReplica); err != nil {
 		return err
 	}
-	return registerValidation("bit-rot", validateBitrot)
+	if err := registerValidation("bit-rot", validateBitrot); err != nil {
+		return err
+	}
+	if err := registerValidation("quota", validateQuota); err != nil {
+		return err
+	}
+	return nil
 }


harigowtham:
@sanoj-unnikrishnan @aravindavk @prashanthpai @kshlm  @atinmu @amarts can you people take a look at it?

glusterd2/xlator/validation.go
```diff
@@ -52,6 +52,29 @@ func validateBitrot(v *volume.Volinfo, key string, value string) error {
...
+		return nil
+	case "soft-timeout":
+		return nil
+	case "hard-timeout":
+		return nil
+	case "alert-time":
+		return nil
+	case "default-soft-limit":
+		return nil
+	}
```
prashanthpai: There should be a `case default` which returns error OR the last return outside the switch block must return error.

glusterd2/volume/struct.go
```diff
@@ -35,6 +35,16 @@ const (
 const (
 	// VkeyFeaturesQuota is the key to check if quota is disabled or enabled
 	VkeyFeaturesQuota = "features.quota"
+	// VkeyQuotaDeemStatfs is the key to disabled or enabled deem statfs
```
prashanthpai: These shouldn't be in `glusterd2/volume/struct.go` as these are xlator (quota) specific. I guess it can be moved during refactor later.

glusterd2/xlator/validation.go
```diff
@@ -52,6 +52,29 @@ func validateBitrot(v *volume.Volinfo, key string, value string) error {
...
 }
 
+func validateQuota(v *volume.Volinfo, key string, value string) error {
+
+	// Check if quota is already enabled
+	if volume.IsQuotaEnabled(v) == false {
+		err := fmt.Errorf("Quota not enabled to set this value: '%s'", key)
+		return err
+	}
+
```
prashanthpai: What exactly is this function for ? Are these values supposed to be allowed to change when quota is enabled or when it's disabled ? `validateQuota` is always called for any volume set option which has `quota` prefix in it.

glusterd2/xlator/options/options.go
```diff
@@ -324,10 +324,84 @@ func ValidateStr(o *Option, val string) error {
...
+		}
+		time = time * 60 * 60 * 24 * 7
+	} else if strings.HasSuffix(val, "wk") {
+		t = strings.TrimSuffix(val, "wk")
+		time, err = strconv.Atoi(t)
+		if err != nil {
+			return err
+		}
+		time = time * 60 * 60 * 24 * 7
+	} else {
```
prashanthpai: If there's no valid suffix, is it an error ?

glusterd2/xlator/options/options.go
```diff
@@ -324,10 +324,84 @@ func ValidateStr(o *Option, val string) error {
...
 // ValidateTime validates if the option is valid time format
 func ValidateTime(o *Option, val string) error {
-	if validate.IsTime(val, "hh:mm:ss") != true {
-		return ErrInvalidArg
+	var time int
+	var t string
+	var err error
+	if strings.HasSuffix(val, "sec") {
+		value := strings.TrimSuffix(val, "sec")
+		err = ValidateRange(o, value)
```
prashanthpai: `ValidateRange` is already being called at the end of the if block.

glusterd2/xlator/options/options.go
```diff
@@ -324,10 +324,84 @@ func ValidateStr(o *Option, val string) error {
...
+		if err != nil {
+			return err
+		}
+	} else if strings.HasSuffix(val, "s") {
+		value := strings.TrimSuffix(val, "s")
+		err = ValidateRange(o, value)
```
prashanthpai: Is `strconv.Atoi` not needed here ?

Also, `ValidateRange` also has a call `strconv.ParseFloat`

glusterd2/xlator/options/options.go
```diff
@@ -324,10 +324,84 @@ func ValidateStr(o *Option, val string) error {
 
 // ValidateTime validates if the option is valid time format
 func ValidateTime(o *Option, val string) error {
-	if validate.IsTime(val, "hh:mm:ss") != true {
```
aravindavk: Is this condition not required in the end? (else part)

glusterd2/xlator/validation.go
```diff
@@ -52,6 +52,29 @@ func validateBitrot(v *volume.Volinfo, key string, value string) error {
...
+	switch key {
```
aravindavk: Use the constants defined above.

    key = "quota." + key
    switch key{
    case VkeyQuotaDeemStatfs:

glusterd2/xlator/validation.go
harigowtham: done

glusterd2/volume/struct.go
harigowtham: If this is only for volume options where do I put the quota options. 
I searched for the right place to put it couldn't find it. 
I can see that bitrot has too put it here. 
Do I have to use the string as such everywhere or is there a place to put it which I'm not able to find?

glusterd2/xlator/validation.go
harigowtham: Yes. All these values can be set only if quota is enabled.

glusterd2/xlator/options/options.go
harigowtham: the ValidateRange called at the end of the if else block is for the input cases with input values that need conversion. The validate range takes a string as an input, If the input from the command is just a number it is considered as a string and passed over to ValidateRange. If it has a string like "s", "hr" and so on, these have to be trimmed off. And if the value is given in any format other than seconds, then it has to be converted. For conversation, we have to convert it from string to int and then to pass it to validarange we need to parse it back to string. that why all this confusion. 

Again the answer for this comment is, the ValidRange will be called for all the input value which needed a conversion. This includes all formats other than "50", "50s", and "50sec"

glusterd2/xlator/options/options.go
harigowtham: no. If the sufix is not given we are supposed to default it to seconds.

glusterd2/xlator/options/options.go
harigowtham: The valid is designed to take input argument as a string. If changing it to int is fine and wont break compatibility for other commands, then the validateRange's arguments have to be changed.

glusterd2/xlator/options/options.go
harigowtham: the glusterd one the "xlator_option_validate_time" is the common validation function for time format and in that the "gf_string2time" doesn't have a parsing for "hh:mm:ss". 

We can add it if we want to suport it from GD2.

glusterd2/xlator/validation.go
harigowtham: As @prashanthpai  wants to remove the VkeyQuota.. options this is contradicting, need to resolve this confusion. and then will make changes as per it.

glusterd2/volume/struct.go
prashanthpai: > I guess it can be moved during refactor later.

We'll move it later. For now, let it be.

glusterd2/xlator/validation.go
prashanthpai: > Yes. All these values can be set only if quota is enabled.

What about other quota keys ?

glusterd2/volume/struct.go
prashanthpai: On second thought, you can move these definitions to `./plugins/quota/consts.go`. In future, we'll find a better place for these validation functions itself.

glusterd2/xlator/options/options.go
prashanthpai: The `else` case applies to both: 
* suffix not given and
* ANY other suffix like say `y`

But that's okay as `ValidateRange` shall fail.

glusterd2/xlator/options/options.go
prashanthpai: I'm sorry, I should've been clearer. I was hinting at taking common code out of the case blocks wherever possible. Example:

```go
var tStr string
multiplier := 1

if strings.HasSuffix(val, "sec") {
    tStr = strings.TrimSuffix(val, "sec")
}
...
else if strings.HasSuffix(val, "min") {
    tStr = strings.TrimSuffix(val, "min")
    multiplier = 60
}
...
else {
    tStr = value
}

time, err := strconv.Atoi(tStr)
if err != nil {
    return err
}
time = time * multiplier
return ValidateRange(o, strconv.Itoa(time))
```

glusterd2/volume/struct.go
harigowtham: done

glusterd2/xlator/validation.go
harigowtham: The other quota keys like limit set, remove limit, etc.. will be added here when we work on those patches and send the PR.

glusterd2/xlator/options/options.go
harigowtham: okay

glusterd2/xlator/options/options.go
harigowtham: done

glusterd2/xlator/validation.go
harigowtham: done

glusterd2/xlator/validation.go
```diff
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/gluster/glusterd2/glusterd2/volume"
+	quota "github.com/gluster/glusterd2/plugins/quota"
```
prashanthpai: nit: You don't need the alias here. You can use the import directly.

glusterd2/xlator/validation.go
```diff
@@ -52,6 +53,34 @@ func validateBitrot(v *volume.Volinfo, key string, value string) error {
...
 
+func validateQuota(v *volume.Volinfo, key string, value string) error {
+
+	// Check if quota is already enabled
+	if volume.IsQuotaEnabled(v) == false {
+		err := fmt.Errorf("Quota not enabled to set this value: '%s'", key)
+		return err
+	}
+
+	key = "quota." + key
```
prashanthpai: As the constants already have the quota prefix, key is gonna be `quota.quota.deem-statfs`. You'll have to remove this statement.

glusterd2/xlator/validation.go
harigowtham: done

glusterd2/xlator/validation.go
harigowtham: done

harigowtham: There was a confusion here. The key we get here is deem-statfs not quota.deem-statfs. having the quota prefix added is necessary!

prashanthpai: I see. In that case, **if** those constants you defined in `plugins/quota/consts.go` are not used anywhere else in gd2, use them directly here and remove that file altogether. This is along the lines of bitrot validation function.

glusterd2/xlator/options/options.go
```diff
@@ -336,10 +336,46 @@ func ValidateStr(o *Option, val string) error {
...
-		return ErrInvalidArg
+	var time int    // to convert given value from other formates to seconds
```
prashanthpai: typo: formats

PR #589, opened 2018-03-01
Handle peer connects/disconnects in store

prashanthpai:
**Watch for liveness of nodes in store:**
     Log and emit local event every time a node connects to or disconnects
     from the store.

**Explicitly delete liveness key when an instance shuts down:**
    The default etcd session TTL is 30 seconds which is quite high.
    When a glusterd2 instance is shutdown, it takes a while for other
    nodes to notice this as TTL for the liveness key should expire. When
    a glusterd2 instance is shutdown gracefully (SIGTERM), it can simply
    delete its liveness key from the store and hence other nodes can see
    this instantly.

### DIFF
--- a/glusterd2/events/events.go
+++ b/glusterd2/events/events.go
@@ -58,11 +58,13 @@ func Start() error {
 	StartGlobal()
 	startEventLogger()
 	registerGaneshaHandler()
+	startLivenessWatcher()
 	return nil
 }
 
 // Stop stops the events framework, events will no longer be broadcast
 func Stop() error {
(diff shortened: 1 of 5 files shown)


kshlm:
retest this please

kshlm:
The CI failures here and elsewhere is caused by the change https://review.gluster.org/19466 which introduced a new field in the options struct, but added it into the middle. This causes GD2 to segfault when it tries to read options when running against builds that contain this change. The CI uses nightly-master builds, which has this change.

Issue #594, opened 2018-03-08
[RFE] CLI: External subcommands support


aravindavk:
External sub command support need to be added just like `git` extensions. If we create a executable with name `git-hello` then it will be available as git subcommand(`git hello`). This feature allows users to create/extend Gluster tools using Rich REST APIs.

http://lists.gluster.org/pipermail/gluster-users/2016-April/026301.html

aravindavk:
This is good to have feature, but requires more effort to achieve it securely. Closing this issue for now.

PR #595, opened 2018-03-08
cli: Support for external sub commands

aravindavk:
Added support for external sub commands to Gluster CLI.(Similar to
`git` external sub command support)

If any executable available in `PATH` with name `glustercli-<cmd>` then
it will be available as `glustercli <cmd>`. External sub command can be
written in any programming language.

All the known commands will get precedence over external commands.

Following arguments will be included while executing external commands.

    --xml
    --json
    --glusterd-host <host>
    --glusterd-https
    --glusterd-port <port>
    --logdir <path>
    --logfile <name>
    --loglevel <level>
    --cacert <path>
    --insecure

Example of simple external plugin/command, which connects to Glusterd
and just displays Volume name and State. (File name
`/usr/local/bin/glustercli-volinfo2` with executable permission)

    #!/usr/bin/env python
    from argparse import ArgumentParser
    import sys
    import json

    import requests

    if __name__ == "__main__":
        parser = ArgumentParser()
        parser.add_argument("--xml", action="store_true")
        parser.add_argument("--json", action="store_true")
        parser.add_argument("--glusterd-host", default="localhost")
        parser.add_argument("--glusterd-https", action="store_true")
        parser.add_argument("--glusterd-port", default=24007, type=int)
        parser.add_argument("--logdir")
        parser.add_argument("--logfile")
        parser.add_argument("--loglevel")
        parser.add_argument("--cacert")
        parser.add_argument("--insecure", action="store_true")

        parser.add_argument("--volume")

        args = parser.parse_args()
        url = "https://" if args.glusterd_https else "http://"
        url += "%s:%d/v1/volumes" % (args.glusterd_host, args.glusterd_port)
        if args.volume:
            url += "/" + args.volume

        resp = requests.get(url)

        if resp.status_code != 200:
            print("Unable to get Volume Info")
            sys.exit(1)

        for v in json.loads(resp.content):
            print("%15s %s" % (v["name"], v["state"]))

Fixes: #594

### DIFF
--- a/glustercli/cmd/common.go
+++ b/glustercli/cmd/common.go
@@ -41,13 +41,13 @@ func handleGlusterdConnectFailure(msg string, err error, https bool, host string
 			scheme = "https"
 		}
 		os.Stderr.WriteString(msg + "\n\n")
-		os.Stderr.WriteString(fmt.Sprintf(errFailedToConnectToGlusterd, scheme, flagHostname, flagPort))
+		os.Stderr.WriteString(fmt.Sprintf(errFailedToConnectToGlusterd, scheme, FlagHostname, FlagPort))
 		os.Exit(errcode)
 	}
 }
 
 func failure(msg string, err error, errcode int) {
-	handleGlusterdConnectFailure(msg, err, flagHTTPS, flagHostname, flagPort, errcode)
+	handleGlusterdConnectFailure(msg, err, FlagHTTPS, FlagHostname, FlagPort, errcode)
 
 	// If different error
 	os.Stderr.WriteString(msg + "\n")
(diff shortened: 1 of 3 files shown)


brainfunked:
What use is the CLI expected to be run as? I see a potential security implication here where arbitrary code could be deployed and run as a glustercli subcommand.

aravindavk:
> What system user is the CLI expected to be run as? I see a potential security implication here where arbitrary code could be deployed and run as a glustercli subcommand.

Makes sense if I execute binary if available in PATH. Thanks for pointing that. I will modify the patch to look for a particular directory for sub commands, so that directory can be protected and only admins can deploy the sub commands.

Gluster CLI does not require root user to run the commands, since it is based on REST API Authentication. Subcommands also executed as the same user as Gluster CLI is executed. I think we can give warning(or restrict) if `glustercli` is run as root.

brainfunked:
> I think we can give warning(or restrict) if glustercli is run as root.

Another way would be to warn and then drop to an unprivileged user such as `nobody`.

aravindavk:
Abandoning this patch for now since many changes required to make it work securely. I will work on this when this feature is really needed.

PR #600, opened 2018-03-13
Edit Peer functionality

rishubhjain:
W.r.t https://github.com/gluster/glusterd2/issues/416, editing metadata of peer.
Need some discussions on this before getting it reviewed


glusterd2/commands/peers/editmetadata.go
```diff
@@ -0,0 +1,44 @@
...
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "peerid not present in request", api.ErrCodeDefault)
+		return
+	}
+	peerInfo, err := peer.GetPeer(peerID)
+	if err != nil {
+		logger.WithError(err).WithField("peerid", peerID).Error("Peer ID not found in store")
+		restutils.SendHTTPError(ctx, w, http.StatusNotFound, "Peer Id not found in store", api.ErrCodeDefault)
+		return
+	}
+	peerInfo.MetaData = req.MetaData
```
aravindavk: Overwrite only what is set.

    for k, v := range req.MetaData {
        peerInfo.MetaData[k] = v
    }

pkg/api/peer_req_resp.go
```diff
@@ -18,6 +18,11 @@ type PeerAddReq struct {
 	MetaData  map[string]string `json:"metadata"`
 }
 
+type MetaDataEdit struct {
+	PeerID   uuid.UUID         `json:"peerid"`
```
prashanthpai: The peer id is an existing resource. This shall be part of the URL, not in the body.

I also see an opportunity to allow editing of any field of Peer object (such as name and addresses), not just metadata. There's a [related issue](https://github.com/gluster/glusterd2/issues/356)

glusterd2/commands/peers/commands.go
```diff
@@ -46,6 +46,15 @@ func (c *Command) Routes() route.Routes {
 			ResponseType: utils.GetTypeString((*api.PeerAddResp)(nil)),
 			HandlerFunc:  addPeerHandler,
 		},
+		route.Route{
+			Name:         "EditMetaData",
+			Method:       "POST",
+			Pattern:      "/peers/{peerid}/metadata",
+			Version:      1,
+			RequestType:  utils.GetTypeString((*api.MetaDataEdit)(nil)),
+			ResponseType: utils.GetTypeString((*api.PeerAddResp)(nil)),
```
prashanthpai: Don't use `PeerAddResp`. Create and use a separate alias.

```go
type PeerEditResp Peer
```

glusterd2/peer/self.go
```diff
@@ -46,7 +46,8 @@ func AddSelfDetails() error {
 		Name:          gdctx.HostName,
 		PeerAddresses: []string{config.GetString("peeraddress")},
 	}
-
+	p.MetaData = make(map[string]string)
+	p.MetaData["group"] = "1"
```
prashanthpai: The value `1` seems very arbitrary. You don't need to set a default group at all. Unless metadata is explicitly is set by the user (or a plugin/program), it shouldn't exist.

glusterd2/commands/peers/editmetadata.go
```diff
@@ -0,0 +1,44 @@
...
+	if err := restutils.UnmarshalRequest(r, &req); err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, err.Error(), api.ErrCodeDefault)
+		return
+	}
+	peerID := mux.Vars(r)["peerid"]
+	if peerID == "" {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "peerid not present in request", api.ErrCodeDefault)
+		return
+	}
+	peerInfo, err := peer.GetPeer(peerID)
```
prashanthpai: Editing of peer object must be done under locks.

glusterd2/commands/peers/editmetadata.go
```diff
@@ -0,0 +1,116 @@
...
+	if err := restutils.UnmarshalRequest(r, &req); err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, err.Error(), api.ErrCodeDefault)
+		return
+	}
+	peerID := mux.Vars(r)["peerid"]
+	if peerID == "" {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "Peer ID not present in request", api.ErrCodeDefault)
+		return
+	}
+
```
aravindavk: Return error if metadata key starts with `_`

    for k, _ := range req.MetaData{
        if strings.HasPrefix(k, "_") {
               // Error: Restricted metadata
        }
    }

pkg/api/peer_req_resp.go
```diff
@@ -18,9 +18,17 @@ type PeerAddReq struct {
 	MetaData  map[string]string `json:"metadata"`
 }
 
+// MetaDataEditReq represents an incoming request to edit metadata of peer
+type MetaDataEditReq struct {
```
prashanthpai: This should be `PeerEditReq`. The same API can be used to update/edit peer later.

pkg/api/peer_req_resp.go
```diff
@@ -18,9 +18,17 @@ type PeerAddReq struct {
...
+	MetaData map[string]string `json:"metadata"`
+}
+
 // PeerAddResp is the success response sent to a PeerAddReq request
 type PeerAddResp Peer
 
+// MetaDataEditResp is the success response sent to a MetaDataEditReq request
+type MetaDataEditResp Peer
```
prashanthpai: PeerEditResp

glusterd2/commands/peers/commands.go
```diff
@@ -46,10 +46,19 @@ func (c *Command) Routes() route.Routes {
...
+			ResponseType: utils.GetTypeString((*api.MetaDataEditResp)(nil)),
+			HandlerFunc:  editMetaData,
+		},
 	}
 }
 
 // RegisterStepFuncs implements a required function for the Command interface
 func (c *Command) RegisterStepFuncs() {
-	return
+	registerEditMetaDataStepFuncs()
```
prashanthpai: registerEditPeerStepFuncs

glusterd2/commands/peers/commands.go
```diff
@@ -46,10 +46,19 @@ func (c *Command) Routes() route.Routes {
...
 			HandlerFunc:  addPeerHandler,
 		},
+		route.Route{
+			Name:         "EditMetaData",
+			Method:       "POST",
+			Pattern:      "/peers/{peerid}/metadata",
+			Version:      1,
+			RequestType:  utils.GetTypeString((*api.MetaDataEditReq)(nil)),
+			ResponseType: utils.GetTypeString((*api.MetaDataEditResp)(nil)),
+			HandlerFunc:  editMetaData,
```
prashanthpai: editPeer

glusterd2/commands/peers/editmetadata.go
```diff
@@ -0,0 +1,116 @@
...
+	ctx := r.Context()
+	logger := gdctx.GetReqLogger(ctx)
+
+	var req api.MetaDataEditReq
+	if err := restutils.UnmarshalRequest(r, &req); err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, err.Error(), api.ErrCodeDefault)
+		return
+	}
+	peerID := mux.Vars(r)["peerid"]
```
prashanthpai: Can you check if this code will be hit at all ? Wouldn't mux filter it out as it doesn't fit the pattern ?

glusterd2/commands/peers/editmetadata.go
```diff
@@ -0,0 +1,116 @@
...
+	if peerID == "" {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "Peer ID not present in request", api.ErrCodeDefault)
+		return
+	}
+
+	txn := transaction.NewTxn(ctx)
+	defer txn.Cleanup()
+	lock, unlock, err := transaction.CreateLockSteps(string(peerID))
```
prashanthpai: `peerID` is already a string.

glusterd2/commands/peers/editmetadata.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		logger.WithError(err).WithField("peerid", peerID).Error("Failed to get lock on peer")
```
prashanthpai: Incorrect log message (and returned error). You're not taking lock. You are obtaining lock and unlock functions.

glusterd2/commands/peers/editmetadata.go
```diff
@@ -0,0 +1,116 @@
...
+
+	txn.Steps = []*transaction.Step{
+		lock,
+		{
+			DoFunc: "edit-metadata",
+			Nodes:  []uuid.UUID{gdctx.MyUUID},
+		},
+		unlock,
+	}
+	err = txn.Ctx.Set("peerid", string(peerID))
```
prashanthpai: It's already a string.

glusterd2/commands/peers/editmetadata.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		logger.WithError(err).Error("Failed to set peerID data for transaction")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err.Error(), api.ErrCodeDefault)
+		return
+	}
+	err = txn.Ctx.Set("req", req)
+	if err != nil {
+		logger.WithError(err).Error("Failed to set Req data for transaction")
```
prashanthpai: A better error message will be `Failed to set request in transaction context`. Same comment for other calls to `txn.Ctx.Set`

glusterd2/commands/peers/editmetadata.go
```diff
@@ -0,0 +1,116 @@
...
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err.Error(), api.ErrCodeDefault)
+		return
+	}
+	err = txn.Do()
+	if err != nil {
+		logger.WithError(err).Error("Transaction to update metadata failed")
```
prashanthpai: Add "peer" throughout your PR when you refer to metadata. Metadata can also belong to a volume.

glusterd2/commands/peers/editmetadata.go
```diff
@@ -0,0 +1,116 @@
...
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, "Transaction to update metadata failed", api.ErrCodeDefault)
+		return
+	}
+	peerInfo, err := peer.GetPeer(peerID)
```
prashanthpai: There's a race here as it's not done under lock. Theoretically, the peer may have been deleted from the point when you finish the transaction and get here.

glusterd2/commands/peers/editmetadata.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		logger.WithError(err).WithField("peerid", peerID).Error("Peer ID not found in store")
+		restutils.SendHTTPError(ctx, w, http.StatusNotFound, "Peer Id not found in store", api.ErrCodeDefault)
+		return
+	}
+	resp := createPeerAddResp(peerInfo)
```
prashanthpai: As pointed out in earlier review comments, this isn't add peer. It can be `createPeerEditResp`

glusterd2/commands/peers/editmetadata.go
```diff
@@ -0,0 +1,116 @@
...
+	restutils.SendHTTPResponse(ctx, w, http.StatusCreated, resp)
+
+}
+
+func txnEditMetaData(c transaction.TxnCtx) error {
+	var peerID string
```
prashanthpai: Declaring variables together with where they are used is a good convention to follow when you get/unmarshall something and IMO it improves code readability.

```go
var peerID string
if err := c.Get("peerid", &peerID); err != nil {
...
}

var req api.MetaDataEditReq
if err := c.Get("req", &req); err != nil {
...
}
```

glusterd2/commands/peers/editmetadata.go
```diff
@@ -0,0 +1,116 @@
...
+	peerInfo, err := peer.GetPeer(peerID)
+	if err != nil {
+		c.Logger().WithError(err).WithField("peerid", peerID).Error("Peer ID not found in store")
+		return err
+	}
+
+	for k, v := range req.MetaData {
+		peerInfo.MetaData[k] = v
+	}
+	err = peer.AddOrUpdatePeer(peerInfo)
```
prashanthpai: You can set this new peerInfo in transaction context and retrieve it in the handler. This will handle the theoretical race.

glusterd2/commands/peers/editmetadata.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	return nil
+}
+
+func registerEditMetaDataStepFuncs() {
+	var sfs = []struct {
+		name string
+		sf   transaction.StepFunc
+	}{
+		{"edit-metadata", txnEditMetaData},
```
prashanthpai: prefix "peer" wherever applicable.

glusterd2/commands/peers/editmetadata.go
rishubhjain: @prashanthpai So would you suggest me a method to address this race condition.

glusterd2/commands/peers/editpeer.go
```diff
@@ -0,0 +1,144 @@
...
+	logger := gdctx.GetReqLogger(ctx)
+
+	var req api.PeerEditReq
+	if err := restutils.UnmarshalRequest(r, &req); err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, err.Error(), api.ErrCodeDefault)
+		return
+	}
+	for k := range req.MetaData {
+		if strings.HasPrefix(k, "_") {
+			logger.WithField("Metadata field", req.MetaData).Error("Restricted key passed in Metadata Field")
```
prashanthpai: Fields are meant to be parsed by utilities that parse logs. I'd suggest to name the key name as `key` or `metadata-key`.

glusterd2/commands/peers/editpeer.go
```diff
@@ -0,0 +1,144 @@
...
+			restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, "Restricted Metadata keys used in Metadata field", api.ErrCodeDefault)
```
prashanthpai: A more useful error message to the client would be to include a reference to underscores being disallowed.

glusterd2/commands/peers/editpeer.go
```diff
@@ -0,0 +1,144 @@
...
+			return
+		}
+	}
+	peerID := mux.Vars(r)["peerid"]
+	if peerID == "" {
```
prashanthpai: Please use `uuid.Parse` to validate. This check isn't good enough.

glusterd2/commands/peers/editpeer.go
```diff
@@ -0,0 +1,144 @@
...
+
+	txn.Steps = []*transaction.Step{
+		lock,
+		{
+			DoFunc: "edit-peer",
+			Nodes:  []uuid.UUID{gdctx.MyUUID},
+		},
+		unlock,
+	}
+	err = txn.Ctx.Set("peerid", string(peerID))
```
prashanthpai: peerID is already a string. There is no need for type conversion here.

glusterd2/commands/peers/editpeer.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		logger.WithError(err).Error("Failed to set peerID data for transaction")
```
prashanthpai: A better log message is to end it with - "in transaction context" instead of "for transaction". Follow the same at other places too.

glusterd2/commands/peers/editpeer.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	return nil
+}
+
+func registerPeerEditStepFuncs() {
+	var sfs = []struct {
+		name string
+		sf   transaction.StepFunc
+	}{
+		{"edit-peer", txnEditPeer},
```
prashanthpai: Just to be consistent with other naming conventions, please name this as `peer-edit`

glusterd2/peer/self.go
```diff
@@ -46,7 +46,6 @@ func AddSelfDetails() error {
 		Name:          gdctx.HostName,
 		PeerAddresses: []string{config.GetString("peeraddress")},
 	}
-
```
prashanthpai: This is unintended change and shows up in `git log` and later in `git blame`. Avoid these.

glusterd2/commands/peers/editpeer.go
rishubhjain: @prashanthpai Is this comment for logger.WithField() method?, am I required to pass key or metada-key as the key-name in the WithField function?

rishubhjain:
@prashanthpai Please review.

glusterd2/commands/peers/editpeer.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	err = txn.Ctx.Set("peerid", peerID)
+	if err != nil {
+		logger.WithError(err).WithField("peerid", peerID).Error("Failed to set peerID in transaction context")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err.Error())
+		return
+	}
+	err = txn.Ctx.Set("req", req)
+	if err != nil {
+		logger.WithError(err).WithField("req-key", req).Error("Failed to set unmarshalled request information in transaction context")
```
prashanthpai: The variable `req` is a structure and not a string!

Change it as follows:
```go
// for set
logger.WithError(err).WithField("key", "req").Error("Failed to set key in transaction context")

// for get
logger.WithError(err).WithField("key", "req").Error("Failed to get key from transaction context")
```

Follow the same **everywhere else** wherever you call `txn.Set` or `txn.Get`. You don't have to specify the key in the error message. It's easier for it to be field. Error message won't change.

rishubhjain: @prashanthpai Is this for req only or should I do what you mentioned for peerid aswell

prashanthpai: Everywhere including `peer-id`

rishubhjain: @prashanthpai I am changing the error message, but  I am keeping the key for peer-id as "peer-id" as we discussed in office,

glusterd2/commands/peers/editpeer.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	resp := createPeerEditResp(&peerInfo)
+	restutils.SendHTTPResponse(ctx, w, http.StatusCreated, resp)
+
+}
+
+func txnPeerEdit(c transaction.TxnCtx) error {
+	var peerID string
+	if err := c.Get("peerid", &peerID); err != nil {
+		c.Logger().WithError(err).WithField("key", peerID).Error("Failed to get key from transaction context")
```
prashanthpai: This must be "peerid" in the value of field. The variable `peerID` will be empty string in case of failure.

PR #613, opened 2018-03-21
Simplify SendHTTPError()

prashanthpai:
* Remove having to specify default error code. It is made implicit
  instead. This removes some of the code clutter.
* Support returning multiple errors in a single API response.
  Although this will be rarely used, introducing the provision in API
  right now is better than later.
* Allow callers to pass both string and error types.
* Introduce a error code map which maps error codes (integers) to their
  brief textual description.

### DIFF
--- a/glusterd2/servers/rest/utils/utils.go
+++ b/glusterd2/servers/rest/utils/utils.go
@@ -4,17 +4,12 @@ package utils
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 
 	"github.com/gluster/glusterd2/glusterd2/gdctx"
 	"github.com/gluster/glusterd2/pkg/api"
 )
-
-// APIError is the placeholder for error string to report back to the client
-type APIError struct {
-	Code  api.ErrorCode `json:"error_code"`
-	Error string        `json:"error"`
-}
 
 // UnmarshalRequest unmarshals JSON in `r` into `v`
 func UnmarshalRequest(r *http.Request, v interface{}) error {
@@ -44,8 +39,11 @@ func SendHTTPResponse(ctx context.Context, w http.ResponseWriter, statusCode int
 	}
 }
 
-// SendHTTPError sends an error response to the client.
-func SendHTTPError(ctx context.Context, w http.ResponseWriter, statusCode int, errMsg string, errCode api.ErrorCode) {
+// SendHTTPError sends an error response to the client. The caller of this
+// function can pass either the error or one or more error code(s) exported by
+// api package.
+func SendHTTPError(ctx context.Context, w http.ResponseWriter, statusCode int,
+	err interface{}, errCodes ...api.ErrorCode) {
 
 	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
 	w.Header().Set("X-Gluster-Node-Id", gdctx.MyUUID.String())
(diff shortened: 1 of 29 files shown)


prashanthpai:
**Note to reviewers:**

The core code changes are in these files:
`glusterd2/servers/rest/utils/utils.go`
`pkg/api/errors.go`
`pkg/restclient/common.go`

Rest of the changes are at the callers end.

glusterd2/servers/rest/utils/utils.go
```diff
@@ -44,16 +39,32 @@ func SendHTTPResponse(ctx context.Context, w http.ResponseWriter, statusCode int
...
 	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
 	w.Header().Set("X-Gluster-Node-Id", gdctx.MyUUID.String())
 	w.Header().Set("X-Gluster-Cluster-Id", gdctx.MyClusterID.String())
 
 	w.WriteHeader(statusCode)
 
-	resp := APIError{Code: errCode, Error: errMsg}
+	var resp api.ErrorResp
+	errMsg := fmt.Sprint(err)
+	if errMsg != "" || len(errCodes) == 0 {
```
aravindavk: `errMsg != ""`  can be removed from the condition. If error code is sent along with error message then this condition will ignore error code. Also need a way to include actual error message in the response in case of custom error.

Example, adding custom error code for already started error helps caller to safely ignore the error messages.

    restutils.SendHTTPError(ctx, w, http.StatusBadRequest, errors.ErrVolAlreadyStarted, api.ErrCodeEexist)

prashanthpai: > If error code is sent along with error message then this condition will ignore error code

That was intentional. If it's a custom error message, the code will be set to default/generic. Your example can be re-written as:

```go
restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "", api.ErrVolAlreadyStarted)
or
restutils.SendHTTPError(ctx, w, http.StatusBadRequest, errors.ErrVolAlreadyStarted)

```

But I see your point. You are referring to a whole class of errors that can be safely ignored by client based on the error code.


> Also need a way to include actual error message in the response in case of custom error

The error response API has only two fields - `Code` and `Message`. It's good for the code and message to match and be deterministic. The convention I'm following is along [these lines](https://developer.twitter.com/en/docs/basics/response-codes) (see `Error Code` section)

Which of these would you prefer ? 

* Convert custom message and append to the errors list as `<generic code, error message>`. The API supports returning multiple `<code, message>` in it.
* Add an additional field to response to add actual error message.
* Concatenate and put custom error message it in the `Message` field itself.

aravindavk: `restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "", api.ErrVolAlreadyStarted)` looks good to me. All custom messages are forced to define in errors list if they need custom code.

pkg/api/errors.go
```diff
@@ -1,14 +1,26 @@
...
 }
 
 // ErrorCode represents API Error code Type
 type ErrorCode uint16
 
 const (
-	// ErrCodeDefault represents default error code for API responses
-	ErrCodeDefault ErrorCode = iota + 1
+	// ErrCodeGeneric represents generic error code for API responses
+	ErrCodeGeneric ErrorCode = iota
```
aravindavk: Starting with 0 may cause confusion since Return code 0 is success.

PR #615, opened 2018-03-23
quota: disable quotad

harigowtham:
disable quota daemon from api.
cli yet to be worked on.

Updates: #421


harigowtham:
@prashanthpai @aravindavk @kshlm @sanoj-unnikrishnan. pinging to add as reviewers.

plugins/quota/rest.go
```diff
@@ -40,6 +40,21 @@ func QuotadStart(c transaction.TxnCtx) error {
...
 }
 
+//QuotadStop is the function to stop the quota daemon
+func QuotadStop(c transaction.TxnCtx) error {
+	quotadDaemon, err := NewQuotad()
+	if err != nil {
+		return err
+	}
+
+	err = daemon.Stop(quotadDaemon, true)
```
aravindavk: Quotad is a Cluster level process, so it should not be stopped if any other Volume in the same node is using Quota.

    if number_of_quota_enabled_volumes_running > 0 {
        // Do not stop, Notify the running Quotad(Restart if Restart required)
    } else {
        // Stop Quotad
    }

plugins/quota/rest.go
```diff
@@ -115,9 +130,77 @@ func quotaEnableHandler(w http.ResponseWriter, r *http.Request) {
...
+	volName := p["volname"]
+
 	ctx := r.Context()
-	restutils.SendHTTPResponse(ctx, w, http.StatusOK, "todo: quota disable")
+	logger := gdctx.GetReqLogger(ctx)
+
+	// Validate volume existence
+	vol, err := volume.GetVolume(volName)
+	if err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusNotFound, errors.ErrVolNotFound.Error(), api.ErrCodeDefault)
```
rishubhjain: @harigowtham api.ErrCodeDefault is no more needed, Please see changes made in https://github.com/gluster/glusterd2/pull/613/files

plugins/quota/rest.go
harigowtham: Done. Got the restart part (for quotad on start for second volume or stop on only one of the volume) in with this change.

plugins/quota/rest.go
harigowtham: done

plugins/quota/rest.go
```diff
@@ -114,9 +169,81 @@ func quotaEnableHandler(w http.ResponseWriter, r *http.Request) {
...
 func quotaDisableHandler(w http.ResponseWriter, r *http.Request) {
-	// implement the help logic and send response back as below
+	// Collect inputs from URL
+	p := mux.Vars(r)
+	volName := p["volname"]
+
 	ctx := r.Context()
-	restutils.SendHTTPResponse(ctx, w, http.StatusOK, "todo: quota disable")
+	logger := gdctx.GetReqLogger(ctx)
+
```
prashanthpai: Lock must be held here, instead of in the txn step. Any call to the store such as below one must be under lock. See `glusterd2/commands/volumes/volume-start.go` for example.

harigowtham: done.

prashanthpai: I don't see it yet.

harigowtham: I did force push it. I'll check it for else were. In the mean time I'll push it again.

harigowtham: Done. have pushed it.

plugins/quota/rest.go
```diff
@@ -13,8 +13,33 @@ import (
...
+// IsQuotaAffectedNode returns true if the volume exists on that node and has quota enabled
+func IsQuotaAffectedNode() bool {
+	volumes, e := volume.GetVolumes()
+	if e != nil {
+		log.WithError(e).Error("Failed to get volumes")
+		return true
+	}
+	for _, v := range volumes {
+		val, exists := v.Options[volume.VkeyFeaturesQuota]
+		if exists && val == "off" {
```
aravindavk: Incomplete condition: `if !exists` is not considered. 

    if (exists && val == "off") || !exists {
        continue
    }

prashanthpai:
This needs a rebase. We [moved](https://github.com/gluster/glusterd2/commit/aaaaf9980a4076617b65b54aaa4a62b079653c1d) some quota specific code to the quota plugin.

prashanthpai:
> @prashanthpai quota disable can also use Actor framework? #708

Sure, but I don't know if quota daemon enabling/disabling can be done via volume set.

harigowtham:
@prashanthpai,  @aravindavk, I had a talk with @sanoj-unnikrishnan and we feel that having quota enable/disable as a command will be better than having it as a volume set option.
What are your thoughts on this?

aravindavk:
> @prashanthpai, @aravindavk, I had a talk with @sanoj-unnikrishnan and we feel that having quota enable/disable as a command will be better than having it as a volume set option.
What are your thoughts on this?

I think CLI can have separate commands for enable/disable if that provides better usability. But internally it can call Volume set API.

harigowtham:
@prashanthpai,  @aravindavk, I had a talk with @sanoj-unnikrishnan and we feel that having quota enable/disable as a command will be better than having it as a volume set option.
What are your thoughts on this?

harigowtham:
@aravindavk @prashanthpai Can you guys take a look at it? need this re-fractoring to work on further dependent code.

e2e/quota_enable.go
```diff
@@ -52,8 +52,14 @@ func testQuotaEnable(t *testing.T) {
...
 
-	err = client.QuotaEnable(volname)
-	r.Nil(err)
+	var optionReq api.VolOptionReq
+	optionReq.Advanced = true
+
+	quotaKey := "quota.enable"
+	optionReq.Options = map[string]string{quotaKey: "on"}
+	r.Nil(client.VolumeSet(volname, optionReq))
+	//err = client.QuotaEnable(volname)
```
prashanthpai: Remove commented code.

glusterd2/commands/volumes/volume-option.go
```diff
@@ -185,6 +185,10 @@ func volumeOptionsHandler(w http.ResponseWriter, r *http.Request) {
 			DoFunc: "vol-option.Validate",
 			Nodes:  []uuid.UUID{gdctx.MyUUID},
 		},
+		{
```
prashanthpai: Remove the redundant one that's down below. Also, add undo function (that already exists).

plugins/quota/actor.go
```diff
@@ -0,0 +1,68 @@
...
+		return nil
+	}
+	quotadDaemon, err := NewQuotad()
+	if err != nil {
+		return err
+	}
+
+	switch value {
+	case "on":
+
```
prashanthpai: Just a suggestion:

For this repeated pattern where error is the only value returned, there's a short notation in golang:

```go
if err := os.MkdirAll(path.Dir(quotadDaemon.pidfilepath), os.ModeDir|os.ModePerm); err != nil {
    return err
}
```

plugins/quota/actor.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+		// Create logFiledir dir
+		err = os.MkdirAll(path.Dir(quotadDaemon.logfilepath), os.ModeDir|os.ModePerm)
+		if err != nil {
+			return err
+		}
+		err = daemon.Stop(quotadDaemon, true)
+		if err != nil && err != gderrors.ErrProcessNotRunning && err != gderrors.ErrPidFileNotFound {
+			return err
+		}
```
prashanthpai: Over the long term, we should be sending an RPC to quota daemon to notify about volfile change without needing a restart.

plugins/quota/actor.go
```diff
@@ -0,0 +1,68 @@
...
+		err = daemon.Stop(quotadDaemon, true)
+		if err != nil && err == gderrors.ErrProcessNotRunning {
+			return err
+		}
+	}
+	return nil
+}
+
+func (actor *quotadActor) Undo(v *volume.Volinfo, key string, value string) error {
+	return nil
```
prashanthpai: TBD ?

plugins/quota/validate.go
```diff
@@ -5,34 +5,67 @@ import (
...
+		}
 	case "alert-time":
-		return nil
+		if isQuotaEnabled(v) == false {
+			err := fmt.Errorf("Quota not enabled to set this value: '%s'", key)
+			return err
+		}
 	case "default-soft-limit":
-		return nil
+		if isQuotaEnabled(v) == false {
```
prashanthpai: There's a keyword in go called `fallthrough`. Please look it up. It can help you get rid of this repetitive code.

pkg/errors/error.go
```diff
@@ -33,6 +33,7 @@ var (
 	ErrPeerLocalNode           = errors.New("The peer being added is the local node")
 	ErrProcessNotFound         = errors.New("The process is not running or is inaccessible")
 	ErrProcessAlreadyRunning   = errors.New("Process is already running")
+	ErrProcessNotRunning       = errors.New("Process is not running")
```
prashanthpai: Recently, @Madhu-1 sent a PR to make all error strings to start with small caps. You can do that here.

plugins/quota/actor.go
```diff
@@ -0,0 +1,105 @@
...
+
+type quotadActor struct{}
+
+const (
+	quotaDaemonKey = "enable"
+)
+
+// IsQuotadAffectedNode returns true if there are local bricks of volume on
+// which quota is enabled
+func IsQuotadAffectedNode() bool {
```
prashanthpai: Do not export this function is it's not used outside of this package.

plugins/quota/actor.go
```diff
@@ -0,0 +1,105 @@
...
+	volumes, e := volume.GetVolumes()
```
prashanthpai: Please consistently use `err` as variable for errors.

plugins/quota/actor.go
```diff
@@ -0,0 +1,105 @@
...
+	if e != nil {
+		log.WithError(e).Error("Failed to get volumes")
```
prashanthpai: Pass a logger to this function and use that to log.

plugins/quota/actor.go
```diff
@@ -0,0 +1,105 @@
...
+			os.ModeDir|os.ModePerm); err != nil {
+			return err
+		}
+
+		err = daemon.Stop(quotadDaemon, true, logger)
+		if err != nil && err != gderrors.ErrProcessNotRunning && err != gderrors.ErrPidFileNotFound {
+			return err
+		}
+		err = daemon.Start(quotadDaemon, true, logger)
+		if err != gderrors.ErrProcessAlreadyRunning || err != nil {
```
prashanthpai: Shouldn't it be `&&` here ?

Also, check for `nil` first

plugins/quota/validate.go
```diff
@@ -5,34 +5,55 @@ import (
...
+	quotaEnabledKey = "quota.enable"
 )
 
 func validateOptions(v *volume.Volinfo, key string, value string) error {
 
-	if isQuotaEnabled(v) == false {
-		err := fmt.Errorf("Quota not enabled to set this value: '%s'", key)
+	// Check if volume is started
+	if v.State != volume.VolStarted {
+		err := errors.ErrVolNotStarted
```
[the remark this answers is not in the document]
harigowtham: done

pkg/errors/error.go
harigowtham: done

plugins/quota/actor.go
harigowtham: done

plugins/quota/actor.go
harigowtham: done

plugins/quota/actor.go
harigowtham: done

plugins/quota/actor.go
harigowtham: this had a case missing. have fixed it.

plugins/quota/actor.go
```diff
@@ -0,0 +1,87 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+type quotadActor struct{}
+
+const (
+	quotaDaemonKey = "enable"
+)
+
+// IsQuotadAffectedNode returns true if there are local bricks of volume on
```
prashanthpai: nit: `isQuotadAffectedNode` in comment too.

plugins/quota/actor.go
```diff
@@ -0,0 +1,87 @@
...
+// which quota is enabled
+func isQuotadAffectedNode(logger log.FieldLogger) bool {
+	volumes, err := volume.GetVolumes()
```
prashanthpai: This error should be propagated back to caller. You can move `volume.GetVolumes()` call to the `Do()` and pass volumes to this function. In that case, you don't have to pass logger too.

plugins/quota/actor.go
```diff
@@ -0,0 +1,87 @@
...
+		return err
+	}
+
+	// Create logFiledir dir
+	if err := os.MkdirAll(path.Dir(quotadDaemon.logfilepath),
+		os.ModeDir|os.ModePerm); err != nil {
+		return err
+	}
+
+	// Should the bitd and scrubber untouched on this node
```
prashanthpai: > Should the bitd and scrubber untouched on this node

Can you explain this a little better ?

plugins/quota/actor.go
```diff
@@ -0,0 +1,87 @@
...
+	if !isQuotadAffectedNode(logger) {
+		// This condition is for disable
+		// TODO: Need to ignore errors where process is already running.
+		daemon.Stop(quotadDaemon, true, logger)
+	} else {
+		//TODO: Handle ENOENT of pidFile
+		err = daemon.Stop(quotadDaemon, true, logger)
```
prashanthpai: The first error gets ignored. You should handle that or at lease log it.

plugins/quota/validate.go
```diff
@@ -3,44 +3,66 @@ package quota
...
 	}
+
+	return nil
 }
 
 func isQuotaEnabled(v *volume.Volinfo) bool {
-	val, exists := v.Options[quotaEnabledKey]
-	if exists && val == "on" {
-		return true
+	pid, err := daemon.ReadPidFromFile("/usr/local/var/run/glusterd2/quotad.pid")
```
prashanthpai: I'm sorry, I should've been clearer - I expect two separate functions to exist:

* `isQuotaEnabled` checks if quota is enabled
* `isQuotadRunning` checks if the daemon is actually running

You may have to use both of these functions.

In `isQuotadRunning()`, do not read the hard-coded pidfile directly. Instead, use the `PidFile()` method defined on the quota type that implements the daemon interface,

harigowtham: done

plugins/quota/actor.go
harigowtham: done

plugins/quota/actor.go
harigowtham: thats a nit. fixed it.

plugins/quota/actor.go
harigowtham: done.

plugins/quota/actor.go
harigowtham: done

prashanthpai:
@harigowtham needs a rebase

plugins/quota/actor.go
```diff
@@ -0,0 +1,89 @@
...
+			if len(bricks) > 0 {
+				return true
+			}
+			continue
+		}
+	}
+	return false
+}
+
+func (actor *quotadActor) Do(v *volume.Volinfo, key string, value string, logger log.FieldLogger) error {
```
Madhu-1: suggestion: 
you can change 
```
func (actor *quotadActor) Do(v *volume.Volinfo, key string, value string, logger log.FieldLogger)
```
to 
```
func (actor *quotadActor) Do(v *volume.Volinfo, key , value string, logger log.FieldLogger)
```

plugins/quota/actor.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	// Check for making changes to quotad on this node
+	if !isQuotadAffectedNode(volumes) {
+		// This condition is for disable
+		// TODO: Need to ignore errors where process is already running.
+		daemon.Stop(quotadDaemon, true, logger)
+		logger.Info("quotad is already running")
+	} else {
+		//TODO: Handle ENOENT of pidFile
+		err = daemon.Stop(quotadDaemon, true, logger)
```
Madhu-1: don't we need to log error for daemon stop?

plugins/quota/actor.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// Create logFiledir dir
+	if err := os.MkdirAll(path.Dir(quotadDaemon.logfilepath),
+		os.ModeDir|os.ModePerm); err != nil {
+		return err
+	}
+
+	volumes, err := volume.GetVolumes()
+	if err != nil {
+		logger.WithError(err).Error("Failed to get volumes")
```
Madhu-1: question:
In case of failure here, do we need to delete ```pidfile dir```?

plugins/quota/actor.go
```diff
@@ -0,0 +1,89 @@
...
+		err = daemon.Start(quotadDaemon, true, logger)
+		if err != nil {
+			logger.WithError(err).Error("quotad start failed")
+			return err
+		}
+	}
+	return err
+}
+
+func (actor *quotadActor) Undo(v *volume.Volinfo, key string, value string, logger log.FieldLogger) error {
```
Madhu-1: suggestion:
you can change
```
func (actor *quotadActor) Undo(v *volume.Volinfo, key string, value string, logger log.FieldLogger)
```
to 
```
func (actor *quotadActor) Undo(v *volume.Volinfo, key , value string, logger log.FieldLogger)
```

plugins/quota/validate.go
```diff
@@ -3,38 +3,65 @@ package quota
...
 	"github.com/gluster/glusterd2/glusterd2/xlator"
+	"github.com/gluster/glusterd2/pkg/errors"
 )
 
 const (
-	quotaEnabledKey = "features.quota"
+	quotaEnabledKey = "quota.enable"
 )
 
 func validateOptions(v *volume.Volinfo, key string, value string) error {
```
Madhu-1: suggestion:
same here
we can change 
```
key string, value string
```
to 
```
key,value string
```

plugins/quota/validate.go
```diff
@@ -3,38 +3,65 @@ package quota
...
+		fallthrough
 	case "soft-timeout":
-		return nil
+		fallthrough
 	case "alert-time":
-		return nil
+		fallthrough
 	case "default-soft-limit":
-		return nil
+		if isQuotaEnabled(v) == false {
```
Madhu-1: for all the fallthough this condition is checked. is it expected?

plugins/quota/validate.go
```diff
@@ -3,38 +3,65 @@ package quota
...
+			return fmt.Errorf("Quota not enabled to set this value: '%s'", key)
```
Madhu-1: error message should start with small letter

plugins/quota/validate.go
```diff
@@ -3,38 +3,65 @@ package quota
...
+		return errors.ErrVolNotStarted
 	}
 
 	switch key {
-	case "deem-statfs":
+	case "enable":
+		switch value {
+		case "on":
+			ret, err := isQuotadRunning()
+			if err == nil && ret {
```
Madhu-1: I think we can check only error condition here, no need to check ```ret```

harigowtham: done

plugins/quota/validate.go
harigowtham: done

plugins/quota/validate.go
harigowtham: yes

plugins/quota/validate.go
harigowtham: done

plugins/quota/actor.go
harigowtham: done

plugins/quota/actor.go
harigowtham: Not necessary. the pidfile dir is not used for any comparison (only the pidfile itself is used to check if the process is there or not). And failed to get volumes can happen for some reason that is irrelevant to the quotad running. in case if it was already running, the pidfile dir will have the pidfile and deleteing the dir is not the thing to do by then.

plugins/quota/actor.go
harigowtham: As its the expected way of restarting, we need not error log it. As error log might create panic and that is not necessary here.

plugins/quota/actor.go
harigowtham: done

plugins/quota/actor.go
```diff
@@ -0,0 +1,89 @@
+package quota
+
+import (
+	"os"
+	"path"
+
+	"github.com/gluster/glusterd2/glusterd2/daemon"
+	"github.com/gluster/glusterd2/glusterd2/volume"
+	"github.com/gluster/glusterd2/glusterd2/xlator"
+	log "github.com/sirupsen/logrus"
```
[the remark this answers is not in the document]
harigowtham: Being new to Go didn't know about this standard.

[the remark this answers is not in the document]
Madhu-1: its gd2 coding convention mentioned here https://github.com/gluster/glusterd2/blob/master/doc/coding.md

Madhu-1:
@harigowtham CI failure looks genuine, can you please look at it.
```
07:08:35 Building glusterd2 v4.1.0-7.git241ef0b
07:08:53 # github.com/gluster/glusterd2/plugins/quota
07:08:53 plugins/quota/validate.go:27:16: ret declared and not used
07:08:53 make: *** [build/glusterd2] Error 1
```

plugins/quota/validate.go
```diff
@@ -3,38 +3,65 @@ package quota
...
+	if v.State != volume.VolStarted {
+		return errors.ErrVolNotStarted
 	}
 
 	switch key {
-	case "deem-statfs":
+	case "enable":
+		switch value {
+		case "on":
+			ret, err := isQuotadRunning()
```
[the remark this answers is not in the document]
harigowtham: done.

plugins/quota/actor.go
harigowtham: Debug is a bit too aggressive. We might need that info when we try to debug the customer setup. and if its debug, we wont get it. Being in info its not going to fill up the log as we are not going to run it a lot of times. and in the times when it is run its better to have it in the log to know whats happening.

plugins/quota/actor.go
```diff
@@ -0,0 +1,90 @@
...
+	// Check for making changes to quotad on this node
+	if !isQuotadAffectedNode(volumes) {
+		// This condition is for disable
+		// TODO: Need to ignore errors where process is already running.
+		daemon.Stop(quotadDaemon, true, logger)
+		logger.Info("quotad is already running")
+	} else {
+		//TODO: Handle ENOENT of pidFile
+		err = daemon.Stop(quotadDaemon, true, logger)
+		logger.Info("quotad pidfile missing")
```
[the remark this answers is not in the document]
harigowtham: have changed it. and the quota pidfile missing situation will result during the first start. its expected and as we need to differentiate between other failures and this I have added different log messages

plugins/quota/validate.go
```diff
@@ -3,38 +3,65 @@ package quota
...
-		err := fmt.Errorf("Quota not enabled to set this value: '%s'", key)
-		return err
+	// Check if volume is started
+	if v.State != volume.VolStarted {
+		return errors.ErrVolNotStarted
 	}
 
 	switch key {
-	case "deem-statfs":
+	case "enable":
```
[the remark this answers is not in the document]
harigowtham: The daemon start stop them self have the check for is daemon running(pidfile exists). so having it there wont be of much use as we will be doing a redundant work. and in case we have it there, we change the volfile and update it. if we have it this way we can avoid it. As it is also one type of validation, I think its better to have it in the validation itself.

harigowtham:
@Madhu-1 apologies tested the patch with other changes and made it at the end and missed testing that portion. have updated now.

plugins/quota/actor.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		logger.WithError(err).Error("failed to get volumes")
+		return err
+	}
+	// Check for making changes to quotad on this node
+	if !isQuotadAffectedNode(volumes) {
+		// This condition is for disable
+		// TODO: Need to ignore errors where process is already running.
+		daemon.Stop(quotadDaemon, true, logger)
+		logger.Info("quotad is already running")
```
prashanthpai: I do not follow this log message. The code stops quotad and immediately logs something else.

plugins/quota/actor.go
```diff
@@ -0,0 +1,95 @@
...
+	volumes, err := volume.GetVolumes()
+	if err != nil {
+		logger.WithError(err).Error("failed to get volumes")
+		return err
+	}
+	// Check for making changes to quotad on this node
+	if !isQuotadAffectedNode(volumes) {
+		// This condition is for disable
+		// TODO: Need to ignore errors where process is already running.
+		daemon.Stop(quotadDaemon, true, logger)
```
prashanthpai: The error message of `daemon.Stop` shouldn't be ignored. It should at least be logged as Debug.

plugins/quota/actor.go
```diff
@@ -0,0 +1,95 @@
...
+
+	volumes, err := volume.GetVolumes()
+	if err != nil {
+		logger.WithError(err).Error("failed to get volumes")
+		return err
+	}
+	// Check for making changes to quotad on this node
+	if !isQuotadAffectedNode(volumes) {
+		// This condition is for disable
+		// TODO: Need to ignore errors where process is already running.
```
prashanthpai: You can address the `TODO` in this PR itself. See self heal daemon code for example

plugins/quota/actor.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	volumes, err := volume.GetVolumes()
+	if err != nil {
+		logger.WithError(err).Error("failed to get volumes")
+		return err
+	}
+	// Check for making changes to quotad on this node
+	if !isQuotadAffectedNode(volumes) {
+		// This condition is for disable
```
prashanthpai: I'm not sure how you figure out that this is disable without checking the user input - key and value variables

plugins/quota/actor.go
```diff
@@ -0,0 +1,95 @@
...
+		// TODO: Need to ignore errors where process is already running.
+		daemon.Stop(quotadDaemon, true, logger)
+		logger.Info("quotad is already running")
+	} else {
+		//TODO: Handle ENOENT of pidFile
+		err = daemon.Stop(quotadDaemon, true, logger)
+		if err == errors.ErrPidFileNotFound {
+			logger.Info("quotad stop failed as pidfile missing")
+		} else {
+			logger.Warn("quotad stop failed")
```
prashanthpai: If you have found that quotad was running during validate step, if pidfile is missing by the time we get here, it's an error.

plugins/quota/validate.go
```diff
@@ -43,6 +70,26 @@ func isQuotaEnabled(v *volume.Volinfo) bool {
 	return false
 }
 
+// isQuotadRunning returns true for running and false for
+// failures and not running on that machine.
+// It is not volume based.
+func isQuotadRunning() (bool, error) {
```
prashanthpai: This can be a function that only returns true or false.

harigowtham: the err is necessary to handle the following failure.

 77         quotadDaemon, err := NewQuotad()
 78         if err != nil {
 79                 return false, err
 80         }

plugins/quota/actor.go
harigowtham: If the pidfile is missing then the daemon is not running, as its a stop whatever was supposed to be done was done. so i think we are fine here.

plugins/quota/actor.go
harigowtham: For both disable and enable we need to do the same (restart). 

Lets take a case where we have only one volume on a node. it goes to the else part were we stop and the start. stop fails with pidfile not found and start , starts the daemon. 
if we issue stop on this volume, then it goes to the if part where we stop it. (Nit: need not correct the log message)

If we have two volumes in the same node. then they share the quotad (difference lies only in the volfile) during start for the second volume, we go to the else part and we stop the quotad and then we go to the daemon start and start the quotad.(now it will start with the volfile changes. ie its common for both volumes) 
During the stop for the second volume, we again go to the else part were we stop and start with the volfile changes for a single volume.

plugins/quota/actor.go
harigowtham: The todo is for the else part and it is handled. the todo here is common error handling and is done now.

plugins/quota/actor.go
harigowtham: done

plugins/quota/actor.go
harigowtham: it was messed up. fixed it,

plugins/quota/actor.go
prashanthpai: I'll get some more confidence is there were a e2e testcase for it. Can you please add one ? A simple test which enables quota, checks if quotad process is running, disables quota, checks if quoatd process is running.

plugins/quota/actor.go
prashanthpai: > If the pidfile is missing then the daemon is not running

That assumption may not be true in corner cases (gd2 going down etc). But that's okay, for now.

The validate step would've confirmed that it's running before getting here during a "off" operation. Who would stop quotad after validate step and before we reach this point ? That's the case you described - pidfile not existing because quotad is stopped.

harigowtham: If gd2 goes down after validate and before daemon stop, the quotad is still up. and its pidfile holds fine. once it comes up we will still have quotad running. To fix such corner cases we will have to come up with force disable later. correct me if i understood what you are saying.

plugins/quota/validate.go
prashanthpai: That error is possible only if quota (glusterfs) isn't installed on the system. You can ignore that. If quota isn't installed, the validate in rest handler will return "option not found" error.

plugins/quota/actor.go
prashanthpai: > correct me if i understood what you are saying

> That assumption may not be true in corner cases (gd2 going down etc). But that's okay, for now.

Ignore ^ for now.

> The validate step would've confirmed that it's running before getting here during a "off" operation. Who would stop quotad after validate step and before we reach this point ? That's the case you described - pidfile not existing because quotad is stopped.

I want to know how will the pidfile can be missing when `Do()` is executed  ?

harigowtham: During the first quotad start for a node. now no pidfile would be there as quota was never run. and the code flows through the stop and then the start. 
In the stop it will fail as it couldn't find the pidifle and and then it would be actually started in start.

plugins/quota/validate.go
harigowtham: if so then i'll remove it and send a single return value and handle is as per that.

harigowtham: and if thats the case can i remove it the err check for NewQuotad here?

plugins/quota/actor.go
harigowtham: I had a error string missing for quotad disable on disabled volume. have added that part. I have updated the e2e testcase for quota enable with what you have asked for.

plugins/quota/actor.go
```diff
@@ -0,0 +1,100 @@
...
+			logger.Error("quotad stop failed")
+		}
+	} else {
+		// This is to handle the all the start and also the stop
+		// is there is another volume with quota enabled on the same
+		// node
+		err = daemon.Stop(quotadDaemon, true, logger)
+		if err == errors.ErrPidFileNotFound {
+			logger.Info("quotad stop failed as pidfile missing")
+		} else {
```
Madhu-1: I think you missed handling ```err==nil``` part

plugins/quota/validate.go
```diff
@@ -43,6 +76,26 @@ func isQuotaEnabled(v *volume.Volinfo) bool {
...
+	quotadDaemon, err := NewQuotad()
+	if err != nil {
+		return false
+	}
+	pid, err := daemon.ReadPidFromFile(quotadDaemon.PidFile())
+	if err != nil {
+		return false
+	}
+
+	_, err = daemon.GetProcess(pid)
```
Madhu-1: suggestion: this can be simplified
```
if _, err = daemon.GetProcess(pid);err != nil{
return false
}
```

plugins/quota/actor.go
Madhu-1: I think you missed handling err==nil part

pkg/errors/error.go
```diff
@@ -36,6 +36,8 @@ var (
 	ErrBitrotAlreadyEnabled    = errors.New("bitrot is already enabled")
 	ErrBitrotAlreadyDisabled   = errors.New("bitrot is already disabled")
 	ErrBitrotNotEnabled        = errors.New("bitrot is not enabled")
+	ErrProcessNotRunning       = errors.New("process is not running")
```
Madhu-1: Suggestion/question

instead of returning process is already running error.
can we be more specific while returning an error like which process is already running

plugins/quota/actor.go
```diff
@@ -0,0 +1,100 @@
...
+		if !isQuotaEnabled(v) {
+			continue
+		} else if v.State != volume.VolStarted {
+			continue
+		} else {
+			bricks := v.GetLocalBricks()
+			if len(bricks) > 0 {
+				return true
+			}
+			continue
```
Madhu-1: I don't think continue is required here, can you please check

plugins/quota/validate.go
```diff
@@ -3,38 +3,71 @@ package quota
...
+			}
+		case "off":
+			running := isQuotadRunning()
+			if !running {
+				return errors.ErrProcessNotRunning
+			}
+			// As quotad for various process share a
+			// single pid, it is necessary to check the
+			// options too.
+			if !isQuotaEnabled(v) {
```
Madhu-1: question:
I think checking options should have the highest priority
first, we should check ```isQuotaEnabled``` and then we should check ``` isQuotadRunning```

harigowtham: I think this is fine, as quotad is shared between a number of volumes in the same node. and if the quotad is brought down because of some other volume or action and if we want to start it, this is better. we check if its not running then we start it. we done worry about the option which is "enabled". If you have a better case, do let me know. we will discuss and decide which one to have at the top.

plugins/quota/actor.go
harigowtham: yep. its not required.

pkg/errors/error.go
harigowtham: thinking of reuse, i had a common error. as we have a lot of errors that we can create as per need and also during restart situation, this change will be helpful to know which process isnt coming up.

plugins/quota/validate.go
harigowtham: done

plugins/quota/actor.go
harigowtham: done

plugins/quota/validate.go
```diff
@@ -3,38 +3,71 @@ package quota
...
+				// As quotad for various process share a
+				// single pid, it is necessary to check the
+				// options too.
+				if isQuotaEnabled(v) {
+					return errors.ErrProcessAlreadyRunning
+				}
+
+			}
+		case "off":
+			running := isQuotadRunning()
```
Madhu-1: This can also be simplified as @prashanthpai suggested.
```
if isQuotadRunning()  {
    return errors.ErrProcessAlreadyRunning
}
```

harigowtham: done

plugins/quota/actor.go
```diff
@@ -0,0 +1,101 @@
...
+		// If its the last node that is going to stop the daemon,
+		// it enters the if part.
+		err = daemon.Stop(quotadDaemon, true, logger)
+		if err != nil {
+			logger.Error("quotad stop failed")
+		}
+	} else {
+		// This is to handle the all the start and also the stop
+		// is there is another volume with quota enabled on the same
+		// node
```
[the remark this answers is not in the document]
harigowtham: it had a lot of typos. have fixed it. if needs more change let me know

plugins/quota/actor.go
```diff
@@ -0,0 +1,101 @@
...
+			logger.Info("quotad stop failed as pidfile missing")
+		} else if err != nil {
+			logger.Warn("quotad stop failed")
+		} else {
+			logger.Info("quotad stopped for restart")
+		}
+		err = daemon.Start(quotadDaemon, true, logger)
+		if err != nil {
+			logger.WithError(err).Error("quotad start failed")
+			return err
```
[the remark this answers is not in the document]
harigowtham: done

plugins/quota/validate.go
```diff
@@ -3,38 +3,71 @@ package quota
...
+	if v.State != volume.VolStarted {
+		return errors.ErrVolNotStarted
 	}
 
 	switch key {
-	case "deem-statfs":
+	case "enable":
+		switch value {
+		case "on":
+			running := isQuotadRunning()
```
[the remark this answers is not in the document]
harigowtham: done

plugins/quota/validate.go
```diff
@@ -3,38 +3,71 @@ package quota
...
+	quotaEnabledKey = "quota.enable"
 )
 
-func validateOptions(v *volume.Volinfo, key string, value string) error {
+func validateOptions(v *volume.Volinfo, key, value string) error {
 
-	if isQuotaEnabled(v) == false {
-		err := fmt.Errorf("Quota not enabled to set this value: '%s'", key)
-		return err
+	// Check if volume is started
```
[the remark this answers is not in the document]
harigowtham: done

plugins/quota/validate.go
```diff
@@ -3,38 +3,71 @@ package quota
...
 	case "soft-timeout":
-		return nil
+		fallthrough
 	case "alert-time":
-		return nil
+		fallthrough
 	case "default-soft-limit":
-		return nil
+		if !isQuotaEnabled(v) {
+			return fmt.Errorf("quota not enabled to set this value: '%s'", key)
```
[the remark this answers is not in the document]
harigowtham: done

e2e/quota_enable.go
```diff
@@ -52,8 +52,36 @@ func testQuotaEnable(t *testing.T) {
...
+	// Enable quota
+	quotaKey = "quota.enable"
+	optionReqOn.Options = map[string]string{quotaKey: "on"}
+
+	// Quotad should be there
+	r.Nil(client.VolumeSet(volname, optionReqOn))
+
+	// check the error for enabling it again
+	err = client.VolumeSet(volname, optionReqOn)
+	r.Contains(err.Error(), "process is already running")
```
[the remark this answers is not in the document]
harigowtham: will update this in the next patch set

e2e/quota_enable.go
```diff
@@ -52,8 +52,36 @@ func testQuotaEnable(t *testing.T) {
...
 
-	err = client.QuotaEnable(volname)
-	r.Nil(err)
+	quotaKey := "quota.enable"
+	var optionReqOff api.VolOptionReq
+	optionReqOff.Advanced = true
+
+	optionReqOff.Options = map[string]string{quotaKey: "off"}
+
+	// Quota not enabled no not daemon
```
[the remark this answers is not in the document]
harigowtham: typo. sorry

plugins/quota/actor.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	volumes, err := volume.GetVolumes()
+	if err != nil {
+		logger.WithError(err).Error("failed to get volumes")
+		return err
+	}
+	// Check for making changes to quotad on this node
+	// For both disable and enable we need to do the same (restart).
+	if !isQuotadAffectedNode(volumes) {
```
prashanthpai: I thought a little more about this code organisation. As mentioned earlier, you must rely on user inputs `key` and `value`. Your code is focusing on the end action to be taken i.e restarting quotad but ignores how you arrive at that decision and what data you'd use to arrive at that decision.

For better readability, this condition can be:

```go
if value == "off" && isQuotadStopRequired(volumes) {
    if err = daemon.Stop(quotadDaemon, true, logger); err != nil {
        ....
    }
}
```

Your code is relying on the fact that the `Volinfo` passed is the new/modified one which contains the user input. That makes understanding this code very hard. You shouldn't rely on contents of `Volinfo` but rather on the user input (key and value variables) even though `Volinfo` contains the same information.

You have re-ordered the txn steps such that updating `Volinfo` comes first. The purpose of that ordering is only to allow volgen to regenerate the volfiles and not to use rely on `Volinfo` here for decision making. In future, we may change the volgen to generate volfiles from custom/temporary `Volinfo`. The code written here should be resilient to that.

harigowtham: done

plugins/quota/actor.go
```diff
@@ -0,0 +1,101 @@
...
+		// This condition is for disable
+		// If its the last node that is going to stop the daemon,
```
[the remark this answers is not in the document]
harigowtham: done

plugins/quota/actor.go
```diff
@@ -0,0 +1,101 @@
...
+		// it enters this 'if' part.
+		err = daemon.Stop(quotadDaemon, true, logger)
+		if err != nil {
+			logger.Error("quotad stop failed")
+		}
+	} else {
+		// This "else" is to handle all quotad start. It is also for
```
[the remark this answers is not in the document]
harigowtham: done

e2e/quota_enable.go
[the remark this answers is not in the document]
harigowtham: done

e2e/utils_test.go
```diff
@@ -346,3 +348,26 @@ func loopDevicesCleanup(t *testing.T) error {
 func formatVolName(volName string) string {
 	return strings.Replace(volName, "/", "-", 1)
 }
+
+func IsDaemonRunning(pidpath string) bool {
```
prashanthpai: As all tests that use this function are present in the same `e2e` package, please do not export this function. 
Suggestion: You can rename this function as `isProcessRunning` to be slightly broader. In future, tests may use this to check if any process is running, not necessarily those processes that turn into daemons.  

Clean implementation though :+1:

plugins/quota/actor.go
```diff
@@ -0,0 +1,96 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+type quotadActor struct{}
+
+const (
+	quotaDaemonKey = "enable"
+)
+
+// isQuotadAffectedNode returns true if there are local bricks of volume on
```
prashanthpai: The comment needs to be updated. Here's a suggested text:

```
isQuotadStopRequired checks if the quota daemon has to be stopped.
The quotad process needs to be stopped on a peer only when that peer
does not have bricks belonging to a quota enabled volume.
```

plugins/quota/actor.go
```diff
@@ -0,0 +1,96 @@
...
+		os.ModeDir|os.ModePerm); err != nil {
+		return err
+	}
+
+	volumes, err := volume.GetVolumes()
+	if err != nil {
+		logger.WithError(err).Error("failed to get volumes")
+		return err
+	}
+	// Check for making changes to quotad on this node
```
prashanthpai: You can remove these two lines of comment as the comment in the else part would suffice

plugins/quota/actor.go
```diff
@@ -0,0 +1,96 @@
...
+	// For both disable and enable we need to do the same (restart).
+	if value == "off" && !isQuotadStopRequired(volumes) {
```
aravindavk: flip the condition? `if value == "off" && isQuotadStopRequired(volumes) `

plugins/quota/actor.go
```diff
@@ -0,0 +1,96 @@
...
+
+type quotadActor struct{}
+
+const (
+	quotaDaemonKey = "enable"
+)
+
+// isQuotadAffectedNode returns true if there are local bricks of volume on
+// which quota is enabled
+func isQuotadStopRequired(volumes []*volume.Volinfo) bool {
```
aravindavk: reverse the return value, 

- return `true` if stop required(No volumes in started state with Quota enabled)
- return `false` if stop not required(Some volumes still running with Quota enabled)

harigowtham: done

plugins/quota/actor.go
harigowtham: done

plugins/quota/actor.go
harigowtham: done

plugins/quota/actor.go
harigowtham: make better sense

e2e/utils_test.go
harigowtham: done

e2e/quota_enable.go
```diff
@@ -44,16 +55,77 @@ func testQuotaEnable(t *testing.T) {
...
 		},
 		Force: true,
 	}
-	_, err = client.VolumeCreate(reqVol)
+
+	_, err := client.VolumeCreate(createReq)
 	r.Nil(err)
 
-	r.Nil(client.VolumeStart(volname, false), "volume start failed")
+	testQuotaEnable(t)
```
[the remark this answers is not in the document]
harigowtham: done

e2e/quota_enable.go
```diff
@@ -23,18 +24,28 @@ func testQuotaEnable(t *testing.T) {
 	brickDir, err := ioutil.TempDir(baseWorkdir, t.Name())
 	r.Nil(err)
 	defer os.RemoveAll(brickDir)
+	t.Logf("Using temp dir: %s", brickDir)
```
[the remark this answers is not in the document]
harigowtham: done

prashanthpai:
retest this please

prashanthpai:
Requesting other reviewers to take a look at this.

Issue #621, opened 2018-03-26
Port allocation happening twice while starting a daemon


samikshan:
During the daemon start process, port allocation is being done two times per daemon. This creates two entries in the portmap for each daemon and leads to the portmap entries not being cleared correctly when daemons are stopped.

PR #622, opened 2018-03-26
Do not allocate port two times while starting a daemon.

samikshan:
During daemon start the port allocation was being done twice. This
led to the portmap storing two entries corresponding to two different
ports for a single daemon.

Fixes: #621

Signed-off-by: Samikshan Bairagya <samikshan@gmail.com>


glusterd2/daemon/daemon.go
```diff
@@ -52,11 +52,12 @@ type Daemon interface {
 // When wait == true, this function can be used to spawn short term processes
 // which will be waited on for completion before this function returns.
 func Start(d Daemon, wait bool) error {
+	args := strings.Fields(d.Args())
 
 	log.WithFields(log.Fields{
 		"name": d.Name(),
 		"path": d.Path(),
-		"args": d.Args(),
+		"args": args,
```
atinmu: Sorry, but could you explain how only args instead of d.Args() help in fixing this issue?

samikshan: The method Args() in brick/glusterfsd.go assigns a port and then returns it along with other arguments required to start the daemon as a string. Now calling d.Args() two times means port allocation happens twice while trying to start the daemon.

aravindavk: I think you can fix `Args()` itself. Add below lines in the beginning of `Args()` function,

    if b.args != "" {
        return b.args
    }

prashanthpai:
retest this please

aravindavk:
I don't see update branch button for this patch.

aravindavk:
@samikshan looks like this is protected branch, I am not able to rebase from UI.(https://blog.github.com/2015-09-03-protected-branches-and-required-status-checks/)

Please rebase this branch.

samikshan:
retest this please

prashanthpai:
@samikshan Can you put a tick mark in the checkox in the right pane that goes something like `Allow edits from maintainers`. Hopefully, that'll allow us to rebase your patch when you're not around.

prashanthpai:
retest this please

prashanthpai:
@samikshan This might be a genuine failure.

glusterd2/daemon/daemon.go
samikshan: Making changes in glusterfsd.Args() is causing volume start to fail in the functional tests. Having these changes in daemon.Start() however doesn't fail these tests. I'm at a lack for any logical explanation behind this.

prashanthpai:
I had a quick look at this. As we run multiple glusterd2 on a single node in our tests, there seems to be port clash:

```
time="2018-04-20 11:32:49.799719" level=debug msg="Starting daemon." args=" --volfile-server 127.0.0.1 --volfile-server-port 23007 --volfile-id testvol.60203c9c-b8c0-49e3-a50d-e958782c64c4.tmp-TestVolume39615556
0-brick092930586 -p /tmp/gd2_func_test/w2/run/gluster/60203c9c-b8c0-49e3-a50d-e958782c64c4-tmp-TestVolume396155560-brick092930586.pid -S /tmp/gd2_func_test/w2/run/gluster/e6a9a95ee748c89.socket --brick-name /tmp/TestVolume396155560/brick092930586 --brick-port 49152 -l /tmp/gd2_func_test/w2/log/glusterfs/bricks/tmp-TestVolume396155560-brick092930586.log --xlator-option *-posix.glusterd-uuid=60203c9c-b8c0-49e3-a50d-e958782c64c4 --xlator-option testvol-server.transport.socket.listen-port=49152" name=glusterfsd path=/usr/local/sbin/glusterfsd source="[daemon.go:60:daemon.Start]"
time="2018-04-20 11:32:49.829047" level=info msg="client connected" address="127.0.0.1:49150" server=sunrpc source="[server.go:167:sunrpc.(*SunRPC).acceptLoop]" transport=tcp
time="2018-04-20 11:32:49.838482" level=debug msg="Child exited" pid=4047 source="[daemon.go:88:daemon.Start]" status="exit status 98"
```

The exit status 98 is the errno value of `EADDRINUSE`

This is code that retries to start the bricks again:

```go
func (b Brickinfo) StartBrick() error {

        brickDaemon, err := NewGlusterfsd(b)
        if err != nil {
                return err
        }

        for i := 0; i < BrickStartMaxRetries; i++ {
                err = daemon.Start(brickDaemon, true)
                if err != nil {
                        if errorContainsErrno(err, syscall.EADDRINUSE) || errorContainsErrno(err, anotherEADDRINUSE) {
                                // Retry iff brick failed to start because of port being in use.
                                // Allow the previous instance to cleanup and exit
                                time.Sleep(1 * time.Second)
                        } else {
                                return err
                        }
                } else {
                        break
                }
        }
        return nil
}
```

And as you can see, this will try to restart the brick with the same port.

prashanthpai:
We cannot co-ordinate brick port allocation across multiple glusterd2 instances. A long term good solution would be to let the bricks choose a free port themselves and then report it to glusterd2 during sign in.

We can live with stale ports in pmap as they are tagged with `LEASED` state and can be cleaned up if we don't get a corresponding sign in request in some fixed time period.

samikshan:
Shouldn't isPortFree() be able to help us avoid such port clashes while assigning port in registryAlloc()?

prashanthpai:
> Shouldn't isPortFree() be able to help us avoid such port clashes while assigning port in registryAlloc()?

It does, but only within a single glusterd2 instance. When there are multiple glusterd2 running on same node (CI test), there are races.

prashanthpai:
PR https://github.com/gluster/glusterd2/pull/900 supersedes this.

PR #624, opened 2018-03-27
volume-set: Add support for advanced/experimental options

kshlm:


glusterd2/commands/volumes/common.go
```diff
@@ -28,16 +28,27 @@ const volumeIDXattrKey = "trusted.glusterfs.volume-id"
...
 		if err != nil {
 			return err
 		}
 
-		if !o.IsSettable() {
+		switch {
+		case !o.IsSettable():
 			return fmt.Errorf("Option %s cannot be set", k)
+
+		case o.IsAdvanced() && !adv:
```
prashanthpai: I thought the usecase was only to classify options into types to filter them out in listing and responses so that advanced options don't show up.

@nbalacha Isn't it ?

kshlm: It is needed for both. We need to filter out options in volume get and need to add an additional hurdle when setting.

kshlm:
The e2e test TestVolumeOptions fails becuase as of this time, all options are advanced as none of them have been updated and the zero value of level is advanced.
@nbalacha is this how you intend to this to remain?

glusterd2/commands/volumes/common.go
```diff
@@ -28,16 +28,27 @@ const volumeIDXattrKey = "trusted.glusterfs.volume-id"
...
+			return fmt.Errorf("Option %s is an advanced option. To set it pass the advanced flag", k)
+
+		case o.IsExperimental() && !exp:
+			return fmt.Errorf("Option %s is an experimental option. To set it pass the experimental flag", k)
+
+		case o.IsDeprecated() && !!dep:
```
prashanthpai: `!dep`

Issue #510, opened 2017-12-28
Fix TOCTOU races in rest handlers


prashanthpai:
In REST handlers, there are TOCTOU (Time of check to time of use) races in code paths where we fetch something from store, do validations and then act on it. Something might have changed in between. Currently, locks are only held during the lifetime of the transaction. Locks need to be held outside of transactions too or the validations itself be a step (under locks) in the transaction.

Example:
```
1: volinfo = get_from_store()
2: do validation of request w.r.t volinfo
3. change volinfo based on request and store
```

The state of volinfo may have changed between step1 and step3 which makes the validation invalid.

Reported by @rafikc30

kshlm:
2 ways to fix this.
Either,
1. Do request validation as a txn step. Or,
2. Lock outside txn and then validate request.

2 is easier to do, so I'd prefer we go that way.

prashanthpai:
- [ ] Volume Create
- [ ] Volume Start
- [ ] Volume Stop
- [ ] Volume Delete
- [ ] Volume Set

prashanthpai:
- [ ] Bitrot
- [ ] Georep
- [ ] Quota

PR #625, opened 2018-03-27
Refactor volume create and other minor fixes

prashanthpai:
* Move disperse related code and txn code to its own files.
* Move volume create validation and volinfo creation into a step func
  which will be executed under lock. Updated issue #510 
* Slightly simplify validation of incoming request.
* Log failures inside txn.Get and txn.Set
* Include more info in errors returned at some places

### DIFF
--- a/glusterd2/commands/volumes/common.go
+++ b/glusterd2/commands/volumes/common.go
@@ -68,7 +68,7 @@ func validateXlatorOptions(opts map[string]string, volinfo *volume.Volinfo) erro
 	return nil
 }
 
-func expandOptions(opts map[string]string) (map[string]string, error) {
+func expandGroupOptions(opts map[string]string) (map[string]string, error) {
 	resp, err := store.Store.Get(context.TODO(), "groupoptions")
 	if err != nil {
 		return nil, err
--- a/glusterd2/commands/volumes/common.go
+++ b/glusterd2/commands/volumes/common.go
@@ -116,6 +116,7 @@ func notifyVolfileChange(c transaction.TxnCtx) error {
 	return nil
 }
 
+// This txn step is used in volume create and in volume expand
 func validateBricks(c transaction.TxnCtx) error {
 
 	var err error
@@ -145,6 +146,7 @@ func validateBricks(c transaction.TxnCtx) error {
 	return nil
 }
 
+// This txn step is used in volume create and in volume expand
 func initBricks(c transaction.TxnCtx) error {
 
 	var err error
@@ -190,6 +192,7 @@ func initBricks(c transaction.TxnCtx) error {
 	return nil
 }
 
+// This txn step is used in volume create and in volume expand
 func undoInitBricks(c transaction.TxnCtx) error {
 
 	var bricks []brick.Brickinfo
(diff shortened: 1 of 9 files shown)


glusterd2/commands/volumes/common.go
```diff
@@ -41,7 +41,7 @@ func validateOptions(opts map[string]string) error {
 		}
 
 		if err := o.Validate(v); err != nil {
-			return err
+			return fmt.Errorf("Failed to validate value(%s) for key(%s): %s", k, v, err.Error())
```
rishubhjain: @prashanthpai Shouldn't this be just "err" instead of "err.Error()", as Errorf also accepts interface instead of string

glusterd2/transaction/context.go
```diff
@@ -84,24 +84,18 @@ func (c *Tctx) WithPrefix(prefix string) *Tctx {
...
 	storeKey := c.prefix + "/" + key
-	_, e = store.Store.Put(context.TODO(), storeKey, string(json))
-	if e != nil {
-		c.log.WithFields(log.Fields{
-			"error": e,
-			"key":   storeKey,
-		}).Error("failed to set value")
+	_, err = store.Store.Put(context.TODO(), storeKey, string(json))
+	if err != nil {
+		c.log.WithError(err).WithField("key", key).Error("failed to set key in transaction context")
```
rishubhjain: I think here the key is being set in the store and not in transaction context

glusterd2/transaction/context.go
```diff
@@ -119,7 +113,7 @@ func (c *Tctx) Get(key string, value interface{}) error {
 	storeKey := c.prefix + "/" + key
 	r, err := store.Store.Get(context.TODO(), storeKey)
 	if err != nil {
-		c.log.WithError(err).WithField("key", storeKey).Error("failed to get value from store")
+		c.log.WithError(err).WithField("key", storeKey).Error("failed to get value from transaction context")
```
rishubhjain: I think here the key is being fetched from the store and not in transaction context

glusterd2/transaction/context.go
prashanthpai: That's right. I want to cleanup this repetitive pattern:

```go
if err := txn.Set(); err != nil {
    log.(....)
    return err
}
```

to 

```go
if err := txn.Set(); err != nil {
    return err
}
```

It reduces code clutter and greatly reduces lines of codes.

glusterd2/commands/volumes/common.go
prashanthpai: Either of it works **in this case**. But as `%s` is specified, and that's a string, I thought it would be explicit to mention `err.Error()` which is also a string.

Further, it won't work for custom error type.

The error type is an interface type. An error variable represents any value that can describe itself as a string. Here is the interface's declaration:

```go
type error interface {
    Error() string
}
```
I wrote a playground example for you to check this out: https://play.golang.org/p/v0uSi8hn1VD

prashanthpai: @rishubhjain This is a good read: https://blog.golang.org/error-handling-and-go

glusterd2/transaction/context.go
rishubhjain: so where are you gonna log the errors, in the calling function?

prashanthpai: Inside `txn.Set()`. Like done here in the PR

PR #634, opened 2018-04-02
Volume start/stop - Do validations under lock

prashanthpai:
Updates #510


glusterd2/commands/volumes/volume-start.go
```diff
@@ -76,64 +59,78 @@ func stopAllBricks(c transaction.TxnCtx) error {
...
 	logger := gdctx.GetReqLogger(ctx)
-
 	volname := mux.Vars(r)["volname"]
-	vol, e := volume.GetVolume(volname)
-	if e != nil {
+
+	lock, unlock := transaction.CreateLockFuncs(volname)
+	// Taking a lock outside the txn as volinfo.Nodes() must also
+	// be populated holding the lock. See issue #510
+	if err := lock(ctx); err != nil {
```
kshlm: If we made getting the volinfo and setting it into the context a transaction step, the transaction framework would take care of handling the locking/unlocking. We don't take a lock on the actual volume, but on its name. We can take a lock on a name even if the thing doesn't exist.

I'm now wondering why we didn't do it before.

I'm now wondering why we didn't do this earlier.

kshlm: Well,I know why now. We need to know the nodes where steps need to run. That's why.

prashanthpai: > If we made getting the volinfo and setting it into the context a transaction step, the transaction framework would take care of handling the locking/unlocking

I thought of putting the validations in a step function like as we did for volume create. But in this case we populate the list of nodes for the transaction step from volinfo. We do not want that list of nodes that make up the volume to change.

```go
                {
                        DoFunc:   "vol-start.StartBricks",
                        UndoFunc: "vol-start.StartBricksUndo",
                        Nodes:    volinfo.Nodes(),
                },
        }
```

PR #636, opened 2018-04-04
Do not auto-generate template files

prashanthpai:
...now that we have moved to using volgen2


prashanthpai:
retest this please

Issue #576, opened 2018-02-20
Use the name 'peer' through-out


prashanthpai:
I noticed that our code has different names (server, peer, node) to refer to peers. This can get confusing to others. For example, `BrickInfo` has a field `NodeID` which is really peer ID.

atinmu:
+1, let's get that fixed.

rishubhjain:
@prashanthpai  This issue can be closed now

atinmu:
@prashanthpai @rishubhjain Anything pending on this? If not can this be closed?

rishubhjain:
@atinmu Few were remaining, https://github.com/gluster/glusterd2/pull/910 will address this issue.

rishubhjain:
@prashanthpai I guess the issue has been addressed, we can close this issue now.

prashanthpai:
@rishubhjain There's a lot more places:

```sh
$ grep -nR "node" . | grep -v vendor | grep -v .git | grep -v tags | less
```

PR #638, opened 2018-04-05
Using Peer instead of Node

rishubhjain:
Changes for BrickInfo and BrickReq

https://github.com/gluster/glusterd2/issues/576


rishubhjain:
@prashanthpai  @aravindavk Please review

PR #643, opened 2018-04-06
Allow only one cluster join/leave request at a time

### LINKED ISSUE #210: Handle simultaneous peer add/delete requests
When a glusterd instance concurrently receives add peer and/or delete peer, the current behaviour is undefined. And during add peer or delete peer, the transaction framework also isn't usable. We need some form of locking during peer operations which modify the node's cluster membership. The current etcd management code only takes locks for etcd start/restart.

prashanthpai:
If a glusterd2 instance is already processing a cluster membership
request (leave or join), return 409 (Conflict) to the other such
client requests immediately.

Closes #210


prashanthpai:
> Just so I"m clear, with this you prevent the gd2 that is being detached/attached from processing multiple such requests at a time. But this doesn't prevent the TSP from processing multiple attach/detach requests.
This PR solves part of the issue #210, but not completely.

I opened #210 to track and handle it for a single node. There was no elasticetcd back then. I'll open a separate issue under elastic etcd for that.

PR #645, opened 2018-04-09
Small bugs and unreachable code in code-base

rishubhjain:


glusterd2/servers/rest/utils/utils.go
```diff
@@ -56,7 +56,7 @@ func SendHTTPError(ctx context.Context, w http.ResponseWriter, statusCode int,
 
 	var resp api.ErrorResp
 	errMsg := fmt.Sprint(err)
-	if errMsg != "" || errMsg != "<nil>" || len(errCodes) == 0 {
+	if errMsg != "" && errMsg != "<nil>" || len(errCodes) == 0 {
```
prashanthpai: You're changing the logic here.

Issue #559, opened 2018-02-07
Use 'required' field in Gopkg.toml to include linters


prashanthpai:
More about the `required` field [here](https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md#required)

PR #646, opened 2018-04-09
Adding gometalinter as required field in Gopkg.toml file

rishubhjain:
@prashanthpai Please review 
Issue: https://github.com/gluster/glusterd2/issues/559
Added gometalinter since we use linters provided by gometalinter package, would lik eto know if this is right way to do it.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,6 +2,18 @@
 
 
 [[projects]]
+  name = "github.com/alecthomas/gometalinter"
+  packages = ["."]
+  revision = "46cc1ea3778b247666c2949669a3333c532fa9c6"
+  version = "v2.0.5"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/alecthomas/units"
+  packages = ["."]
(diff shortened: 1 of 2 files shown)


rishubhjain:
@prashanthpai This adds dependencies in vendor/ but does not install them,  to install it we will have to mention the installation step in Makefile, or do it manually

kshlm:
We have a script at `scripts/install-reqs.sh` that installs gometalinter and dep. This is used to install both in centos-ci.

You can add a new `install-reqs` target to the makefile that calls this script. And set this new target as a dependency of `verify`, instead of the current `check-reqs` dependency. This will ensure gometalinter is available when `make verify` is run.

You can do it better by splitting the req script into individual scripts to install dep and gometalinter, and having individual check and install targets for each requirement in the makefile. You can even do away with the check targets, if you ensure that the install scripts check and install.

rishubhjain:
@kshlm ok, so there is no point of adding gometalinter in the gopkg.toml file. I felt this approach was better than having scripts.

PR #649, opened 2018-04-12
group-vol-set: Add description support for profiles

poornimag:
Also change the way profile disable is implemented, offvalue
is not required and having it can create confusion and is not
readable in vol info output.


glusterd2/commands/volumes/common.go
```diff
@@ -96,12 +96,14 @@ func expandGroupOptions(opts map[string]string) (map[string]string, error) {
...
 		} else {
-			for _, option := range optionSet {
+			for _, option := range optionSet.Options {
 				switch val {
 				case "on":
 					options[option.Name] = option.OnValue
 				case "off":
-					options[option.Name] = option.OffValue
+					if volinfo != nil {
+						delete(volinfo.Options, option.Name)
```
prashanthpai: This doesn't feel like the right place to do this. Now there shall be two places where `volinfo.Options` gets modified - once here and again later in the calling function.

`expandGroupOptions`, as the name says is meant to only expand the group options. Do not pass `Volinfo` here. Instead, you can find the default value of each option and explicitly set that in the `options` map. 

```go
case "off":
    op, err := xlator.FindOption(option.Name)
    if err != nil {
        return err
    }
    options[option.Name] = op.DefaultValue
```

A better place for doing `delete(volinfo.Options, option.Name)` would be in the volume `reset` operation when done on a group option. I don't think reset has been implemented yet.

glusterd2/commands/volumes/volume-option.go
```diff
@@ -18,12 +18,22 @@ import (
 
 func optionSetValidate(c transaction.TxnCtx) error {
 
+	var volname string
```
prashanthpai: See comment above. This code movement isn't required here.

pkg/api/volume_req.go
```diff
@@ -50,13 +50,13 @@ type VolExpandReq struct {
...
-	OffValue string `json:"offvalue"`
 }
 
 // OptionGroup represents a group of options
 type OptionGroup struct {
-	Name    string         `json:"name"`
-	Options []VolumeOption `json:"options"`
+	Name         string         `json:"name"`
+	Options      []VolumeOption `json:"options"`
+	Description  string         `json:"description"`
```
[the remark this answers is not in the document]
aravindavk: I feel we should keep `Description` mandatory.

glusterd2/commands/volumes/grouped-options.go
```diff
@@ -5,50 +5,132 @@ import (
...
-		{"cluster.server-quorum-type", "server", "none"},
-		{"cluster.data-self-heal-algorithm", "full", "reset"},
-		{"cluster.locking-scheme", "granular", "full"},
-		{"cluster.shd-max-threads", "8", "1"},
-		{"cluster.shd-wait-qlength", "10000", "1024"},
-		{"features.shard", "on", "off"},
-		{"user.cifs", "off", "on"}},
-	"profile.test": {{"afr.eager-lock", "on", "off"},
-		{"gfproxy.afr.eager-lock", "on", "off"}},
+var defaultGroupOptions = map[string]*api.OptionGroup{
```
aravindavk: Can this be modified to read all groups from external json/yml file from predefined path(`<localstatedir>/group_profiles`)? 

    func GetDefaultGroupOptions()

glusterd2/commands/volumes/common.go
poornimag: ok. But there should be some place, which filters the options if it has default value, so that vol info doesn't display the options as modified? From the user's perspective its really confusing to see the options even after setting <profile> off.

glusterd2/commands/volumes/grouped-options.go
poornimag: ok, will do that as a dependent patch

glusterd2/commands/volumes/common.go
prashanthpai: > From the user's perspective its really confusing to see the options even after setting off

They can do a reset to not see those options.

pkg/api/volume_req.go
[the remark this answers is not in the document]
prashanthpai: Makes sense.

@poornimag  Leave it as is.

aravindavk:
rebase required.

prashanthpai:
retest this please

prashanthpai:
retest this please

prashanthpai:
retest this please

prashanthpai:
retest this please

PR #650, opened 2018-04-12
dht: Add a volume option dependency check

poornimag:

### DIFF
--- a/glusterd2/xlator/validation.go
+++ b/glusterd2/xlator/validation.go
@@ -77,6 +77,22 @@ func validateQuota(v *volume.Volinfo, key string, value string) error {
 	}
 }
 
+func validateDht(v *volume.Volinfo, key string, value string) error {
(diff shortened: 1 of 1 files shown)


glusterd2/xlator/validation.go
```diff
@@ -78,6 +78,22 @@ func validateQuota(v *volume.Volinfo, key string, value string) error {
 	return nil
 }
 
+func validateDht(v *volume.Volinfo, key string, value string) error {
+       var err error
+       if strings.Contains(key, "readdirplus-for-dir") {
```
prashanthpai: I can't find this key `readdirplus-for-dir` in glusterfs sources.

poornimag: Yes, the patch is under review and adding this here was a review comment: https://review.gluster.org/#/c/19559/

Issue #370, opened 2017-09-14
Support for hooks


aravindavk:
Existing `glusterd` provides the pre and post hooks for the following events

- add-brick
- create
- delete
- gsync-create
- remove-brick
- reset
- set
- start
- stop

In addition to above hooks, glusterd2 can provide hooks for all exposed APIs(for internal consumers only) using middleware plugins.

For example, give warning to users who try to stop a Gluster Volume if Geo-rep session is running. To implement this, hook required for stop/pre.

Middleware plugins can provide these hooks for internal users/plugins. It can act on incoming request(pre) and proceed to actual request only if pre satisfied. It can also act on the response(post).

When to use middleware plugins?
- If the plugin introducing new REST API then implement it as regular plugin
- If the plugin is adding validation/extra stuff to existing API then go for middleware plugin

PR #652, opened 2018-04-16
Add support for post hook scripts

aravindavk:
Fixes: #370

### DIFF
--- a/glusterd2/config.go
+++ b/glusterd2/config.go
@@ -86,6 +86,8 @@ func setDefaults() error {
 		config.SetDefault("localstatedir", wd)
 	}
 
+	config.SetDefault("hooksdir", config.GetString("localstatedir")+"/hooks")
+
 	if config.GetString(logging.DirFlag) == "" {
 		config.SetDefault(logging.DirFlag, path.Join(config.GetString("localstatedir"), "log"))
 	}
--- a/glusterd2/main.go
+++ b/glusterd2/main.go
@@ -173,6 +173,14 @@ func createDirectories() error {
 	dirs := []string{config.GetString("localstatedir"),
 		config.GetString("rundir"), config.GetString("logdir"),
 		path.Join(config.GetString("logdir"), "glusterfs/bricks"),
+		path.Join(config.GetString("hooksdir"), "create/post"),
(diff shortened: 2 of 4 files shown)


glusterd2/config.go
```diff
@@ -86,6 +87,10 @@ func setDefaults() error {
 		config.SetDefault("localstatedir", wd)
 	}
 
+	if config.GetString("hooksdir") == "" {
+		config.SetDefault("hooksdir", wd+"/hooks/1")
```
prashanthpai: This should default to `localstatedir` which is equivalent to `/var/lib/glusterd`.

What does the `1` refer to ? API version ?

glusterd2/config.go
```diff
@@ -44,6 +44,7 @@ func parseFlags() {
 	flag.String("workdir", "", "Working directory for GlusterD. (default: current directory)")
 	flag.String("localstatedir", "", "Directory to store local state information. (default: workdir)")
 	flag.String("rundir", defaultRunDir, "Directory to store runtime data.")
+	flag.String("hooksdir", "", "Directory to store Hook scripts.")
```
prashanthpai: I'd prefer that this not be made configurable. I don't see a use-case for making this option configurable because there are pre-designated directories (suc as `create/post`) inside hooks dir.

glusterd2/config.go
aravindavk: > What does the 1 refer to ? API version ?

Hooks API version used in Glusterd1(I think this versioning is based on the parameters sent to hook scripts, if parameters changes then version increment is required)

aravindavk:
Updated code as suggested

glusterd2/main.go
```diff
@@ -173,6 +173,14 @@ func createDirectories() error {
 	dirs := []string{config.GetString("localstatedir"),
 		config.GetString("rundir"), config.GetString("logdir"),
 		path.Join(config.GetString("logdir"), "glusterfs/bricks"),
+		path.Join(config.GetString("hooksdir"), "create/post"),
```
prashanthpai: These can be moved to a function in `glusterd2/events/hooks.go`

Issue #587, opened 2018-02-26
Man Page in glustercli


atinmu:
We need to revisit our existing man page document and add/modify details according to the glustercli format and semantics.

PR #653, opened 2018-04-16
manpage: Added support for manpage

aravindavk:
Updates: #587

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,8 +82,8 @@ install:
 	install -D $(CLI_BUILD) $(CLI_INSTALL)
 	install -D -m 0644 $(GD2CONF_BUILD) $(GD2CONF_INSTALL)
 	install -D -m 0644 $(CLI_BASH_COMPLETION_BUILD) $(CLI_BASH_COMPLETION_INSTALL)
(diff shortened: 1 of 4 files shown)


aravindavk:
retest this please

phlogistonjohn:
We've been discussing the same thing in Heketi land on and off for a while. I agree that no one in this day and age should have to edit the roff files directly. :-)

Both asciidoc and pandoc are mature and generally have OS packages. md2man seems nice and small(ish) but looks immature to me. Using markdown as the input format has the advantage of being simple and can be displayed directly via tools like github. asciidoc is powerful but has a bit of a unique syntax (IMO).

aravindavk:
> Editing roff files (the man file format) is a PITA.
Maybe we should try a tool like md2man to generate man pages from markdown. Or if we want an already established tools, asciidoc and pandoc have man as a target format.

> If we feel that these tools are too heavy (pandoc in particular) as build requirements, we could just commit the generated man files into the repo along with the markedup text files.

+1

We can also explore generating man page from CLI help message.

aravindavk:
Closing this for now, since this PR only addresses the Makefile related things. Can open new PR once manpage content is available

Issue #534, opened 2018-01-24
Support for Special commands


aravindavk:
Following commands are used in Geo-replication, Test scripts etc. Need to support these in Gd2 as well.

- To get Current working directory of Local Glusterd

        gluster system:: getwd

- To get Logdir of current Glusterd

        gluster --print-logdir

- To get Current glusterd UUID or Node ID

        gluster system:: uuid get

- To print statedump directory in local node

        gluster --print-statedumpdir

Following command is used by Geo-rep and Events API(not required with Gd2)

        system:: execute <command>
    
New API(`GET /v1/info` or just `GET /v1`) required to publish the current Glusterd's configuration and CLI can show required information.(Read only API, Set API not required)

PR #663, opened 2018-04-18
Use `glusterfsd --print-xlatordir`

prashanthpai:
Dependent change:
https://review.gluster.org/#/c/19905

### DIFF
--- a/glusterd2/xlator/load.go
+++ b/glusterd2/xlator/load.go
@@ -18,6 +18,8 @@ import (
 	"unsafe"
 
 	"github.com/gluster/glusterd2/glusterd2/xlator/options"
+
+	log "github.com/sirupsen/logrus"
 )
 
 const (
@@ -127,17 +129,14 @@ func loadXlator(xlPath string) (*Xlator, error) {
 
 func getXlatorsDir() string {
 
-	// glusterfs gets the path to xlator dir from a compile time flag named
-	// 'XLATORDIR' which gets passed through a -D flag to GCC. This isn't
-	// available to external programs via gluster CLI yet. When one or more
-	// versions of gluster are installed from source or otherwise, the
-	// following is the most fool-proof but hacky way to get the xlator dir
-	// location without making assumptions.
-
-	cmd := "strings -d `which glusterfsd` | awk '/glusterfs\\/.*\\/xlator$/'"
-	out, err := exec.Command("sh", "-c", cmd).Output()
+	out, err := exec.Command("glusterfsd", "--print-xlatordir").Output()
 	if err != nil {
-		return ""
+		// fallback to the old hack if https://review.gluster.org/19905 isn't present
+		cmd := "strings -d `which glusterfsd` | awk '/glusterfs\\/.*\\/xlator$/'"
+		out, err = exec.Command("sh", "-c", cmd).Output()
+		if err != nil {
+			return ""
+		}
 	}
 
 	return strings.TrimSpace(string(out))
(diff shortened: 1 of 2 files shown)


glusterd2/xlator/load.go
```diff
@@ -127,17 +129,14 @@ func loadXlator(xlPath string) (*Xlator, error) {
...
-	// glusterfs gets the path to xlator dir from a compile time flag named
-	// 'XLATORDIR' which gets passed through a -D flag to GCC. This isn't
-	// available to external programs via gluster CLI yet. When one or more
-	// versions of gluster are installed from source or otherwise, the
-	// following is the most fool-proof but hacky way to get the xlator dir
-	// location without making assumptions.
-
-	cmd := "strings -d `which glusterfsd` | awk '/glusterfs\\/.*\\/xlator$/'"
-	out, err := exec.Command("sh", "-c", cmd).Output()
+	out, err := exec.Command("gluster", "--print-xlatordir").Output()
```
atinmu: this brings in a hard dependency of shipping gluster cli always with GD2. I don't think its a good approach.

prashanthpai: Good catch!

aravindavk: Looks fine till this(https://github.com/gluster/glusterd2/issues/534) gets implemented

amarts: Good catch @atinmu but for atleast another 3 months, for users using/testing the feature, we should have an option to not crash. Should be very useful IMO.

glusterd2/xlator/load.go
```diff
@@ -148,10 +147,10 @@ func getXlatorsDir() string {
 func loadAllXlators() (map[string]*Xlator, error) {
 
 	xlatorsDir := getXlatorsDir()
-
 	if xlatorsDir == "" {
 		return nil, fmt.Errorf("No xlators dir found")
 	}
+	log.WithField("xlatordir", xlatorsDir).Debug("Xlators dir found")
```
amarts: If you need to handle #666 then you should check for version < 4.0.0 here?

glusterd2/xlator/load.go
atinmu: I'm OK to accept this patch as long as there's an issue filed to correct this later.

aravindavk: #534 captures these requirements.

prashanthpai:
@atinmu @aravindavk I have moved the required options from gluster cli to `glusterfsd` binary: https://review.gluster.org/#/c/19905

glusterd2/xlator/load.go
prashanthpai: That's been fixed in https://github.com/gluster/glusterd2/pull/667

aravindavk:
> @atinmu @aravindavk I have moved the required options from gluster cli to glusterfsd binary: https://review.gluster.org/#/c/19905

Nice. Is it possible to add `--print-logdir` also to the `glusterfsd`? It is used in lot of places in glusterfs  regression tests.

prashanthpai:
> Is it possible to add --print-logdir also to the glusterfsd? It is used in lot of places in glusterfs regression tests.

Done. Updated https://review.gluster.org/#/c/19905

prashanthpai:
The CI tests are failing because the latest glusterfs nightly RPM contains this change: https://review.gluster.org/#/c/19734/ which broke starting of glusterd2.

prashanthpai:
retest this please

Issue #655, opened 2018-04-16
Support volfile generation for Thin Arbiter volume type


amarts:
More on what is thin arbiter @ gluster/glusterfs#352

### Support in volume set

User would give the details of tie-breaker for the whole cluster (this is current MVP), there may be an RFE for per volume tie-breaker. That is future.

`glustercli volume set all global.tie-breaker <IP>:<port>://<brick> (or similar).`

### support in volfile generation

`glustercli volume create dr-ta-vol replica 2 <brick0> <brick1> <brick2> <brick3> --type thin-arbiter`

This would generate the volfile of just 4 bricks, and a client volfile which looks like below

```
volume thin-arbiter-client
   type protocol/client
   option remote-host <tie-breaker-node IP>
   option remote-port 24007 # optional (should match what is provided in tie-breaker setup)
end-volume

volume cp-0 # brick0
  type protocol/client
end-volume 

volume cp-1 # brick1
  type protocol/client
end-volume 

volume replicate-0
  type cluster/replicate
  subvolumes cp-0 cp-1 thin-arbiter-client # Notice the order, they want to keep tie-breaker as last child 
end-volume

volume cp-2 # brick2
  type protocol/client
end-volume 

volume cp-3 # brick3
  type protocol/client
end-volume 

volume replicate-1
  type cluster/replicate
  subvolumes cp-2 cp-3 thin-arbiter-client # Notice the order, they want to keep tie-breaker as last child 
end-volume

.....
....
```

For the initial version the requirement is this. This issue would be enhanced if there are more requirements.

@aspandey @itisravi @karthik-us @pranithk

aravindavk:
Sent patch to support this feature in Volfile. https://github.com/gluster/glusterd2/pull/673

Assumptions:
- Thin Arbiter Brick will be set as Volume option `glustercli volume set <volname> cluster.replicate.thin-arbiter <host:brick>`
- Volume option name is `thin-arbiter` in `cluster/replicate`

To test the Volfile generation,
- Add "thin-arbiter" option to `cluster/replicate` options table
- Create Replica 2 Volume using CLI(Glusterd2 CLI)
- Set Volume option using above volume set command.

Generated Client Volfile will have Thin arbiter details in it.

Let us know if any changes required in the behavior.

prashanthpai:
**cc* @aspandey

amarts:
Noticed that patch does handle one thing more than what is described:

[...]

PR #673, opened 2018-04-23
volgen: Add support for Thin Arbiter Volfile

aravindavk:
Updates: #655

### DIFF
--- a/glusterd2/volgen2/clustergraph.go
+++ b/glusterd2/volgen2/clustergraph.go
@@ -9,6 +9,11 @@ import (
 	"github.com/pborman/uuid"
 )
 
+const (
+	thinArbiterOptionName  = "replicate.thin-arbiter"
+	thinArbiterDefaultPort = "24007"
+)
+
 type clusterGraphFilters struct {
 	onlyLocalBricks bool
 	noSubvolParent  bool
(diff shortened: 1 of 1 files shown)


prashanthpai:
**cc* @aspandey

glusterd2/volgen2/clustergraph.go
```diff
@@ -72,6 +72,32 @@ func clusterGraph(volfile *Volfile, dht *Entry, vol *volume.Volinfo, nodeid uuid
 					decommissionedBricks = append(decommissionedBricks, name)
 				}
 			}
+
+			// FIXME: Change the name of thin arbiter option
+			thinarbiter, exists := vol.Options["cluster.replicate.thin-arbiter"]
+			if exists && thinarbiter != "" {
+				taParts := strings.Split(thinarbiter, ":")
```
amarts: I would still like to keep it as 'host<:port>:/brick'  (whatever is the normal standard), so we can handle the case of different port of thin-arbiter from the day1 itself.
If its going to be difficult, it is ok, but assumption is most of the time user doesn't specify the port (in that case 24007 is fine), but if he specifies, we get control of handling this.

glusterd2/volgen2/clustergraph.go
```diff
@@ -72,6 +72,32 @@ func clusterGraph(volfile *Volfile, dht *Entry, vol *volume.Volinfo, nodeid uuid
 					decommissionedBricks = append(decommissionedBricks, name)
 				}
 			}
+
+			// FIXME: Change the name of thin arbiter option
```
prashanthpai: Both the `FIXME` needs to be handled before this can be merged.

prashanthpai:
@aravindavk Ravi gave this a try today. He also needs the value set by user for the thin-arbiter option to end up in xlator options in the volfile. It should've ended up in volfile (as per normal volume set code) but it didn't.

aravindavk:
Modified PR and addressed the review comments. Please review

PR #678, opened 2018-04-23
e2e: make a central, configurable, base workdir

phlogistonjohn:
Previously, e2e tests were hard-coding the base workdir in
the main_test.go file and in the config toml files. This change
centralizes the base path in a top-level variable and makes
it possible to specify custom values for the workdir. This
can be useful if /tmp is inappropriate for this task on a
particular system or if multiple e2e tests are to be run in
parallel so that the caller may specify unique paths for
each test.
The toml files are adjusted to contain relative paths.

Signed-off-by: John Mulligan <jmulligan@redhat.com>


prashanthpai:
add to whitelist

PR #680, opened 2018-04-24
[RFC] add some test infrastructure

obnoxxx:
This enriches the test infrastructure, adopting some of the mechanisms from the heketi project.
It creates a common "test.sh" base test runner that executes scripts in the ./test/ directory in order.

Several ad-hoc tests are consolidated into scripts in ./test/.

In effect the "verify" and "test" make targets are merged, and the functional test run is split out into its own target "functest".

Subsequent PRs will add shell script test (and needed fixes) as well as a govet test (and needed fixes).


prashanthpai:
add to whitelist

PR #681, opened 2018-04-24
second part of test infra: test shellscripts

obnoxxx:
This builds on PR #680 and adds a script for testing shell scripts.
It's a separate PR since it has to fix a lot of shell scripts.


kshlm:
add to whitelist

obnoxxx:
Thanks for merging #680! If it helps, I would be happy to rebase this on top master now.

obnoxxx:
Looking at the centos-ci output, I found that we need to install shellcheck in the centos-ci script in order for the majority of this new test to take effect in CI... I should update this PR with one additional patch.

obnoxxx:
ok. apparently the centos shellcheck is too old ... :-(
Changing the script to not use `-x` for now so centos-ci can pass.

obnoxxx:
finally, the centos-ci run looks completely good :-)

Issue #629, opened 2018-03-28
Consider removing logging from glustercli


prashanthpai:
Currently, glustercli creates and logs to a file named `glustercli.log` in current working directory. All errors/info/warnings can be printed to stdout and the verbosity of printing can be controlled via a command line option.

prashanthpai:
> @prashanthpai Are we sure we want to do that? what happen when an user sees an issue an report back with all the log files? Would we be able to any obvious issues at the cli end?

There's barely any processing at all on the CLI side. Things that would've been in the log will be in stdout instead. It is uncommon for similar tools to log. For example `etcdctl` and `heketi-cli` do not log. Further, there may not permission to create `glustercli.log` in places where the CLI tool is run. Currently it creates the file in $PWD.

aravindavk:
Looks like both @rishubhjain and @Madhu-1 are working on this same issue

https://github.com/gluster/glusterd2/pull/644 and https://github.com/gluster/glusterd2/pull/693

Please assign the issue before starting the work on the issue to avoid the conflict.

prashanthpai:
#693 got merged and it removes logging to a file.

PR #693, opened 2018-04-25
add verbose flag to cli to

Madhu-1:
log error message to stdout

fix for #629  

### DIFF
--- a/glustercli/cmd/root.go
+++ b/glustercli/cmd/root.go
@@ -13,7 +13,6 @@ var RootCmd = &cobra.Command{
 	Use:   "glustercli",
 	Short: "Gluster Console Manager (command line utility)",
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
-		//err := logging.Init(flagLogDir, flagLogFile, flagLogLevel, false)
 		err := logging.Init("", "stdout", flagLogLevel, false)
 		if err != nil {
 			fmt.Println("Error initializing log file ", err)
@@ -36,6 +35,7 @@ var (
 	flagCacert     string
 	flagInsecure   bool
 	flagLogLevel   string
+	verbose        bool
 )
 
 const (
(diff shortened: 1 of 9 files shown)


glustercli/cmd/volume-set.go
```diff
@@ -31,6 +31,7 @@ func init() {
 	volumeSetCmd.Flags().BoolVar(&flagSetAdv, "advanced", false, "Allow setting advanced options")
 	volumeSetCmd.Flags().BoolVar(&flagSetExp, "experimental", false, "Allow setting experimental options")
 	volumeSetCmd.Flags().BoolVar(&flagSetDep, "deprecated", false, "Allow setting deprecated options")
+	volumeSetCmd.Flags().BoolVarP(&verbose, "verbose", "v", false, "verbose output")
```
prashanthpai: This is done at a lot of places. Cobra has provision to set a flag on the root command and all subcommands will inherit that flag.

See `glustercli/cmd/root.go`:

```go
RootCmd.PersistentFlags().BoolVarP(...
```

glustercli/cmd/root.go
```diff
@@ -13,7 +13,8 @@ var RootCmd = &cobra.Command{
 	Use:   "glustercli",
 	Short: "Gluster Console Manager (command line utility)",
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
-		err := logging.Init(flagLogDir, flagLogFile, flagLogLevel, false)
+		//err := logging.Init(flagLogDir, flagLogFile, flagLogLevel, false)
```
prashanthpai: Remove the comment.

glustercli/cmd/root.go
```diff
@@ -13,7 +13,8 @@ var RootCmd = &cobra.Command{
...
+		err := logging.Init("", "stdout", flagLogLevel, false)
```
prashanthpai: We don't have to use `logging.Init()`. If we use logrus directly, we can customize the formatter too,

Madhu-1:
@prashanthpai  fixed review comments

glustercli/cmd/root.go
Madhu-1: I was intentionally left it,  wanted to make use of current function

PR #696, opened 2018-04-25
e2e: don't delete workdir when tearing down

phlogistonjohn:
Deleting the workdir when tearing down deletes the logs and
other evidence needed to debug failing tests. The e2e main
function already deletes the top-level working dir at the
start of the tests and we recently added the ability to
use custom workdirs for different test runs.

Signed-off-by: John Mulligan <jmulligan@redhat.com>


e2e/utils_test.go
```diff
@@ -211,7 +211,6 @@ func setupCluster(configFiles ...string) ([]*gdProcess, error) {
 func teardownCluster(gds []*gdProcess) error {
 	for _, gd := range gds {
 		gd.Stop()
-		gd.EraseWorkdir()
```
prashanthpai: `setupCluster()` and `teardownCluster()` are not used by all tests. You may want to fix this (remove EraseWorkdir) at other places too for completeness.

```sh
[ppai@gd2-1 glusterd2]$ grep -nR "EraseWorkdir" . | grep -v utils
grep: warning: ./vendor/github.com/coreos/etcd/cmd/etcd: recursive directory loop
./e2e/peer_ops_test.go:16:	defer g1.EraseWorkdir()
./e2e/peer_ops_test.go:22:	defer g2.EraseWorkdir()
./e2e/peer_ops_test.go:28:	defer g3.EraseWorkdir()
./e2e/restapi_auth_test.go:18:	defer g1.EraseWorkdir()
```

phlogistonjohn:
> setupCluster() and teardownCluster() are not used by all tests. You may want to fix this (remove EraseWorkdir) at other places too for completeness.

Thanks for the info! I'll look into that.

PR #702, opened 2018-04-26
cli: Add thin arbiter support

### LINKED ISSUE #701: Add CLI support for thin arbiter

prashanthpai:
Closes #701

### DIFF
--- a/glustercli/cmd/volume-create.go
+++ b/glustercli/cmd/volume-create.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/gluster/glusterd2/pkg/api"
 
@@ -21,6 +22,7 @@ var (
 	flagCreateTransport                                                             string
 	flagCreateForce                                                                 bool
 	flagCreateAdvOpts, flagCreateExpOpts, flagCreateDepOpts                         bool
+	flagCreateThinArbiter                                                           string
 
 	volumeCreateCmd = &cobra.Command{
 		Use:   "create <volname> <brick> [<brick>]...",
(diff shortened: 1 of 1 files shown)


glustercli/cmd/volume-create.go
```diff
@@ -110,3 +128,19 @@ func volumeCreateCmdRun(cmd *cobra.Command, args []string) {
...
+
+	s := strings.Split(thinArbiter, ":")
+	if len(s) != 2 && len(s) != 3 {
+		return fmt.Errorf("Thin arbiter brick must be of the form <host>:<brick> or <host>:<brick>:<port>")
+	}
+
+	// TODO: Handle this in a generic way, just like other volume set
+	// options that we're gonna allow later.
+	req.Options = map[string]string{
+		"afr.thin-arbiter": thinArbiter,
```
aravindavk: Use "replicate.thin-arbiter" so that Volgen automatically adds to Volfile

aravindavk: Looks like these options may become issue with afr prefix. None of the options added with afr prefix will appear in generated volfile under `type cluster/replicate`

prashanthpai: Okay. "Replicate" will work as the afr xlator is loaded under both (afr and replicate) names: https://github.com/gluster/glusterd2/blob/master/glusterd2/xlator/load.go#L169

aravindavk: Not related to this patch, but other replicate options set accepts both `afr.*` and `replicate.*`, any suggestion to normalize these during Volume set?

prashanthpai: We can take a call on that later may be. `replicate` sounds more understandable though than `afr`. The same applies to `dht` vs `distribute`.

PR #703, opened 2018-04-26
set quota limit from api.

harigowtham:
involes creating an aux mount and setting the xattr through it.
umounting the aux mount and then storing the path on the etcd for
quota list.

cli yet to be worked.


harigowtham:
@prashanthpai @aravindavk @sanoj-unnikrishnan @kshlm please do take a look.

prashanthpai:
@harigowtham The tests are failing. Please test them locally by running "make test"

PR #708, opened 2018-04-26
Provide actors/hooks for volume set operation

### LINKED ISSUE #664: Provision to allow plugins to register callbacks for volume set operation
Plugins should able to execute some action when user sets a set of specific volume options.

prashanthpai:
...that xlators can hook into to perform some custom action on nodes
that make up the volume. Actions can be something like starting or
stopping a daemon.

Closes #664 

### DIFF
--- a/glusterd2/commands/volumes/volume-option.go
+++ b/glusterd2/commands/volumes/volume-option.go
@@ -9,6 +9,8 @@ import (
 	restutils "github.com/gluster/glusterd2/glusterd2/servers/rest/utils"
 	"github.com/gluster/glusterd2/glusterd2/transaction"
 	"github.com/gluster/glusterd2/glusterd2/volume"
+	"github.com/gluster/glusterd2/glusterd2/xlator"
+	"github.com/gluster/glusterd2/glusterd2/xlator/options"
 	"github.com/gluster/glusterd2/pkg/api"
 	"github.com/gluster/glusterd2/pkg/errors"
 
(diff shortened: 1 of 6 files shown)


glusterd2/commands/volumes/volume-option.go
```diff
@@ -107,11 +167,15 @@ func volumeOptionsHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	txn.Steps = []*transaction.Step{
-		lock,
 		{
 			DoFunc: "vol-option.Validate",
 			Nodes:  []uuid.UUID{gdctx.MyUUID},
 		},
+		{
+			DoFunc: "vol-option.XlatorAction",
```
kshlm: This needs UndoFunc as well. ActorFuncs can do changes that will need to be undone on any failure.

glusterd2/commands/volumes/volume-option.go
```diff
@@ -91,9 +134,26 @@ func volumeOptionsHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	lock, unlock, err := transaction.CreateLockSteps(volname)
+	lock, unlock := transaction.CreateLockFuncs(volname)
```
kshlm: Since this is such a repeated pattern, I'm thinking of extending the transaction framework to make this simpler.
Like so,
```go
// NewTxnWithLock creates a Txn with the provided context, and immediately obtains cluster locks on the provided lockIDS, and attaches the obtained locks to the Txn
func NewTxnWithLock(ctx context.Context, lockIDS ...string) (*Txn, error)
// Lock obtains cluster locks on the provided lockID (and lockIDS), and attaches the obtained locks to the Txn
func (t *Txn)Lock(lockIDS ...string) error
// Done frees up any held locks and cleans up the Txn
// Note: Done replaces Cleanup
func (t *Txn)Done
```

This would create the following pattern for transactions that require locks.
```go

initialLockIDS = parseReq(req)
// Haven't yet looked up the actual resources in the store
txn, _ = transaction.NewTxnWithLocks(ctx, initialLockIDS...)
defer txn.Done()

// Lookup resources in store
moreLockIDs = parseResourcesAndGetLockIDS()
_ = txn.Lock(moreLockIDs)

// Setup rest of the transaction and perform the actual transaction
txn.Steps = []transaction.Step{...}
_ = txn.Do()
```
Transactions that don't require locks continue as before.

How does this sound?

prashanthpai: Sounds good to me. However, we need the ability to take such locks in handlers that do not necessarily need a transaction, for example updating peer metadata.

glusterd2/commands/volumes/volume-option.go
prashanthpai: @kshlm Done

PR #719, opened 2018-05-03
cli: Display Hostname instead of Peer ID in Volume info

aravindavk:
Also changed the Volume info output Field State to Status

### DIFF
--- a/glustercli/cmd/volume.go
+++ b/glustercli/cmd/volume.go
@@ -251,9 +251,9 @@ func volumeInfoDisplay(vol api.VolumeGetResp) {
 	for sIdx, subvol := range vol.Subvols {
 		for bIdx, brick := range subvol.Bricks {
 			if brick.Type == api.Arbiter {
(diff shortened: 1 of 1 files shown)


glustercli/cmd/volume.go
```diff
@@ -245,15 +245,15 @@ func volumeInfoDisplay(vol api.VolumeGetResp) {
 	fmt.Println("Volume Name:", vol.Name)
 	fmt.Println("Type:", vol.Type)
 	fmt.Println("Volume ID:", vol.ID)
-	fmt.Println("State:", vol.State)
+	fmt.Println("Status:", vol.State)
```
prashanthpai: I prefer "state" over "status" as it sounds more natural:

`The volume xyz is in started state` as against `The volume xyz is in started status`

But I understand that this change is driven by backwards compatibility.

aravindavk: Ok, I will revert this back to `State`

aravindavk: Done

aravindavk:
retest this please

PR #722, opened 2018-05-03
check for duplicate brick entry

Madhu-1:
check for the duplicate entry of the brick in volume create.

**example cli command for volume create** 
cli volume create testvolume node1:/brick1 node1:/brick1 --force

check for the duplicate entry of the brick in volume expand (provided brick entry already present in the volume)

**example cli command for volume add-brick** 
cli add volume add-brick testvolume node1:/brick1 --force (brick1 already present in volume)

for both example server should return an error message.


Madhu-1:
fix for #723 

@prashanthpai  @kshlm @aravindavk  , please review

glusterd2/commands/volumes/volume-expand.go
```diff
@@ -47,6 +48,26 @@ func volumeExpandHandler(w http.ResponseWriter, r *http.Request) {
...
+	volinfo, err := volume.GetVolume(volname)
+	if err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+
+	for index := range volinfo.Subvols {
+		for _, brick := range volinfo.Subvols[index].Bricks {
+
+			for _, b := range req.Bricks {
```
aravindavk: duplicate entries need to be checked within `req.Bricks` also.

Madhu-1: @aravindavk added check

Madhu-1:
do not merge, I think need to add logic to extract node UUID if node IP or hostname is passed in the brick entry.

Madhu-1:
@prashanthpai  @kshlm  please review

PR #730, opened 2018-05-04
peer detach with default port 

Madhu-1:
peer detach with default port if the port is not provided in peer detach CLI command

### DIFF
--- a/glustercli/cmd/peer.go
+++ b/glustercli/cmd/peer.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/olekukonko/tablewriter"
+	"github.com/pborman/uuid"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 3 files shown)


glustercli/cmd/peer.go
```diff
@@ -127,3 +131,38 @@ var poolListCmd = &cobra.Command{
 		peerStatusHandler(cmd)
 	},
 }
+
+// getPeerID return peerId of host
+func getPeerID(host string) (string, error) {
+	// Get Peers list to find Peer ID
```
aravindavk: Add one more check to see the input is already peer id. (If Peer detach command is called by giving peer id itself)

    if uuid.Parse(host) != nil {
        return host, nil
    }

Madhu-1: do we need to compare that peerID with the available peer?
or let the server take care of it?

Madhu-1: added check

Madhu-1:
@aravindavk @prashanthpai @kshlm  please review

aravindavk:
retest this please

PR #731, opened 2018-05-04
volume state dump for default values 

Madhu-1:
if we pass --quota ==false the cmd.Flags().Changed will get hit 

get volume state dump with req.brick == true, when the quota is false and the client is empty  

### DIFF
--- a/glustercli/cmd/volume-statedump.go
+++ b/glustercli/cmd/volume-statedump.go
@@ -72,7 +72,9 @@ func volumeStatedumpCmdRun(cmd *cobra.Command, args []string) {
 		pid, _ := strconv.Atoi(s[1])
 		req.Client.Host = s[0]
 		req.Client.Pid = pid
-	} else {
+	}
+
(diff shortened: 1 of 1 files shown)


Madhu-1:
retest this please

Madhu-1:
@aravindavk  @prashanthpai @kshlm  please review

glustercli/cmd/volume-statedump.go
```diff
@@ -72,7 +72,9 @@ func volumeStatedumpCmdRun(cmd *cobra.Command, args []string) {
 		pid, _ := strconv.Atoi(s[1])
 		req.Client.Host = s[0]
 		req.Client.Pid = pid
-	} else {
+	}
+
+	if req.Quota == false && req.Client.Host == "" {
```
prashanthpai: Just the quota check is enough.

prashanthpai:
On second thought, this isn't required at all. Brick statedump is supposed to be taken only when no flags are specified.

Issue #709, opened 2018-04-27
cli: Gluster volume info should display volume options


prashanthpai:

prashanthpai:
Implemented in https://github.com/gluster/glusterd2/pull/747

PR #747, opened 2018-05-11
display volume options in CLI

Madhu-1:

### DIFF
--- a/glustercli/cmd/volume.go
+++ b/glustercli/cmd/volume.go
@@ -254,13 +254,16 @@ func volumeInfoDisplayNumbricks(vol api.VolumeGetResp) {
 }
 
 func volumeInfoDisplay(vol api.VolumeGetResp) {
-
 	fmt.Println()
 	fmt.Println("Volume Name:", vol.Name)
 	fmt.Println("Type:", vol.Type)
 	fmt.Println("Volume ID:", vol.ID)
 	fmt.Println("State:", vol.State)
 	fmt.Println("Transport-type:", vol.Transport)
+	fmt.Println("Options:")
+	for key, value := range vol.Options {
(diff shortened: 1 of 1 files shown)


Madhu-1:
Fixes #709

Madhu-1:
@prashanthpai please review

glustercli/cmd/volume.go
```diff
@@ -261,6 +261,10 @@ func volumeInfoDisplay(vol api.VolumeGetResp) {
 	fmt.Println("Volume ID:", vol.ID)
 	fmt.Println("State:", vol.State)
 	fmt.Println("Transport-type:", vol.Transport)
+	fmt.Println("Number of Options set:", len(vol.Options))
+	for key, value := range vol.Options {
+		fmt.Println(key, "\t", value)
```
prashanthpai: This is how it looks in glusterd1:

```
Volume Name: localdata
Type: Distribute
Status: Started
Number of Bricks: 1
Transport-type: tcp
Bricks:
Brick1: 10.1.30.130:/glu-export
performance.cache-refresh-timeout     1sec
cluster.stripe-block-size             128KB
cluster.self-heal-window-size         16
network.frame-timeout                 1800
network.ping-timeout                  42
auth.allow                            *
auth.reject                           None 
performance.cache-size                32MB
performance.write-behind-window-size  1MB
diagnostics.latency-measurement       off
diagnostics.dump-fd-stats             off
diagnostics.brick-log-level           NONE
diagnostics.client-log-level          NONE
nfs.enable-ino32                      off
```

The number of option set is not very useful.

You can do it this way in glusterd2:
```
Transport-type: tcp
Options:
    key1: value1
    key2: value2
```
i.e prefix a tab before key.

PR #635, opened 2018-04-03
Validate holding lock for volume{expand,delete,option}

prashanthpai:
Updates #510

PR #753, opened 2018-05-14
GD2/rebalance: Rebalance plugin

nbalacha:
Initial code for the rebalance start, stop and status functionality.

Adding Akarsha Rai as an author as she did the bulk of the original work.

Author: akrai@redhat.com

Signed-off-by: N Balachandran <nbalacha@redhat.com>

### DIFF
--- a/pkg/errors/error.go
+++ b/pkg/errors/error.go
@@ -40,4 +40,5 @@ var (
 	ErrUnmarshallFailed        = errors.New("failed to unmarshall from json")
 	ErrClusterNotFound         = errors.New("Cluster instance not found in store")
 	ErrDuplicateBrickPath      = errors.New("Duplicate brick entry")
+	ErrVolFileNotFound         = errors.New("volume file not found")
 )
--- a/plugins/rebalance/api/req_resp.go
+++ b/plugins/rebalance/api/req_resp.go
@@ -1,0 +1,74 @@
+package rebalance
+
+import (
+	"github.com/pborman/uuid"
+)
+
+// Status represents Rebalance Status
+type Status uint64
+
+const (
+	// NotStarted should be set only for a node in which rebalance process is not started
+	NotStarted Status = iota
+	// Started should be set only for a node that has been just started rebalance process
+	Started
+	// Stopped should be set only for a node that has been just stopped rebalance process
+	Stopped
+	// Complete should be set only for a node that the rebalance process is completed
+	Complete
+	// Failed should be set only for a node that are failed to run rebalance process
+	Failed
+)
+
+// Command represents Rebalance Commands
+type Command uint64
+
+const (
+	// CmdNone should be set only when given cmd is none
+	CmdNone Command = iota
+	// CmdStart should be set only when given cmd is rebalance start
+	CmdStart
+	// CmdStop should be set only when given cmd is rebalance stop
+	CmdStop
+	// CmdStatus should be set only when given cmd is rebalance status
+	CmdStatus
+	// CmdFixlayoutStart should be set only when given cmd is rebalance fix-layout start
+	CmdFixLayoutStart
+	// CmdStartForce should be set only when given cmd is rebalance start force
+	CmdStartForce
+)
+
+// NodeInfo represents Node information needed to store
+type RebalNodeStatus struct {
+	NodeID            uuid.UUID
+	Status            string // This is per Node Rebalance daemon Status
+	RebalancedFiles   string
+	RebalancedSize    string
+	LookedupFiles     string
+	SkippedFiles      string
+	RebalanceFailures string
+	ElapsedTime       string
+	TimeLeft          string
+}
+
+// RebalInfo represents Rebalance details for a node
+type RebalInfo struct {
+	Volname     string
+	Status      Status
+	Cmd         Command
+	RebalanceID uuid.UUID
+	CommitHash  uint64
+	RebalStats  []RebalNodeStatus
+}
+
+// RebalInfo represents Rebalance details for a node
+type RebalStatus struct {
+	Volname     string
+	RebalanceID uuid.UUID
+	Nodes       []RebalNodeStatus
+}
+
+// StartReq represents Rebalance Start Request
+type StartReq struct {
+	Option string `omitempty`
+}
--- a/plugins/rebalance/init.go
+++ b/plugins/rebalance/init.go
@@ -1,0 +1,60 @@
+package rebalance
+
+import (
+	"github.com/gluster/glusterd2/glusterd2/servers/rest/route"
+	"github.com/gluster/glusterd2/glusterd2/transaction"
+	"github.com/gluster/glusterd2/pkg/sunrpc"
+	"github.com/gluster/glusterd2/pkg/utils"
+	rebalanceapi "github.com/gluster/glusterd2/plugins/rebalance/api"
+)
+
+// Plugin is a structure which implements GlusterdPlugin interface
+type Plugin struct {
+}
+
+// Name returns name of plugin
+func (p *Plugin) Name() string {
+	return "rebalance"
+}
+
+// SunRPCProgram returns sunrpc program to register with Glusterd
+func (p *Plugin) SunRPCProgram() sunrpc.Program {
+	return nil
+}
+
+// RestRoutes returns list of REST API routes to register with Glusterd
+func (p *Plugin) RestRoutes() route.Routes {
+	return route.Routes{
+		route.Route{
+			Name:        "RebalanceStart",
+			Method:      "POST",
+			Pattern:     "/volumes/{volname}/rebalance/start",
+			Version:     1,
+			RequestType: utils.GetTypeString((*rebalanceapi.StartReq)(nil)),
+			//			ResponseType: utils.GetTypeString((*rebalanceapi.RebalInfo)(nil)),
+			HandlerFunc: rebalanceStartHandler},
+		route.Route{
+			Name:    "RebalanceStop",
+			Method:  "POST",
+			Pattern: "/volumes/{volname}/rebalance/stop",
+			Version: 1,
+			//			ResponseType: utils.GetTypeString((*rebalanceapi.RebalInfo)(nil)),
+			HandlerFunc: rebalanceStopHandler},
+		route.Route{
+			Name:    "RebalanceStatus",
+			Method:  "GET",
+			Pattern: "/volumes/{volname}/rebalance",
+			Version: 1,
+			//			ResponseType: utils.GetTypeString((*rebalanceapi.RebalInfo)(nil)),
+			HandlerFunc: rebalanceStatusHandler},
+	}
+}
+
+// RegisterStepFuncs registers transaction step functions with
+// Glusterd Transaction framework
+func (p *Plugin) RegisterStepFuncs() {
+	transaction.RegisterStepFunc(txnRebalanceStart, "rebalance-start")
+	transaction.RegisterStepFunc(txnRebalanceStop, "rebalance-stop")
+	transaction.RegisterStepFunc(txnRebalanceStatus, "rebalance-status")
+	transaction.RegisterStepFunc(txnRebalanceStoreDetails, "rebalance-store")
+}
--- a/plugins/rebalance/process.go
+++ b/plugins/rebalance/process.go
@@ -1,0 +1,110 @@
+package rebalance
+
+import (
+	"bytes"
+	"fmt"
+	"net"
+	"os/exec"
+	"path"
+
+	"github.com/gluster/glusterd2/glusterd2/gdctx"
+	rebalanceapi "github.com/gluster/glusterd2/plugins/rebalance/api"
+
+	"github.com/cespare/xxhash"
+	config "github.com/spf13/viper"
+)
+
+const (
+	glusterfsBin = "glusterfs"
+)
+
+// Process type represents information about rebalance process
+type Process struct {
+	binarypath     string
+	args           string
+	socketfilepath string
+	pidfilepath    string
+	rInfo          rebalanceapi.RebalInfo
+}
+
+// Name returns the process name
+func (r *Process) Name() string {
+	return "rebalance"
+}
+
+// Path returns absolute path to the binary of rebalance process
+func (r *Process) Path() string {
+	return r.binarypath
+}
+
+// Args returns arguments to be passed to rebalance process
+func (r *Process) Args() string {
+
+	volfileserver, port, _ := net.SplitHostPort(config.GetString("clientaddress"))
+	if volfileserver == "" {
+		volfileserver = "localhost"
+	}
+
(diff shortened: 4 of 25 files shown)


prashanthpai:
add to whitelist

pkg/errors/error.go
```diff
@@ -40,4 +40,6 @@ var (
 	ErrUnmarshallFailed        = errors.New("failed to unmarshall from json")
 	ErrClusterNotFound         = errors.New("Cluster instance not found in store")
 	ErrDuplicateBrickPath      = errors.New("Duplicate brick entry")
+	ErrVolNotDistribute        = errors.New("Not a distribute volume")
```
kshlm: It will be preferrable to define this withing these errors in the rebalance package itself.

plugins/rebalance/api/req_resp.go
```diff
@@ -0,0 +1,77 @@
...
+	CmdFixlayoutStart
+	// CmdStartForce should be set only when given cmd is rebalance start force
+	CmdStartForce
+)
+
+// NodeInfo represents Node information needed to store
+type RebalNodeStatus struct {
+	NodeID            uuid.UUID
+	Status            string  // This is per Node Rebalance daemon Status
+	RebalancedFiles   string
```
kshlm: Just wondering why all members are strings? Wouldn't it be better for these to use relevant types?

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,323 @@
...
+
+	ctx := r.Context()
+	logger := gdctx.GetReqLogger(ctx)
+
+	// collect inputs from url
+	volname := mux.Vars(r)["volname"]
+
+	var req rebalanceapi.StartReq
+
+//  TODO : How do I pass these options?
```
kshlm: You can pass these to the transaction by setting them in the transaction context.

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,323 @@
...
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+
+	err = txn.Do()
+	if err != nil {
+		/* TODO: Need to handle failure case. Unlike other process,
+		 * rebalance process is one per node and depends on volfile change.
+		 * Need to handle scenarios where process is started in
+		 * few nodes and failed in few others */
```
kshlm: If you create and set undo-funcs in the transaction steps, they should automatically take care of stopping any started rebalance processes.

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	// Check for remove- brick pending
+	//TODO
+
+	// A simple transaction to start rebalance on all nodes
+	txn := transaction.NewTxn(ctx)
+	defer txn.Cleanup()
+
+	lock, unlock, err := transaction.CreateLockSteps(volname)
```
kshlm: There has been some changes regarding when we do transaction locking and unlocking.
Refer #510 for why, and #634 and #635 for how.
In short, we need to lock, unlock outside of the transaction, before we do the first GetVolume.
Create the lock/unlock funcs, lock and defer unlock.
```go
lock, unlock, err := transaction.CreateLockFuncs(volname)
lock(ctx)
defer unlock(ctx)
```
@prashanthpai began doing changes

plugins/rebalance/utils.go
```diff
@@ -0,0 +1,108 @@
...
+        if err != nil {
+                log.WithError(err).Error("Couldn't add rebalance info to store")
+                return err
+        }
+        return nil
+}
+
+
+func checkCmd(req *rebalanceapi.StartReq) rebalanceapi.Command {
+	var rebalance rebalanceapi.RebalInfo
```
aravindavk: extra variable can be removed

    if req.Fixlayout == true {
        return rebalanceapi.CmdFixlayoutStart
    } else if req.Force == true {
        return rebalanceapi.CmdStartForce
    } 
    return rebalanceapi.CmdStart

plugins/rebalance/transaction.go
```diff
@@ -0,0 +1,257 @@
...
+                rebalNodeStatus.LookedupFiles = rspDict["lookups"]
+                rebalNodeStatus.SkippedFiles = rspDict["skipped"]
+                rebalNodeStatus.RebalanceFailures = rspDict["failures"]
+                rebalNodeStatus.ElapsedTime = rspDict["run-time"]
+
+        }
+
+        //TODO: Update the rebalinfo.Status
+
+        rebalinfo.RebalStats = rebalNodeStatus;
```
aravindavk: this is over written by each node's transaction step.

plugins/rebalance/transaction.go
```diff
@@ -0,0 +1,257 @@
...
+
+        } else {
+
+                //Send the status request
+
+                reqDict := make(map[string]string)
+
+                xlator_name = fmt.Sprintf("%s-distribute", volname)
+
+                // TODO: FIX THIS!!!
```
aravindavk: what fix is required here?

plugins/rebalance/transaction.go
```diff
@@ -0,0 +1,257 @@
...
+        var rsp brick.GfBrickOpRsp
+        err = client.Call("Brick.OpBrickXlatorDefrag", req, &rsp)
+        if err != nil || rsp.OpRet != 0 {
+                c.Logger().WithError(err).WithField(
+                        "volume", volname).Error("failed to send rebalance stop RPC")
+                return err
+        }
+
+
+        // TODO : Send a response
```
aravindavk: nit: commented code

plugins/rebalance/utils.go
kshlm: Two suggestions here.
1. You don't need the RebalInfo variable. You directly return the relevant Command.
2. The preferred way to write long if else chains it to use switch.
```go
switch {
  case req.Fixlayout:
    return rebalanceapi.CmdFixLayoutStart
  // other conditions as cases
  default:
    return rebalanceapi.CmdStart
}
// No final return required
```

plugins/rebalance/transaction.go
```diff
@@ -0,0 +1,257 @@
...
+
+        }
+
+        // What is the expected behaviour if the process does not exist (rebalance has completed)?
+        // Will it restart the process?
+
+	rebalanceProcess, err := NewRebalanceProcess(rebalinfo)
+	if err != nil {
+
+                // TODO: Send the stored Info?
```
aravindavk: This is different failure, will not check the existence of rebal process here. Only possible failure I see is if `glusterfs` path is not not known

plugins/rebalance/utils.go
```diff
@@ -0,0 +1,108 @@
...
+	rebalanceapi "github.com/gluster/glusterd2/plugins/rebalance/api"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	rebalancePrefix string = "rebalance/"
+)
+
+// TimeVal represents the time value
+type TimeVal struct {
```
kshlm: You don't really need a seperate type here as it isn't used else where.
You can do it with local variables in the setCommitHash function.

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,323 @@
...
+
+	txn.Nodes = vol.Nodes()
+	txn.Steps = []*transaction.Step{
+		lock,
+		{
+			DoFunc: "rebalance-status",
+			Nodes:  txn.Nodes,
+		},
+		{
+			DoFunc: "rebalance-store",
```
aravindavk: I think we can skip this step. Every time when status is requested, we get status from running rebalance daemon(except complete), so we can skip storing in etcd.

plugins/rebalance/utils.go
```diff
@@ -0,0 +1,108 @@
...
+	   consecutive rebalances are separated by exactly 2^29 seconds - about
+	   17 years - and even then there's only a 1/8 chance of a collision in
+	   the low order bits.  It's far more likely that this code will have
+	   changed completely by then.  If not, call me in 2031.
+	   P.S. Time zone changes?  Yeah, right.
+	*/
+
+	tv := new(TimeVal)
+	t := time.Now()
+	tv.TVsec = uint64(t.Day() * t.Hour() * t.Minute() * t.Second())
```
kshlm: Are you expecting to get the number of unix seconds here? If so this is not correct. You can use `t.Unix()` or `t.UnixNano()` to get more granular values.

plugins/rebalance/rest.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	rebalinfo, err := GetRebalanceInfo(volname)
+	if err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusNotFound, err)
+		return
+	}
+
+	// Check rebalance process is started or not
+	if rebalinfo.Status != rebalanceapi.Started {
```
aravindavk: you can also add `rebalinfo.Status != rebalanceapi.Completed`

plugins/rebalance/api/req_resp.go
kshlm: NVM. I got it. Because this is how the xdr dict contains them.
But it would still be preferrable to convert to proper types when sending them to the ReST clients. JSON handles types much better than our dicts.

plugins/rebalance/process.go
```diff
@@ -0,0 +1,111 @@
...
+	r.socketfilepath = fmt.Sprintf("%s-rebalance-%x.socket", glusterdSockDir, xxhash.Sum64String(gdctx.MyUUID.String()))
+	return r.socketfilepath
+}
+
+// PidFile returns path to the pid file of rebalance process
+func (r *Process) PidFile() string {
+	if r.pidfilepath != "" {
+		return r.pidfilepath
+	}
+//	pidDir := path.Join(config.GetString("rundir"), "gluster")
```
aravindavk: nit: commented code

plugins/rebalance/init.go
```diff
@@ -0,0 +1,60 @@
...
+// RestRoutes returns list of REST API routes to register with Glusterd
+func (p *Plugin) RestRoutes() route.Routes {
+	return route.Routes{
+		route.Route{
+			Name:         "RebalanceStart",
+			Method:       "POST",
+			Pattern:      "/volumes/{volname}/rebalance/start",
+			Version:      1,
+			RequestType:  utils.GetTypeString((*rebalanceapi.StartReq)(nil)),
+//			ResponseType: utils.GetTypeString((*rebalanceapi.RebalInfo)(nil)),
```
aravindavk: uncomment

plugins/rebalance/api/req_resp.go
nbalacha: I would prefer proper types as well but could not get them to work with the xdr. If you can point me to how to do this, I can make the changes.

pkg/errors/error.go
nbalacha: Will do.

plugins/rebalance/rest.go
nbalacha: We would need to get the status from any completed rebalance processes as well and those will be stored in etcd. The rebalance processes on different nodes may not finish at the same time. Do you mean we query etcd only if the process is not running?

plugins/rebalance/transaction.go
nbalacha: How do I check the existence of the process ? I do not want it to restart if it has already completed.

plugins/rebalance/transaction.go
nbalacha: I wasn't sure if this was the best way to convert a const uint64 to a string (this was the only way I got it to work)

plugins/rebalance/rest.go
aravindavk: `curl -i -XPOST <rebalance-url> -d '{"force": true}'`

- `-i` to display response headers
- "force" should be defined as `json:"force"` in StartReq

plugins/rebalance/transaction.go
aravindavk: `strconv.FormatUint` Ref: https://golang.org/pkg/strconv/

nbalacha: I tried that but it didn't work:
plugins/rebalance/transaction.go:88:51: cannot use rebalance.CmdStop (type rebalance.Command) as type uint64 in argument to strconv.FormatUint

nbalacha:
[root@rhgs313-6 rebalance]# gofmt -s -w *
[root@rhgs313-6 rebalance]#

It does not return any errors. Am I doing this correctly?

plugins/rebalance/utils.go
nbalacha: Done

aravindavk:
> [root@rhgs313-6 rebalance]# gofmt -s -w *
> [root@rhgs313-6 rebalance]#

> It does not return any errors. Am I doing this correctly?

Yes, it will return error only when there is a syntax error else it re-formats the file inplace.

plugins/rebalance/rest.go
nbalacha: That could also mean that the process failed , in which case it will also not running. So we usually only allow a stop if the status == started.

plugins/rebalance/utils.go
nbalacha: Done

plugins/rebalance/utils.go
nbalacha: Done

kshlm:
> [root@rhgs313-6 rebalance]# gofmt -s -w *
> [root@rhgs313-6 rebalance]#
> 
> It does not return any errors. Am I doing this correctly?

Depending on the editor you're using, I'd suggest that you set it up to run gofmt automatically on saving a go file. Better still setup the golang plugins/extensions for your editor which brings in this and more functionality, for eg. vim-go for Vim/Neovim or vscode-go for Visual Studio Code.

Running `make test` before submitting, runs the static checks and the basic unit-tests, including `gofmt`.

plugins/rebalance/rest.go
nbalacha: When does the unlock go through?

kshlm: When the handler function ends. The `defer` keyword adds the given function to be run after the current function calls return. You can defer as many functions as you want, and they will all get called, in LIFO order, at the end.

nbalacha:
Something has gone wrong here

Madhu-1:
@nbalacha you may need to fix golint and govet issues in this PR

plugins/rebalance/api/req_resp.go
```diff
@@ -0,0 +1,74 @@
...
+// RebalInfo represents Rebalance details for a node
+type RebalStatus struct {
+	Volname     string
+	RebalanceID uuid.UUID
+	Nodes       []RebalNodeStatus
+}
+
+// StartReq represents Rebalance Start Request
+type StartReq struct {
+	Option string `omitempty`
```
Madhu-1: This is not a valid json format 
```
`json:"option,omitempty"`
```

PR #763, opened 2018-05-17
txn: Aggregate results from nodes

prashanthpai:
This change ensures proper aggregation of results from
different peers when step of a transaction is executed.

This PR also makes following cleanup changes:
* Use the built-in "context" package
* Do not export GetStepFunc() and RunStepOn()
* Remove unused TxnCtx from TxnStepResp


prashanthpai:
> Looks good mostly. What do you think about using something like go-multierror, to aggregate errors and bubble it up. We can use that to return more specific information back to clients who sent us the request.

I started this thinking along the same lines but with a custom error type that `txn.Do()` returns. I'd prefer doing that in a separate PR though, if that's okay.

> Or if you believe that clients don't need such specific information, and if required they should look at the logs, that's fine as well

They do. For example, during a volume create txn that visits many nodes, it's necessary to let the client know which step failed on which nodes.

prashanthpai:
retest this please

prashanthpai:
retest this please

prashanthpai:
retest this please

prashanthpai:
retest this please

prashanthpai:
retest this please

glusterd2/transaction/transaction-rpc.proto
```diff
@@ -9,7 +9,6 @@ message TxnStepReq {
 
 message TxnStepResp {
   string Error = 1;
-  bytes Resp = 2; // Resp is JSON encoded TxnCtx
```
kshlm: Please mark the removed field and number as [reserved][1].

[1]: https://developers.google.com/protocol-buffers/docs/proto3#reserved

prashanthpai: Done.

PR #769, opened 2018-05-18
Configuration improvements

aravindavk:
If `PREFIX` is `/usr` then `logdir`, `rundir` and `localstatedir`
should not contain the prefix.

**Global default configuration**: Based on the `--prefix` used during
installation/make global config should be installed in
`$PREFIX/etc/glusterfs/glusterd2.toml`. All the default values from
code should be moved to this file.

**Removed logic to pick config file from current directory**: Since
Glusterd2 will be run as daemon or background process, automatically
picking config file from current directory will add confusion to
users. For development, if global config file not exists then it loads
config file from the current directory.

**Config file as argument**: Glusterd2 applies global configurations
first and then applies this configuration to overwrite the values
if any.


aravindavk:
retest this please

aravindavk:
glusterd2 binary looks for default config file in `$install_prefix/etc/glusterd2/glusterd2.toml` (`/etc/glusterd2/glusterd2.toml` if `prefix=/usr`). During development, config file is not installed so it looks for default config file in `glusterd2` binary directory. (`make glusterd2.toml`)

Do you have `glusterd2` binary in `/home/ppai/workdir/` directory?

aravindavk:
> Regardless of that --config file should work.

Default config is required. (Similar to `/etc/bashrc` and `~/.bashrc`). Without default config file, default configuration ends up in code.

Configuration specified in custom config will get priority.

> I have it in build/ directory in the repo. I don't do make install, just make and use a custom config file.

`make glusterd2.toml` will solve the issue.

prashanthpai:
> Default config is required

That should be in code, right ? If I copy glusterd2 binary to a machine, it should work with defaults, just like any normal go program.

> Configuration specified in custom config will get priority.

Right. Doesn't seem to be happening for me.

> make glusterd2.toml will solve the issue.

```sh
$ make glusterd2.toml
make: Nothing to be done for 'glusterd2.toml'.
$ ls build/
bash_completion  generate_bash_completion  glustercli  glustercli.sh  glusterd2  glusterd2.toml
```

This just generates `glusterd2.toml` in `build` directory.

Glusterd2 doesn't start for me.

aravindavk:
> That should be in code, right ? If I copy glusterd2 binary to a machine, it should work with defaults, just like any normal go program.

You need to copy default config file as well. Before this patch, glusterd2 binary was looking config file in multiple paths 1) /etc/glusterd2, 2) /usr/local/etc/glusterd2 and 3) current directory

If old config file exists in /etc/glusterd2/ and then if glusterd2 is installed as source install. it picks old config file by default.

> Right. Doesn't seem to be happening for me.

Custom config will be merged with default config. Default config is not found here.

All the defaults formed in code are relative to `workdir`, but not according to platform/os specific. For example, default logdir is `$workdir/log`. To make it platform specific, build tools should consider `prefix` and other inputs given during `make`. Default config is generated using these variables. 

> Glusterd2 doesn't start for me.

Copy `glusterd2.toml` file along with the `glusterd2` binary

prashanthpai:
> You need to copy default config file as well

That's really bad dependency. We should fix that. Sane defaults should be in code. Just copying the binary to a machine should work.

prashanthpai:
Please introduce code snippet this back:

```go
const (
	defaultLogLevel      = "debug"
	defaultClientAddress = ":24007"
	defaultPeerAddress   = ":24008"	
	defaultConfName      = "glusterd2"	
	defaultRunDir        = "/var/run/gluster"	
)
```

Not looking in the current directory is a good change.

aravindavk:
> That's really bad dependency. We should fix that. Sane defaults should be in code. Just copying the binary to a machine should work.

I don't think our goal is zero external dependency. Multiple lookup path for config adds confusion and may introduce bugs when stale config file exists in these paths.

This problem only exists during development, source install or rpm install will not have this issue.

> Sane defaults should be in code

Sure, we can replace default config by having them in code.

During `make glusterd2` we should generate Go code which will have the default values based on the inputs provided during make (like prefix, exec_prefix, etc). But note, the binary will not be platform independent with default values(Binary file needs to be regenerated for target platform)

prashanthpai:
> Multiple lookup path for config adds confusion and may introduce bugs when stale config file exists in these paths.

I'm all in favour of removing multiple config lookup paths. You can decide the default lookup path during build time. I just expect `glusterd2` and `glusterd2 --config whatever.toml` to work regardless of whether `/etc/usr/local/gd2/glusterd2.toml` exists or not.

Unlike C programs, Go programs are usually meant to work if you just copy the binary (from say github releases). We should never break this convenience/advantage that go provides. For that reason, defaults should be part of the code. Generating code is an overkill IMO but it's better than creating hard dependency on default config file.

> the binary will not be platform independent with default values(Binary file needs to be regenerated for target platform)

That's true anyway. A binary generated for linux-amd64 will run on any linux distro (fedora, centos etc).

prashanthpai:
And this is the order of preference that's desired:

```
glusterd2 --some-specific-option="some_value"
glusterd2 --config config_file.toml
glusterd2
```

The last invocation example should take defaults from code, not default conf location. That can get very confusing for developers.

When glusterd2 is deployed as an RPM:

The RPM will contain default config file and will place it at `/etc/glusterd2/glusterd2.toml` and the systemd service file shall invoke `glusterd2 --config /etc/glusterd2/glusterd2.toml`

aravindavk:
> I'm all in favour of removing multiple config lookup paths. You can decide the default lookup path during build time. I just expect glusterd2 and glusterd2 --config whatever.toml to work regardless of whether /etc/usr/local/gd2/glusterd2.toml exists or not.

If custom config(specified in `--config`) doesn't contain all the config items, possible ways to find the defaults is by default  config or in code. If the code is picking default values from default config, then it is not possible to make it work when default config not exists.

I will send patch to generate code with default values instead of default config.

> Unlike C programs, Go programs are usually meant to work if you just copy the binary (from say github releases). We should never break this convenience/advantage that go provides. For that reason, defaults should be part of the code. Generating code is an overkill IMO but it's better than creating hard dependency on default config file.

Not true for all kinds of Go programs, All html, css and js files need to be shipped with binary if it is web project.

prashanthpai:
> If custom config(specified in --config) doesn't contain all the config items, possible ways to find the defaults is by default config or in code

Right, get the rest from code.

>If the code is picking default values from default config, then it is not possible to make it work when default config not exists

Get default from code. If default config file exists, you can take from that.

> Not true for all kinds of Go programs, All html, css and js files need to be shipped with binary if it is web project.

True in our case :)

PR #773, opened 2018-05-21
validate volume name

### LINKED ISSUE #772: creation of volume with  special characters in volume name (except underscore and hyphen) should fail

Madhu-1:
It is necessary to validate volume name before volume creation.
Fixes #772 

### DIFF
--- a/glusterd2/commands/volumes/volume-create.go
+++ b/glusterd2/commands/volumes/volume-create.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"net/http"
 	"path/filepath"
+	"regexp"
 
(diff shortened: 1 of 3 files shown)


glusterd2/commands/volumes/volume-create.go
```diff
@@ -17,6 +18,12 @@ import (
 )
 
 func validateVolCreateReq(req *api.VolCreateReq) error {
+	reg := regexp.MustCompile("^[a-zA-Z0-9_-]+$")
```
aravindavk: Move this to global scope, outside this function

Madhu-1: done

PR #777, opened 2018-05-22
Added extra information to os/exec errors

aravindavk:
The err string returned by `exec.Command.{Run,Output,Wait,Start}` will
not contains `stderr`, Output in log/CLI will just display `exited with
status 1`. This patch adds the content of `stderr` along with the error
message.


aravindavk:
retest this please

glusterd2/daemon/daemon.go
```diff
@@ -73,15 +75,18 @@ func Start(d Daemon, wait bool) error {
...
+	err = utils.ExecuteCommandStart(cmd, &stderr)
 	if err != nil {
 		events.Broadcast(newEvent(d, daemonStartFailed, 0))
 		return err
 	}
 
 	if wait == true {
 		// Wait for the child to exit
-		errStatus := cmd.Wait()
+		errStatus := utils.ExecuteCommandWait(cmd, &stderr)
```
prashanthpai: The daemon framework will only run various daemons which don't have anything attached to it's STDOUT and STDERR. This change isn't necessary for daemon framework. Further, in this specific case, it's easier to read and understand `cmd.Wait()`

glusterd2/volume/fs_utils.go
```diff
@@ -60,11 +62,13 @@ func mountVolume(name string, mountpoint string) error {
...
 	args := strings.Fields(buffer.String())
 	cmd := exec.Command("glusterfs", args...)
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
 	if err := cmd.Start(); err != nil {
 		return err
 	}
 
-	return cmd.Wait() // glusterfs daemonizes itself
+	return utils.ExecuteCommandWait(cmd, &stderr) // glusterfs daemonizes itself
```
prashanthpai: Same here.

glusterd2/brick/glusterfsd.go
```diff
@@ -146,15 +146,11 @@ const BrickStartMaxRetries = 3
 const anotherEADDRINUSE = syscall.Errno(0x9E) // 158
 
 func errorContainsErrno(err error, errno syscall.Errno) bool {
-	exiterr, ok := err.(*exec.ExitError)
+	exiterr, ok := err.(*utils.ExecuteCommandError)
```
prashanthpai: Not needed here.

prashanthpai:
> Requesting you to use the utils* versions only where the error is logged/displayed and command writes something to stdout/stderr. This effectively means leave out places where cmd.Wait() is used explicitly.

@aravindavk Can you please update the PR ? This is a very useful change at places where we run shell commands (any process that won't become a daemon)

aravindavk:
Updated the code as suggested. Please review

plugins/georeplication/rest.go
```diff
@@ -601,8 +601,7 @@ func checkConfig(name string, value string) error {
 	if value != "" {
 		args = append(args, "--value", value)
 	}
-	_, err := exec.Command(gsyncdCommand, args...).Output()
-	return err
+	return utils.ExecuteCommandRun(gsyncdCommand, args...)
```
rishubhjain: utils.ExecuteCommandRun(gsyncdCommand, args...) returns two arguments while checkConfig is supposed to return just one argument.

aravindavk: `utils.ExecuteCommandRun` returns only one value. `utils.ExecuteCommandOutput` returns two

aravindavk:
retest this please

PR #780, opened 2018-05-22
makefile: support generating sources via a standalone target

phlogistonjohn:
I tried to run 'make test' out of a clean repo and it failed
due to missing source files that were being generated as
part of the build. I have tweaked the Makefile so that a
new target 'generated' is a prerequisite for the tests and
build.
In the future we might consider switching to calling something
like 'go generate' here. But I'm trying to keep this change
small.


prashanthpai:
retest this please

PR #781, opened 2018-05-22
convert daemon interface to use vector args instead of string

phlogistonjohn:
This change converts the Daemon interface to use an Args
function that returns a []string rather than a string.
It's generally good policy to keep your argv as a
distinct set of arguments rather than mush everything
together in a string that must be pulled apart again
later.
It's easier to see where each argument begins and ends,
it works better for (and futureproofs against) arguments
that may contain spaces, and can potentially avoid
security issues if some of the input strings contain
unusual sequences of characters, and it even ends up
needing less typing.

I had some questions around the storeddaemon and
events changes. I'd like to hear if anyone has opinions
about that part of the change.


prashanthpai:
> I had some questions around the storeddaemon and
events changes. I'd like to hear if anyone has opinions
about that part of the change.

The changes LGTM :+1: 

I'll wait for a day to see if @kshlm has something to say before I merge this.

PR #783, opened 2018-05-23
List devices

rishubhjain:
@aravindavk Please review


plugins/device/rest.go
```diff
@@ -104,3 +104,28 @@ func deviceAddHandler(w http.ResponseWriter, r *http.Request) {
...
+	logger := gdctx.GetReqLogger(ctx)
+	peerID := mux.Vars(r)["peerid"]
+	if uuid.Parse(peerID) == nil {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "Invalid peer-id passed in url")
+		return
+	}
+
+	devices, err := GetDevices(peerID)
+	if err != nil {
+		logger.WithError(err).WithField("peerid", peerID).Error(err)
```
prashanthpai: Pass an error string to Error() such as - `Failed to get devices for peer`

plugins/device/rest.go
```diff
@@ -104,3 +104,28 @@ func deviceAddHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	ctx := r.Context()
+	logger := gdctx.GetReqLogger(ctx)
+	peerID := mux.Vars(r)["peerid"]
+	if uuid.Parse(peerID) == nil {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "Invalid peer-id passed in url")
+		return
+	}
+
+	devices, err := GetDevices(peerID)
```
prashanthpai: Unrelated to this PR: I had a look at `GetDevices` which does `return nil, nil` at the end. That can be improved as follows:

```go
func GetDevices(peerID string) ([]deviceapi.Info, error) {

        peerInfo, err := peer.GetPeer(peerID)
        if err != nil {
                return nil, err
        }

        var deviceInfo []deviceapi.Info
        if devices, ok := peerInfo.MetaData["devices"]; ok {
                if err := json.Unmarshal([]byte(devices), &deviceInfo); err != nil {
                        return nil, err
                }
        }

        return deviceInfo, nil
}
```

As you have to update the PR to address the comment below, you might as well fix this one as well.

prashanthpai:
The REST API documentation needs to be regenerated.

plugins/device/init.go
```diff
@@ -27,6 +27,13 @@ func (p *Plugin) RestRoutes() route.Routes {
 			RequestType:  utils.GetTypeString((*deviceapi.AddDeviceReq)(nil)),
 			ResponseType: utils.GetTypeString((*deviceapi.AddDeviceResp)(nil)),
 			HandlerFunc:  deviceAddHandler},
+		route.Route{
+			Name:         "DeviceList",
+			Method:       "GET",
+			Pattern:      "/peers/{peerid}/devices",
```
aravindavk: can this be changed to `/devices/{peerid}` and `/devices`, when peerid not specified then return devices from all peers else only from that peer.

For uniformity, we can change `POST /peers/{peerid}/devices` to `POST /devices/{peerid}`

plugins/device/rest.go
prashanthpai: @rishubhjain The code on `GetDevices` uses `devices` as key, it's `_devices` at other places. Please find places where it's `devices` and change them to `_devices`

rishubhjain:
@prashanthpai If possible could you check again, I have used _devices only.

plugins/device/init.go
rishubhjain: Will address listing of all devices in another PR as it requires some discussion

plugins/device/init.go
```diff
@@ -22,11 +22,18 @@ func (p *Plugin) RestRoutes() route.Routes {
 		route.Route{
 			Name:         "DeviceAdd",
 			Method:       "POST",
-			Pattern:      "/peers/{peerid}/devices",
+			Pattern:      "/devices/{peerid}",
```
aravindavk: `restclient` patch merged recently. please update this change in `$SRC/pkg/restclient/device.go`

plugins/device/rest.go
```diff
@@ -106,3 +106,52 @@ func deviceAddHandler(w http.ResponseWriter, r *http.Request) {
...
+			restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		}
+		return
+	}
+
+	restutils.SendHTTPResponse(ctx, w, http.StatusOK, devices)
+
+}
+
+func listAllHandler(w http.ResponseWriter, r *http.Request) {
```
aravindavk: you can remove this function if planning to implement this later.

plugins/device/rest.go
```diff
@@ -106,3 +106,52 @@ func deviceAddHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	ctx := r.Context()
+	logger := gdctx.GetReqLogger(ctx)
+	peers, err := GetPeers()
+	if err != nil {
+		logger.WithError(err).Error("Failed to get peers from store")
+	}
+
+	devices, err := GetDevices(peerID)
```
aravindavk: loop is missing. peerID is undefined

PR #789, opened 2018-05-24
config: removed default configuration file

aravindavk:
To view the default values run `glusterd2 --help`


prashanthpai:
> Glusterd2 first looks for default config file in $install_prefix/etc/glusterd2/glusterd2.toml

I'm proposing a simpler solution. Let me know your thoughts on it:

We don't need a default config file. We can just use defaults from code. Correct me if I'm wrong - the actual need (problem that we're solving) is to set a prefix (/usr/local) during build time.

How about passing it during build time with LDFLAGS ?

See `scripts/build.sh` for example usage:

```sh
```

The `version.GitSHA` variable is set during build time. The same can be done for `$PREFIX`. We can derive defaults from that.

aravindavk:
> We don't need a default config file. We can just use defaults from code. Correct me if I'm wrong - the actual need (problem that we're solving) is to set a prefix (/usr/local) during build time.

Looks good, I will remove the default config file

> How about passing it during build time with LDFLAGS ?

looks like neat approach than generating go code with shell script. Let me explore the LDFLAGS and update the PR.

aravindavk:
@prashanthpai made the changes as suggested. `LDFLAGS` is very nice. Thanks for suggesting

glusterd2/config.go
```diff
@@ -18,39 +16,48 @@ import (
...
+	flag.String("clientaddress", defaultclientaddress, "Address to bind the REST service.")
+	flag.String("peeraddress", defaultpeeraddress, "Address to bind the inter glusterd2 RPC service.")
 
 	// TODO: SSL/TLS is currently only implemented for REST interface
 	flag.String("cert-file", "", "Certificate used for SSL/TLS connections from clients to glusterd2.")
 	flag.String("key-file", "", "Private key for the SSL/TLS certificate.")
 
 	// PID file
-	flag.String("pidfile", "", "PID file path(default: rundir/gluster/glusterd2.pid)")
+	flag.String("pidfile", "", "PID file path. (default \"rundir/glusterd2.pid)\"")
```
prashanthpai:
```go
flag.String("pidfile", fmt.Sprintf("%s/glusterd2.pid", defaultrundir), "PID file path")
```

glusterd2/config.go
```diff
@@ -18,39 +16,48 @@ import (
...
 	// metrics
-	expConfig = expvar.NewMap("config")
-)
+	expConfig           = expvar.NewMap("config")
 
-const (
-	defaultLogLevel = "debug"
-	defaultConfName = "glusterd2"
+	// Following variables value set by LDFLAGS
+	defaultworkdir = ""
```
prashanthpai: Is it possible to have just `PREFIX` passed to LDFLAGS and have these derived from that ? That would simplify Makefile too.

For example, in practical use, logdir can default to only `/var/log/glusterfs` or `/usr/local/var/log/glusterfs` where `/usr/local` is the prefix. Similarly, `/var/lib/glusterd2` and `/usr/local/var/lib/glusterd2`

```go
defaultlogdir = prefix + "/var/log/glusterfs"
defaultrundir = prefix + "/var/run/gluster"
```

glusterd2/config.go
```diff
@@ -60,27 +67,8 @@ func parseFlags() {
 // setDefaults sets defaults values for config options not available as a flag,
 // and flags which don't have default values
 func setDefaults() error {
-	cwd, err := os.Getwd()
```
prashanthpai: This is still required to set `workdir` and `localstatedir` properly below.

glusterd2/config.go
aravindavk: thought about it, but that will be wrong when `--rundir` is specified

glusterd2/config.go
aravindavk: I am not sure about the order of execution when it is set through LDFLAGS. I may have to do those assignments in `init`. I will check and update the PR if works.

glusterd2/config.go
aravindavk: why? both workdir and localstatedir are set using LDFLAGS

aravindavk:
Updated PR, only one LDFLAG used and all other variables are derived.

glusterd2/config.go
prashanthpai: If workdir is not explicitly set by user, it (rightly) defaults to current working directory which is only available during run time via `Getwd()`

aravindavk: I am missing something. both `workdir` and `localstatedir` are defaulted to `$prefix/var/lib/glusterd2` now. What is the expected default path for workdir?

aravindavk: Ok, is it when user configures `localstatedir` but not `workdir`? if so is it ok to set `workdir=localstatedir` when not specified

prashanthpai: > What is the expected default path for workdir?

Current working directory.

PR #790, opened 2018-05-24
Filter peers on basis of metadata

vpandey-RH:
Updates [#482](https://github.com/gluster/glusterd2/issues/482)
Filter peers on basis of metadata either by metadata key/value/key+value


pkg/restclient/peer.go
```diff
@@ -8,11 +8,18 @@ import (
 )
 
 // PeerAdd adds a peer to the Cluster
-func (c *Client) PeerAdd(host string) (api.PeerAddResp, error) {
-
-	peerAddReq := api.PeerAddReq{
-		Addresses: []string{host},
-	}
+func (c *Client) PeerAdd(host string, metadata ...map[string]string) (api.PeerAddResp, error) {
```
prashanthpai: Have `PeerAdd` take `api.PeerAddReq` as argument.

glustercli/cmd/peer.go
```diff
@@ -34,6 +36,8 @@ func init() {
 
 	peerCmd.AddCommand(peerStatusCmd)
 
+        peerListCmd.Flags().StringVar(&flagCmdFilterKey, "key", "", "Filter by metadata key")
```
prashanthpai: There's something fishy about these indentations or the way github is showing it to me :-/ 

Are you sure all your go files pass gofmt test ?

rishubhjain: @prashanthpai  Even I was wondering whats wrong, but they seem alright in https://github.com/vpandey-RH/glusterd2/blob/c0dbae250d1b4d905798e09a82f43d0d378a9207/glustercli/cmd/peer.go

vpandey-RH: @prashanthpai  I am  pretty sure it passed the gofmt test. I will look into it though.

pkg/api/peer_req_resp.go
```diff
@@ -35,4 +35,14 @@ type PeerEditResp Peer
 type PeerGetResp Peer
 
 // PeerListResp is the response sent for a peer list request
+/*
+PeerListResp can also be filtered based on query parameters
+sent along with peer list api.
```
prashanthpai: `sent with peer list request`

jFYI: API is not "sent". It's an interface the client calls or interacts with.

The explanation can also be better. Something like:

```
The client can request to filter peer listing based on metadata key/value using query parameters

Example:
```

prashanthpai:
@vpandey-RH 

This PR introduced `gofmt` issues:
https://goreportcard.com/report/github.com/gluster/glusterd2

PR #802, opened 2018-05-28
List all devices

rishubhjain:


plugins/device/rest.go
```diff
@@ -132,3 +132,35 @@ func deviceListHandler(w http.ResponseWriter, r *http.Request) {
...
+
+func listAllDevicesHandler(w http.ResponseWriter, r *http.Request) {
+
+	ctx := r.Context()
+	logger := gdctx.GetReqLogger(ctx)
+	peerIDs, err := peer.GetPeerIDs()
+	if err != nil {
+		logger.WithError(err).Error("Failed to get peers from store")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+	}
```
Madhu-1: missing return?

plugins/device/api/resp.go
```diff
@@ -12,10 +12,14 @@ type Info struct {
 	AvailableSize uint64 `json:"available-size"`
 	ExtentSize    uint64 `json:"extent-size"`
 	Used          bool   `json:"used"`
+	PeerID        string `json:"peerid,omitempty"`
```
prashanthpai: No `omitempty` as this will always be present.

Also, `peer-id`

plugins/device/api/resp.go
```diff
@@ -12,10 +12,14 @@ type Info struct {
...
 }
 
 // AddDeviceResp is the success response sent to a AddDeviceReq request
 type AddDeviceResp api.Peer
 
 // ListDeviceResp is the success response sent to a ListDevice request
 type ListDeviceResp []Info
+
+// ListAllDevicesResp is success response sent to list all devices request
+type ListAllDevicesResp []Info
```
prashanthpai: Now that `PeerID` is a field, you can just use `ListDeviceResp` directly.

plugins/device/init.go
```diff
@@ -34,6 +34,13 @@ func (p *Plugin) RestRoutes() route.Routes {
 			Version:      1,
 			ResponseType: utils.GetTypeString((*deviceapi.ListDeviceResp)(nil)),
 			HandlerFunc:  deviceListHandler},
+		route.Route{
+			Name:         "ListAllDevices",
+			Method:       "GET",
+			Pattern:      "/devices",
+			Version:      1,
+			ResponseType: utils.GetTypeString((*deviceapi.ListAllDevicesResp)(nil)),
```
prashanthpai: Please regenerate API doc.

plugins/device/rest.go
```diff
@@ -129,3 +129,33 @@ func deviceListHandler(w http.ResponseWriter, r *http.Request) {
...
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+
+	var devices []deviceapi.Info
+	for _, peerID := range peerIDs {
+		deviceList, err := deviceutils.GetDevices(peerID.String())
+		if err != nil {
+			logger.WithError(err).WithField("peerid", peerID).Error("Failed to get devices for peer")
+			if err == errors.ErrPeerNotFound {
```
prashanthpai: If it's `errors.ErrPeerNotFound`, do not treat is an error to be returned to client. As you have not taken locks (which is right), it's possible that a peer remove happens between line `137` and line `147`. Just log and `continue` if the peer is not found.

Issue #788, opened 2018-05-23
Restrict size of peer and volume metadata


prashanthpai:
Impose a size (in bytes) based restriction on volume and peer metadata. This is the overall size and not a per request based size. Requests to store peer and volume metadata should be scrutinised.

prashanthpai:
Pseudo-code:

```go
const maxSize = 4096

currentSize := 0
for key, value := range volume.Metadata {
    currentSize += len(key) + len(value)
}

newSize := currentSize
for key, value := range req.Metadata {
    newSize += len(key) + len(value)
    if newSize > maxSize {
        // do something
    }
}
```

prashanthpai:
You can also do:

```go
func (*v Volume) MetadataLen() int {
    // loop and return size
}
```

PR #805, opened 2018-05-28
Add size constraint to the metadata field of a volume/peer.

vpandey-RH:
Closes [#788](https://github.com/gluster/glusterd2/issues/788)
Add size constraint to the metadata field of a volume/peer.
Size is checked while creating volume, adding peer, editing volume/peer metadata.
Current limit is 4KB.


glusterd2/commands/peers/addpeer.go
```diff
@@ -27,6 +31,11 @@ func addPeerHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if peer.GetMetadataLen(req.Metadata) > maxMetadataSizeLimit {
```
prashanthpai: You have to check this:

(current metadata size + metadata size from request) > maxMetadataSizeLimit

vpandey-RH:
retest this please

glusterd2/commands/peers/addpeer.go
vpandey-RH: As discussed, current metadata size + metadata size from request for addpeer won't make any sense since add peer current metadata size will be 0.

glusterd2/commands/volumes/volume-create.go
```diff
@@ -42,6 +46,10 @@ func validateVolCreateReq(req *api.VolCreateReq) error {
 			return gderrors.ErrEmptyBrickList
 		}
 	}
+	metadataSize := volume.GetMetadataLen(req.Metadata)
```
prashanthpai: Same comment here. The method should on the type and not in volume package. Name the method as `MetadataSize` and the call will look like:

```go
if req.MetadataSize() > maxMetadataSizeLimit {
```

glusterd2/commands/peers/addpeer.go
```diff
@@ -96,6 +105,7 @@ func addPeerHandler(w http.ResponseWriter, r *http.Request) {
 	if req.Zone != "" {
 		newpeer.Metadata["_zone"] = req.Zone
 	}
+
```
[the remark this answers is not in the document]
prashanthpai: Oh wait, you can do better. While calculation existing metadata, skip over those keys that start with underscore. They are internal metadata and users can't set them

glusterd2/commands/volumes/volume-edit.go
```diff
@@ -60,6 +62,14 @@ func volumeEditHandler(w http.ResponseWriter, r *http.Request) {
 		if req.DeleteMetadata {
 			delete(volinfo.Metadata, key)
 		} else {
+			offset := len(key) + len(value)
```
prashanthpai: The variable name `offset` is a little misleading. What do you think of below approach ?


Check `if req.MetadataSize()` is more than metadata size limit
If yes, return error.

If not, loop over and update `volinfo.Metadata`.
Now check if `volinfo.MetadataSize()` is more than metadata size limit.

glusterd2/peer/store-utils.go
```diff
@@ -210,3 +210,12 @@ func GetPeerIDByAddr(addr string) (uuid.UUID, error) {
 	}
 	return p.ID, nil
 }
+
```
prashanthpai: See earlier comment, this should be a method of the type and named `MetadataSize()`

glusterd2/volume/store-utils.go
```diff
@@ -196,3 +196,12 @@ func Exists(name string) bool {
 
 	return resp.Count == 1
 }
+
```
prashanthpai: See earlier comment, this should be a method of the type and named `MetadataSize()`

vpandey-RH:
retest this please

glusterd2/peer/peer.go
```diff
@@ -24,3 +26,14 @@ type ETCDConfig struct {
 	ClusterState   string
 	DeletePeer     bool
 }
+
+// GetMetadataSize returns the size of metadata from peer info
+func (p Peer) GetMetadataSize() int {
```
prashanthpai: Name this as `MetadataSize()`

Do the same for other versions of this function for peer, volume and request.

glusterd2/peer/peer.go
```diff
@@ -24,3 +26,14 @@ type ETCDConfig struct {
...
+	currentSize := 0
```
prashanthpai: Name this as just `size`

Do the same for other versions of this function for peer, volume and request.

pkg/api/peer_req_resp.go
```diff
@@ -44,3 +48,25 @@ Example:
 Note - Cannot use query parameters if peerid is also supplied.
 */
 type PeerListResp []PeerGetResp
+
+// GetMetadataSize returns the size of the peer metadata in PeerAddReq
+func (p PeerAddReq) GetMetadataSize() int {
```
prashanthpai: `PeerAddReq` should be a pointer handle - `*PeerAddReq`

Same for `PeerEditReq`

glusterd2/commands/volumes/volume-edit.go
```diff
@@ -51,6 +51,12 @@ func volumeEditHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	metadataSize := volinfo.GetMetadataSize()
```
prashanthpai: This doesn't handle replacing the keys case.

```go
if req.MetadataSize() > maxMetadataSizeLimit {
    // return error
}

// update volinfo.Metadata and replace keys, as done below

if volinfo.MetadataSize() > maxMetadataSizeLimit {
    return error
}

```

Use the same logic for peer edit

vpandey-RH: Ooh yes, you are right. So, sorry. Will update it.

glusterd2/commands/peers/editpeer.go
```diff
@@ -104,10 +104,20 @@ func txnPeerEdit(c transaction.TxnCtx) error {
 		return err
 	}
 
+	reqMetadataSize := req.MetadataSize()
+	if reqMetadataSize > maxMetadataSizeLimit {
```
prashanthpai: Use the function call directly in if condition, there's no need for extra variable.

glusterd2/commands/peers/editpeer.go
```diff
@@ -104,10 +104,20 @@ func txnPeerEdit(c transaction.TxnCtx) error {
...
+		return gderrors.ErrMetadataSizeOutOfBounds
+	}
+
 	for k, v := range req.Metadata {
 		peerInfo.Metadata[k] = v
 	}
 
+	metadataSize := peerInfo.MetadataSize()
```
prashanthpai: Use the function call directly in if condition, there's no need for extra variable.

glusterd2/commands/volumes/volume-create.go
```diff
@@ -42,6 +46,10 @@ func validateVolCreateReq(req *api.VolCreateReq) error {
 			return gderrors.ErrEmptyBrickList
 		}
 	}
+	metadataSize := req.MetadataSize()
+	if metadataSize > maxMetadataSizeLimit {
```
prashanthpai: Use the function call directly in if condition, there's no need for extra variable.

glusterd2/commands/volumes/volume-edit.go
```diff
@@ -63,6 +68,12 @@ func volumeEditHandler(w http.ResponseWriter, r *http.Request) {
 			volinfo.Metadata[key] = value
 		}
 	}
+
+	metadataSize := volinfo.MetadataSize()
```
prashanthpai: Use the function call directly in if condition, there's no need for extra variable.

glusterd2/volume/store-utils.go
```diff
@@ -196,3 +196,12 @@ func Exists(name string) bool {
...
+// GetMetadataLen return the length of volume metadata
+func GetMetadataLen(metadata map[string]string) int {
```
prashanthpai: Remove this file from commit as changes done to this file as it's not used.

pkg/api/peer_req_resp.go
```diff
@@ -44,3 +48,25 @@ Example:
 Note - Cannot use query parameters if peerid is also supplied.
 */
 type PeerListResp []PeerGetResp
+
```
prashanthpai: As both these are in the same package, you can take the common code out to a function:

```go
func mapSize(metadata map[string]string) int {
    size := 0
    for key, value := range p.Metadata {
        ...
    }
    return size
}

func (p *PeerAddReq) MetadataSize() int {
    return mapSize(p.Metadata)
}

func (p *PeerEditReq) MetadataSize() int {
    return mapSize(p.Metadata)
}
```

prashanthpai: Do the same for volume structs too as it's in the same `api` package. You can put `mapSize` in `pkg/api/utils.go`

PR #808, opened 2018-05-28
New transaction locking mechanisms

kshlm:
This PR introduces a new NewTxnWithLocks function and a (*Txn).Lock method.

With this change, support for having locking as part of transaction steps has
been removed. Doing locks within steps could lead to TOCTOU race problems, as
in several cases, volinfos would be fetched/modified in the rest handler before
the transaction steps were performed.

@prashanthpai started doing changes with regards to the TOCTOU problem with
them.

With this change, users now need to explicitly create a transaction and obtain
locks on the required IDs before fetching/modifying stored data in the rest
handlers.

The rest of the transaction framework remains as is.

### DIFF
--- a/glusterd2/transaction/transaction.go
+++ b/glusterd2/transaction/transaction.go
@@ -54,6 +54,7 @@ func NewTxn(ctx context.Context) *Txn {
 	}).WithPrefix(prefix)
 
 	t.Ctx.Logger().Debug("new transaction created")
+	expTxn.Add("initiated_txn_in_progress", 1)
 	return t
 }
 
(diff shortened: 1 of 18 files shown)


glusterd2/transaction/transaction.go
```diff
@@ -39,19 +42,43 @@ type Txn struct {
...
+	}
+
+	return t, nil
+}
+
+// Done releases any obtained locks and cleans up the transaction namespace
+// Done must be called after a transaction ends
+func (t *Txn) Done() {
+	// Release obtained locks
+	for _, locker := range t.locks {
```
prashanthpai: Does the ordering in list of locks matter ? I was wondering whether the last lock taken should be the first to get unlocked.

kshlm: Not really ATM, and I don't expect it to. The locks are taken for the lifetime of the period of the transaction, and are released all at once at the end. I don't expect it to change.

kshlm:
> CreateLockFuncs was a temporary solution to take locks without a transaction i.e there are places where we need locks but no other steps are needed. One such example is editing volume.Metadata or peer.Metadata. Does this PR handle those cases too ? IOW, can you confirm that the new lock functions can be called without populating steps and calling txn.Do() ?

Yes. You can create a new transaction without populating the steps or running the steps. And after that you can add as many locks as you want with txn.Lock(). You need to ensure that you call txn.Done() when you're done.

prashanthpai:
Closes https://github.com/gluster/glusterd2/issues/510

prashanthpai:
retest this please

kshlm:
Okay. This was fast. I was hoping to wait till a few of the other larger PRs got merged (rebalance, snapshot, IVC).

Issue #794, opened 2018-05-25
Collate store access when setting/getting from txn context


prashanthpai:
Operations such as volume create heavily depend on setting and getting values in etcd as a way to communicate information between transaction steps and also to share intermediate/node results.

This is a proposal to change `txn.Set` to `txn.Get` to only set/get it in a local map when called. When the transaction step finishes, we sync it to the store.

Implementation in pseudo-code:

`glusterd2/transaction/rpc-service.go`:

```go
// RunStep handles the incoming request. It executes the requested step and returns the results
func (p *txnSvc) RunStep(rpcCtx context.Context, req *TxnStepReq) (*TxnStepResp, error) {

        var ctx Tctx
        json.Unmarshal(req.Context, &ctx)

        f, ok := getStepFunc(req.StepFunc)

        // sync key/values from store and populate ctx.localMap

        f(&ctx)

        // sync key/values in ctx.localMap back to store (if modified or there are new values)
}
```

This will ensure that there is exactly one etcd `Get` and at most one etcd `Put` operation per transaction step. This is abstracted away from actual calls to `txn.Get` and `txn.Set` inside the step functions.

Related issue #620

prashanthpai:
@kshlm @aravindavk I'd appreciate your thoughts/inputs on this.

PR #809, opened 2018-05-28
 txn: Cache keys/values of transaction context

prashanthpai:
This should result in at most one etcd GET and one etcd PUT operation
per transaction step.

Closes https://github.com/gluster/glusterd2/issues/794


prashanthpai:
@kshlm I'll wait for https://github.com/gluster/glusterd2/pull/808 to get in first.

glusterd2/transaction/context.go
```diff
@@ -110,23 +84,30 @@ func (c *Tctx) SetNodeResult(nodeID uuid.UUID, key string, value interface{}) er
...
-	if r.Count == 0 {
-		c.log.WithError(err).WithField("key", storeKey).Debug("key not found in store")
+	// return cached key
+	storeKey := c.config.StorePrefix + key
+	if data, ok := c.readSet[storeKey]; ok {
+		if err := json.Unmarshal(data, value); err != nil {
+			c.logger.WithError(err).WithField("key", storeKey).Error("failed to unmarshall value")
+		}
+	} else {
 		return errors.New("key not found")
```
kshlm: If you don't find a key in the cache you need to get it from the store. You cannot just return an error. The context might have had newer key-values set between the first Get and later Gets.

glusterd2/transaction/rpc-service.go
```diff
@@ -19,35 +19,44 @@ func init() {
...
+	var (
+		resp   TxnStepResp
+		f      StepFunc
+		err    error
+		ok     bool
+		logger log.FieldLogger
+	)
+
+	var ctx Tctx
+	if err = json.Unmarshal(req.Context, &ctx); err != nil {
```
kshlm: Instead of embedding the context marshalled to JSON, we can now just define the context config as a protobuf message struct and add it to the TxnStepReq. The only reason I went with embedding json was because I didn't find a way to implement custom protobuf marshalling for the old-style contexts with parent contexts.

glusterd2/transaction/context.go
prashanthpai: > The context might have had newer key-values set between the first Get and later Gets.

I thought about that and it's very unlikely to be the case in real usage:

```go

txn.Set("somekey", something)

// do something 

txn.Get("somekey", &something)
```

The step functions run as a single isolated unit. The Get and Set in the previous code violated this. In the above snippet, there is no case for the `Get` to consume the key set by any other concurrently running step function. In effect, it'll only be expecting the value it set earlier. There's no reason do that as the value doesn't and shouldn't change.

glusterd2/transaction/rpc-service.go
prashanthpai: The field logger only has txnid as the only field. Hence, would it suffice is only txn id is sent on wire ?

glusterd2/transaction/context.go
prashanthpai: Refreshed the PR to aggregate Set() calls and defer committing to store. This should fix the problem you described.

kshlm: I did look at this a little later. You do always clear the cache each time a Set it done (though just locally), so the first Get later has to refresh the cache. My earlier comment becomes invalid.

But I don't agree with clearing out the full cache on a Set. That would throw away the gains to be had by caching. I'd suggest that you delete the cached key/value, and on a Get of the key, when you don't find it in the cache, get from store and cache it.

glusterd2/transaction/rpc-service.go
kshlm: > The field logger only has txnid as the only field. Hence, would it suffice if only txn id is sent on wire ?

This sounds okay with me. The only reason the full TxnCtx was being on-wire was because the first implementation didn't use the store to save txn data. So I had send across the full context. Now that we use the store, it's enough that we just pass along the identifier. The remote end can get the context from the store.

glusterd2/transaction/context.go
```diff
@@ -139,69 +162,36 @@ func (c *Tctx) GetNodeResult(nodeID uuid.UUID, key string, value interface{}) er
...
 func (c *Tctx) Delete(key string) error {
-	storeKey := c.prefix + "/" + key
-	_, e := store.Store.Delete(context.TODO(), storeKey)
-	if e != nil {
-		c.log.WithFields(log.Fields{
-			"error": e,
-			"key":   storeKey,
-		}).Error("failed to delete key")
-		return e
+	storeKey := c.config.StorePrefix + key
```
kshlm: What about the cached values here? The next get would still get the value from the cache.

glusterd2/transaction/rpc-service.go
kshlm: > The type logrus.Fields is of type map[string]interface{} and there's no simple way to define that in proto syntax. Proto does support map but the value must be a concrete type. A simpler alternative is to use gob or json which supports map[string]interface{}.

Oh well. That was another reason why I decided to embed JSON in protobuf.

glusterd2/transaction/context.go
prashanthpai: I had overlooked this. I was supposed to remove Delete () altogther, after confirming that it wasn't being used anywhere.

kshlm: Just for the sake of completeness, I'd like to have delete as well.

prashanthpai: Done.

prashanthpai:
retest this please

prashanthpai:
retest this please

prashanthpai:
retest this please

Issue #815, opened 2018-05-29
failed to enable bitrot


Madhu-1:
**step to reproduce**

1. createvolume 
2. start volume
3. enable bitrot on volume
```
./build/glustercli  bitrot enable test_volume
Failed to enable bitrot for volume test_volume


Error: Request failed with HTTP Status code 500
Step bitrot-enable.Commit failed on 2 nodes

```

4.try to enable bitrot again

```
./build/glustercli --glusterd-host=10.70.43.235 bitrot enable test_volume
Failed to enable bitrot for volume test_volume


Error: Request failed with HTTP Status code 400
Bitrot is already enabled

```
**Expected**: Error response should be same in above both conditions

5. try to get vol info, volume info shows bitrot option is enabled 
**bitrot option shouldn't be enabled** 
```
 [
    {
        "id": "335b6682-4b1f-429a-96c9-429777dbe10d",
        "name": "test_volume2",
        "type": "Distribute",
        "transport": "tcp",
        "distribute-count": 4,
        "replica-count": 0,
        "arbiter-count": 0,
        "options": {
            "bit-rot.scrubber": "true",
            "bitrot-stub.bitrot": "on"
        },
        "state": "Started",
        "subvols": [
            {
                "name": "test_volume2-distribute-0",
                "type": "Distribute",
                "bricks": [
                    {
                        "id": "a3b931a5-bd58-4e9f-8247-d8d6c3d3f927",
                        "path": "/export/brick1/data",
                        "volume-id": "335b6682-4b1f-429a-96c9-429777dbe10d",
                        "volume-name": "test_volume2",
                        "peer-id": "240f4082-e951-4f52-85ff-46210efd4b78",
                        "host": "dhcp42-235.lab.eng.blr.redhat.com",
                        "type": "Brick"
                    }
                ],
                "replica-count": 1,
                "arbiter-count": 0
            },
            {
                "name": "test_volume2-distribute-1",
                "type": "Distribute",
                "bricks": [
                    {
                        "id": "872a55de-bf5b-4bc7-98ca-7306483f5e0a",
                        "path": "/export/brick2/data",
                        "volume-id": "335b6682-4b1f-429a-96c9-429777dbe10d",
                        "volume-name": "test_volume2",
                        "peer-id": "9f62c6db-944e-4e5c-a5c0-658e4fea369b",
                        "host": "dhcp43-75.lab.eng.blr.redhat.com",
                        "type": "Brick"
                    }
                ],
[...]

PR #827, opened 2018-05-30
volgen: Fix bitrot volgen issue when no local bricks in a subvolume

aravindavk:
Updates: #815

### DIFF
--- a/glusterd2/volgen2/clustergraph.go
+++ b/glusterd2/volgen2/clustergraph.go
@@ -46,6 +46,13 @@ func clusterGraph(volfile *Volfile, dht *Entry, vol *volume.Volinfo, nodeid uuid
 			}
 		}
 
+		// If Graph is required only for Local Bricks then
+		// do not include sub volume in graph if no local bricks
+		// exists for the sub volume
+		if filters != nil && filters.onlyLocalBricks && len(subvol.GetLocalBricks()) == 0 {
+			continue
+		}
+
 		// If Not set in prev filter checks
 		if parent == nil {
 			switch subvol.Type {
--- a/glusterd2/volume/struct.go
+++ b/glusterd2/volume/struct.go
@@ -273,3 +273,16 @@ func (v *Volinfo) MetadataSize() int {
 	}
 	return size
 }
+
+// GetLocalBricks returns a list of local Bricks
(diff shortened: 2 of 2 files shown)


glusterd2/volgen2/clustergraph.go
```diff
@@ -46,6 +46,23 @@ func clusterGraph(volfile *Volfile, dht *Entry, vol *volume.Volinfo, nodeid uuid
 			}
 		}
 
+		// If Graph is required only for Local Bricks then
```
Madhu-1: to be cleaner can we use continue with LABEL like below 

```
LABEL:
if filters != nil && filters.onlyLocalBricks {
			
	for _, b := range subvol.Bricks {
			if uuid.Equal(b.PeerID, nodeid) {
				continue LABEL
			}
         }
}
```

aravindavk: nice feature, but in this case not much useful. Your example will go wrong and loop will be skipped if one local brick found

Madhu-1: @aravindavk agreed

prashanthpai:
retest this please

glusterd2/volgen2/clustergraph.go
```diff
@@ -46,6 +46,23 @@ func clusterGraph(volfile *Volfile, dht *Entry, vol *volume.Volinfo, nodeid uuid
...
+		// do not include sub volume in graph if no local bricks
+		// exists for the sub volume
+		if filters != nil && filters.onlyLocalBricks {
+			localBricksExists := false
+			for _, b := range subvol.Bricks {
```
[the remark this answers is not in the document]
aravindavk: > Why not add a method on Subvol to get local bricks. Like you have it for Volinfo. That would simplify this to,

Nice, I will add the method.

> I saw a similar comment by you on another PR, so just wondering why you didn't think of it. 

that comment is for the existing method :) similar loop code is used in this file so didn't think about it. Thanks, I will add the method to subvol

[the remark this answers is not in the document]
aravindavk: but performance wise loop is better(over optimization) here, exits early once it finds atleast one local brick. :)

PR #828, opened 2018-05-30
Creating utility functions

rishubhjain:


rishubhjain:
@prashanthpai @aravindavk Please review

plugins/device/deviceutils/store-utils.go
```diff
@@ -42,8 +41,29 @@ func GetDevices(peerIds ...string) ([]deviceapi.Info, error) {
...
+func CheckIfDeviceExist(peerInfo *peer.Peer) bool {
+	_, exists := peerInfo.Metadata["_devices"]
+	if !exists {
+		return false
+	}
+	return true
+}
+
+// DeviceInDeviceList checks whether the given device is in list of devices or not.
+func DeviceInDeviceList(reqDevice string, devices []deviceapi.Info) bool {
```
prashanthpai: You can simply name this as `DeviceInList`

plugins/device/deviceutils/store-utils.go
```diff
@@ -42,8 +41,29 @@ func GetDevices(peerIds ...string) ([]deviceapi.Info, error) {
...
+	if len(peerInfo.Metadata["_devices"]) > 0 {
+		if err := json.Unmarshal([]byte(peerInfo.Metadata["_devices"]), &deviceInfo); err != nil {
+			return nil, err
+		}
+	}
+	return deviceInfo, nil
+}
+
+// CheckIfDeviceExist checks whether any device is registered with the peer or not.
+func CheckIfDeviceExist(peerInfo *peer.Peer) bool {
```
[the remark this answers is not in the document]
rishubhjain: @prashanthpai This function is required since there is a possibility that the path where device is stored changes, so its better if we keep this check in-store utils instead of directly accessing path in the code.

rishubhjain:
@prashanthpai I have created the util functions and made changes in the code. These functions are required so that the direct use of Devices is stopped, and in future we can put necessary checks in just one place.

plugins/device/deviceutils/store-utils.go
```diff
@@ -42,8 +41,20 @@ func GetDevices(peerIds ...string) ([]deviceapi.Info, error) {
 	return devices, nil
 }
 
-// DeviceExist checks the given device existence
-func DeviceExist(reqDevice string, devices []deviceapi.Info) bool {
+// FetchDevices returns devices from peer object.
+func FetchDevices(peerInfo *peer.Peer) ([]deviceapi.Info, error) {
```
prashanthpai: This function name is misleading. This could very well mean fetching it from store which is not true. How about naming this `GetDevicesFromPeer` or just `DevicesFromPeer`

plugins/device/deviceutils/store-utils.go
```diff
@@ -42,8 +41,20 @@ func GetDevices(peerIds ...string) ([]deviceapi.Info, error) {
...
+
+	var deviceInfo []deviceapi.Info
+	if len(peerInfo.Metadata["_devices"]) > 0 {
```
prashanthpai: You should be checking for existence of the key and not length

plugins/device/deviceutils/store-utils.go
```diff
@@ -28,12 +28,11 @@ func GetDevices(peerIds ...string) ([]deviceapi.Info, error) {
 			return nil, err
 		}
 	}
-
 	var devices []deviceapi.Info
 	for _, peerInfo := range peers {
-		var deviceInfo []deviceapi.Info
-		if len(peerInfo.Metadata["_devices"]) > 0 {
-			if err := json.Unmarshal([]byte(peerInfo.Metadata["_devices"]), &deviceInfo); err != nil {
+		if _, exists := peerInfo.Metadata["_devices"]; exists {
```
prashanthpai: This check will be done within the function and isn't necessary here. The consumer only has to check if the devices list is empty and not other about how it's stored.

plugins/device/rest.go
```diff
@@ -55,19 +54,19 @@ func deviceAddHandler(w http.ResponseWriter, r *http.Request) {
...
-		if deviceutils.DeviceExist(req.Device, devices) {
-			logger.WithError(err).WithField("device", req.Device).Error("Device already exists")
-			restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "Device already exists")
-			return
-		}
+	_, exists := peerInfo.Metadata["_devices"]
+	devices, err := deviceutils.FetchDevices(peerInfo)
+
+	if devices == nil && exists && err != nil {
+		logger.WithError(err).WithField("peerid", peerID).Error(err)
```
prashanthpai: This log entry doesn't have a message. The `err` should be passed to `WithError` and is not a substitute for message.

PR #842, opened 2018-06-01
allow only peerID as input for peer removal in glustercli

### LINKED ISSUE #837: cli: Peer remove should only support peerid as input
Currently, peer remove can be given both peer ID and hostname. We desire most of the APIs to use IDs wherever it's present to reduce ambiguity. Using IDs will also prevent user from inadvertently giving wrong IP. it also simplifies peer removal when multiple glusterd2s are running on the same node.

Madhu-1:
Fixed #837 


Madhu-1:
@aravindavk @prashanthpai, please review

PR #533, opened 2018-01-23
Snapshot: patches for basic functionality

rafikc30:

PR #846, opened 2018-06-01
snapshot: Implement clone, status, delete, restore 

rafikc30:
This PR implements the snapshot clone, status, delete, restore. Along with that, this PR contains  review comments captured here[1]. It also contains changes related to refactoring, pending TODO.

[1] : https://github.com/gluster/glusterd2/pull/533

### DIFF
--- a/e2e/lvmtest/lvmtest.go
+++ b/e2e/lvmtest/lvmtest.go
@@ -13,15 +13,13 @@ import (
 	"github.com/gluster/glusterd2/glusterd2/volume"
 )
 
-const (
-	lvmPrefix    string = "patchy_snap"
-	xfsFormat    string = "/usr/sbin/mkfs.xfs"
-	fallocateBin string = "/usr/bin/fallocate"
-	mknodBin     string = "/usr/bin/mknod"
-)
+const lvmPrefix string = "patchy_snap"
 
 var (
-	brickPrefix string
+	xfsFormat    = lvm.GetBinPath("mkfs.xfs")
+	fallocateBin = lvm.GetBinPath("fallocate")
+	mknodBin     = lvm.GetBinPath("mknod")
+	brickPrefix  string
 )
 
 func verifyLVM() bool {
--- a/e2e/lvmtest/lvmtest.go
+++ b/e2e/lvmtest/lvmtest.go
@@ -2,20 +2,28 @@ package lvmtest
 
 import (
 	"errors"
+	"fmt"
 	"os"
 	"os/exec"
 	"strconv"
 	"strings"
+
+	"github.com/gluster/glusterd2/glusterd2/snapshot/lvm"
+)
+
+const (
+	lvmPrefix    string = "patchy_snap"
+	xfsFormat    string = "/usr/sbin/mkfs.xfs"
+	fallocateBin string = "/usr/bin/fallocate"
+	mknodBin     string = "/usr/bin/mknod"
 )
 
 var (
-	brickPrefix = "/d/backends/"
-	lvmPrefix   = "patchy_snap"
-	pathPrefix  []string
+	brickPrefix string
 )
 
 func verifyLVM() bool {
-	out, err := exec.Command("/sbin/lvcreate", "--help").Output()
+	out, err := exec.Command(lvm.CreateCommand, "--help").Output()
 	if err != nil {
 		return false
 	}
@@ -28,9 +36,8 @@ func createBrickPath(num int) ([]string, error) {
 	var brickPath []string
 
 	for i := 1; i <= num; i++ {
-		prefix := brickPrefix + strconv.Itoa(i) + "/" + lvmPrefix
-		pathPrefix = append(pathPrefix, prefix)
-		path := prefix + "_mnt"
+		prefix := fmt.Sprintf("%s%d/%s", brickPrefix, i, lvmPrefix)
+		path := fmt.Sprintf("%s_mnt", prefix)
 		err := os.MkdirAll(path, os.ModeDir|os.ModePerm)
 		if err != nil {
 			return brickPath, err
@@ -44,29 +51,31 @@ func createBrickPath(num int) ([]string, error) {
 func createLV(num int, thinpoolSize, virtualSize string) error {
 
 	for i := 1; i <= num; i++ {
-		brickPath := pathPrefix[i-1] + "_mnt"
-		devicePath := pathPrefix[i-1] + "_loop"
-		vg := lvmPrefix + "_vg_" + strconv.Itoa(i)
-		poolPath := "/dev/" + vg + "/thinpool"
-		xfsPath := "/dev/" + vg + "/brick_lvm"
+		prefix := fmt.Sprintf("%s%d/%s", brickPrefix, i, lvmPrefix)
+		brickPath := fmt.Sprintf("%s_mnt", prefix)
+		devicePath := fmt.Sprintf("%s_loop", prefix)
 
-		if _, err := exec.Command("/sbin/pvcreate", devicePath).Output(); err != nil {
+		vg := fmt.Sprintf("%s_vg_%d", lvmPrefix, i)
+		poolPath := fmt.Sprintf("/dev/%s/thinpool", vg)
+		xfsPath := fmt.Sprintf("/dev/%s/brick_lvm", vg)
+
+		if _, err := exec.Command(lvm.PvCreateCommand, devicePath).Output(); err != nil {
 			return err
 		}
 
-		if _, err := exec.Command("/sbin/vgcreate", vg, devicePath).Output(); err != nil {
+		if _, err := exec.Command(lvm.VgCreateCommand, vg, devicePath).Output(); err != nil {
 			return err
 		}
 
-		if _, err := exec.Command("/sbin/lvcreate", "-L", thinpoolSize, "-T", poolPath).Output(); err != nil {
+		if _, err := exec.Command(lvm.CreateCommand, "-L", thinpoolSize, "-T", poolPath).Output(); err != nil {
 			return err
 		}
 
-		if _, err := exec.Command("/sbin/lvcreate", "-V", virtualSize, "-T", poolPath, "-n", "brick_lvm").Output(); err != nil {
+		if _, err := exec.Command(lvm.CreateCommand, "-V", virtualSize, "-T", poolPath, "-n", "brick_lvm").Output(); err != nil {
 			return err
 		}
 
-		if _, err := exec.Command("/usr/sbin/mkfs.xfs", "-f", xfsPath).Output(); err != nil {
+		if _, err := exec.Command(xfsFormat, "-f", xfsPath).Output(); err != nil {
 			return err
 		}
 		if _, err := exec.Command("mount", "-t", "xfs", "-o", "nouuid", xfsPath, brickPath).Output(); err != nil {
@@ -79,15 +88,17 @@ func createLV(num int, thinpoolSize, virtualSize string) error {
 
 func deleteLV(num int) error {
 	for i := 1; i <= num; i++ {
-		brickPath := pathPrefix[i-1] + "_mnt"
-		vg := lvmPrefix + "_vg_" + strconv.Itoa(i)
+		prefix := fmt.Sprintf("%s%d/%s", brickPrefix, i, lvmPrefix)
+		brickPath := fmt.Sprintf("%s_mnt", prefix)
+		vg := fmt.Sprintf("%s_vg_%d", lvmPrefix, i)
+
 		if _, err := exec.Command("umount", "-f", brickPath).Output(); err != nil {
 			return err
 		}
 		if err := os.RemoveAll(brickPath); err != nil {
 			return err
 		}
-		if _, err := exec.Command("/sbin/vgremove", "-f", vg).Output(); err != nil {
+		if _, err := exec.Command(lvm.RemoveCommand, "-f", vg).Output(); err != nil {
 			return err
 		}
 
@@ -100,8 +111,9 @@ func deleteLV(num int) error {
 func deleteVHD(num int) error {
 
 	for i := 1; i <= num; i++ {
-		vhdPath := pathPrefix[i-1] + "_vhd"
-		devicePath := pathPrefix[i-1] + "_loop"
+		prefix := fmt.Sprintf("%s%d/%s", brickPrefix, i, lvmPrefix)
+		vhdPath := fmt.Sprintf("%s_vhd", prefix)
+		devicePath := fmt.Sprintf("%s_loop", prefix)
 		_, err := exec.Command("losetup", "-d", devicePath).Output()
 		if err != nil {
 			return err
@@ -121,14 +133,15 @@ func deleteVHD(num int) error {
 func createVHD(num int, size string) error {
 
 	for i := 1; i <= num; i++ {
-		vhdPath := pathPrefix[i-1] + "_vhd"
-		devicePath := pathPrefix[i-1] + "_loop"
+		prefix := fmt.Sprintf("%s%d/%s", brickPrefix, i, lvmPrefix)
+		vhdPath := fmt.Sprintf("%s_vhd", prefix)
+		devicePath := fmt.Sprintf("%s_loop", prefix)
 		//TODO replace exec command with syscall.Fallocate
-		_, err := exec.Command("/usr/bin/fallocate", "-l", size, vhdPath).Output()
+		_, err := exec.Command(fallocateBin, "-l", size, vhdPath).Output()
 		if err != nil {
 			return err
 		}
-		_, err = exec.Command("/usr/bin/mknod", "-m", "660", devicePath, "b", "7", strconv.Itoa(i+8)).Output()
+		_, err = exec.Command(mknodBin, "-m", "660", devicePath, "b", "7", strconv.Itoa(i+8)).Output()
 		loosetupCmd := exec.Command("losetup", devicePath, vhdPath)
 		_, err = loosetupCmd.Output()
 		if err != nil {
@@ -141,10 +154,11 @@ func createVHD(num int, size string) error {
 }
 
 //CreateLvmBricks provides an lvm mount point created using loop back devices
-func CreateLvmBricks(num int) ([]string, error) {
+func CreateLvmBricks(prefix string, num int) ([]string, error) {
 	var brickPath []string
+	brickPrefix = prefix
 	var err error
-	if lvm := verifyLVM(); lvm == false {
+	if !verifyLVM() {
 		return brickPath, errors.New("lvm or thinlv is not available on the machine")
 	}
 
@@ -168,10 +182,9 @@ func CreateLvmBricks(num int) ([]string, error) {
 //CleanupLvmBricks provides an lvm mount point created using loop back devices
 func CleanupLvmBricks(num int) error {
 	var err error
-	if lvm := verifyLVM(); lvm == false {
+	if !verifyLVM() {
 		return errors.New("lvm or thinlv is not available on the machine")
 	}
-
 	err = deleteVHD(num)
 	if err != nil {
 		return err
--- a/e2e/lvmtest/lvmtest.go
+++ b/e2e/lvmtest/lvmtest.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/gluster/glusterd2/glusterd2/snapshot/lvm"
 	"github.com/gluster/glusterd2/glusterd2/volume"
-	config "github.com/spf13/viper"
 )
 
 const (
@@ -183,20 +182,18 @@ func CreateLvmBricks(prefix string, num int) ([]string, error) {
 }
 
 //Cleanup will kill all process, and remove mount points
-func Cleanup(prefix string, brickCount int) {
+func Cleanup(baseWorkdir, prefix string, brickCount int) {
 
 	brickPrefix = prefix
 	exec.Command("pkill", "gluster").Output()
 	time.Sleep(3 * time.Second)
 
-	snapDirPrefix := config.GetString("rundir") + "/snaps/"
-	cloneDirPrefix := config.GetString("rundir") + "/clones/"
 	mtabEntries, err := volume.GetMounts()
 	if err != nil {
 		return
 	}
 	for _, m := range mtabEntries {
-		if strings.HasPrefix(m.MntDir, snapDirPrefix) || strings.HasPrefix(m.MntDir, cloneDirPrefix) {
+		if strings.HasPrefix(m.MntDir, baseWorkdir) {
 
 			//Remove any dangling mount pounts
 			exec.Command("umount", "-f", "-l", m.MntDir).Output()
--- a/e2e/lvmtest/lvmtest.go
+++ b/e2e/lvmtest/lvmtest.go
@@ -7,8 +7,10 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/gluster/glusterd2/glusterd2/snapshot/lvm"
+	config "github.com/spf13/viper"
 )
 
 const (
@@ -86,19 +88,19 @@ func createLV(num int, thinpoolSize, virtualSize string) error {
 	return nil
 }
 
-func deleteLV(num int) error {
+func deleteLV(num int, force bool) error {
 	for i := 1; i <= num; i++ {
 		prefix := fmt.Sprintf("%s%d/%s", brickPrefix, i, lvmPrefix)
 		brickPath := fmt.Sprintf("%s_mnt", prefix)
 		vg := fmt.Sprintf("%s_vg_%d", lvmPrefix, i)
 
-		if _, err := exec.Command("umount", "-f", brickPath).Output(); err != nil {
+		if _, err := exec.Command("umount", "-f", brickPath).Output(); err != nil && !force {
 			return err
 		}
-		if err := os.RemoveAll(brickPath); err != nil {
+		if err := os.RemoveAll(brickPath); err != nil && !force {
 			return err
 		}
-		if _, err := exec.Command(lvm.RemoveCommand, "-f", vg).Output(); err != nil {
+		if _, err := exec.Command(lvm.RemoveCommand, "-f", vg).Output(); err != nil && !force {
 			return err
 		}
 
@@ -108,20 +110,20 @@ func deleteLV(num int) error {
 }
 
 //create given number of virtual hard disk
-func deleteVHD(num int) error {
+func deleteVHD(num int, force bool) error {
 
 	for i := 1; i <= num; i++ {
 		prefix := fmt.Sprintf("%s%d/%s", brickPrefix, i, lvmPrefix)
 		vhdPath := fmt.Sprintf("%s_vhd", prefix)
 		devicePath := fmt.Sprintf("%s_loop", prefix)
 		_, err := exec.Command("losetup", "-d", devicePath).Output()
-		if err != nil {
+		if err != nil && !force {
 			return err
 		}
-		if err := os.RemoveAll(vhdPath); err != nil {
+		if err := os.RemoveAll(vhdPath); err != nil && !force {
 			return err
 		}
-		if err := os.RemoveAll(devicePath); err != nil {
+		if err := os.RemoveAll(devicePath); err != nil && !force {
 			return err
 		}
 
@@ -179,18 +181,35 @@ func CreateLvmBricks(prefix string, num int) ([]string, error) {
 
 }
 
+//Cleanup will kill all process, and remove mount points
+func Cleanup(prefix string, brickCount int) {
+
+	brickPrefix = prefix
+	exec.Command("pkill", "gluster").Output()
+	time.Sleep(2 * time.Second)
+
+	snapDirPrefix := config.GetString("rundir") + "/snaps/*"
+	//Remove any dangling snapshot mount pounts
+	exec.Command("umount", snapDirPrefix).Output()
+	deleteVHD(brickCount, true)
+	deleteLV(brickCount, true)
+	os.RemoveAll(brickPrefix)
+
+}
+
 //CleanupLvmBricks provides an lvm mount point created using loop back devices
-func CleanupLvmBricks(num int) error {
+func CleanupLvmBricks(prefix string, num int) error {
 	var err error
+	brickPrefix = prefix
 	if !verifyLVM() {
 		return errors.New("lvm or thinlv is not available on the machine")
 	}
-	err = deleteVHD(num)
+	err = deleteVHD(num, false)
 	if err != nil {
 		return err
 	}
 
-	err = deleteLV(num)
+	err = deleteLV(num, false)
 	if err != nil {
 		return err
 	}
--- a/e2e/lvmtest/lvmtest.go
+++ b/e2e/lvmtest/lvmtest.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/gluster/glusterd2/glusterd2/snapshot/lvm"
+	"github.com/gluster/glusterd2/glusterd2/volume"
 	config "github.com/spf13/viper"
 )
 
@@ -186,13 +187,32 @@ func Cleanup(prefix string, brickCount int) {
 
 	brickPrefix = prefix
 	exec.Command("pkill", "gluster").Output()
-	time.Sleep(2 * time.Second)
+	time.Sleep(3 * time.Second)
 
-	snapDirPrefix := config.GetString("rundir") + "/snaps/*"
-	//Remove any dangling snapshot mount pounts
-	exec.Command("umount", snapDirPrefix).Output()
+	snapDirPrefix := config.GetString("rundir") + "/snaps/"
+	cloneDirPrefix := config.GetString("rundir") + "/clones/"
+	mtabEntries, err := volume.GetMounts()
+	if err != nil {
+		return
+	}
+	for _, m := range mtabEntries {
+		if strings.HasPrefix(m.MntDir, snapDirPrefix) || strings.HasPrefix(m.MntDir, cloneDirPrefix) {
+
+			//Remove any dangling mount pounts
+			exec.Command("umount", "-f", "-l", m.MntDir).Output()
+		}
+	}
+
 	deleteVHD(brickCount, true)
 	deleteLV(brickCount, true)
+
+	vg := fmt.Sprintf("%s_vg_", lvmPrefix)
+	out, err := exec.Command(lvm.LVSCommand, "--noheadings", "-o", "vg_name").Output()
+	for _, entry := range strings.Split(string(out), "\n") {
+		if strings.HasPrefix(entry, vg) {
+			exec.Command(lvm.RemoveCommand, "-f", entry)
+		}
+	}
 	os.RemoveAll(brickPrefix)
 
 }
--- a/e2e/lvmtest/lvmtest.go
+++ b/e2e/lvmtest/lvmtest.go
@@ -1,0 +1,188 @@
+package lvmtest
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"strconv"
+	"strings"
+)
+
+var (
+	brickPrefix = "/d/backends/"
+	lvmPrefix   = "patchy_snap"
+	pathPrefix  []string
+)
+
+func verifyLVM() bool {
+	out, err := exec.Command("/sbin/lvcreate", "--help").Output()
+	if err != nil {
+		return false
+	}
+	thinLV := strings.Contains(string(out), "thin")
+	return thinLV
+}
+
+func createBrickPath(num int) ([]string, error) {
+
+	var brickPath []string
+
+	for i := 1; i <= num; i++ {
+		prefix := brickPrefix + strconv.Itoa(i) + "/" + lvmPrefix
+		pathPrefix = append(pathPrefix, prefix)
+		path := prefix + "_mnt"
+		err := os.MkdirAll(path, os.ModeDir|os.ModePerm)
+		if err != nil {
+			return brickPath, err
+		}
+		brickPath = append(brickPath, path)
+	}
+
+	return brickPath, nil
+}
+
+func createLV(num int, thinpoolSize, virtualSize string) error {
+
+	for i := 1; i <= num; i++ {
+		brickPath := pathPrefix[i-1] + "_mnt"
+		devicePath := pathPrefix[i-1] + "_loop"
+		vg := lvmPrefix + "_vg_" + strconv.Itoa(i)
+		poolPath := "/dev/" + vg + "/thinpool"
+		xfsPath := "/dev/" + vg + "/brick_lvm"
+
+		if _, err := exec.Command("/sbin/pvcreate", devicePath).Output(); err != nil {
+			return err
+		}
+
+		if _, err := exec.Command("/sbin/vgcreate", vg, devicePath).Output(); err != nil {
+			return err
+		}
+
+		if _, err := exec.Command("/sbin/lvcreate", "-L", thinpoolSize, "-T", poolPath).Output(); err != nil {
+			return err
+		}
+
+		if _, err := exec.Command("/sbin/lvcreate", "-V", virtualSize, "-T", poolPath, "-n", "brick_lvm").Output(); err != nil {
+			return err
+		}
+
+		if _, err := exec.Command("/usr/sbin/mkfs.xfs", "-f", xfsPath).Output(); err != nil {
+			return err
+		}
+		if _, err := exec.Command("mount", "-t", "xfs", "-o", "nouuid", xfsPath, brickPath).Output(); err != nil {
+			return err
+		}
+
+	}
+	return nil
+}
+
+func deleteLV(num int) error {
+	for i := 1; i <= num; i++ {
+		brickPath := pathPrefix[i-1] + "_mnt"
+		vg := lvmPrefix + "_vg_" + strconv.Itoa(i)
+		if _, err := exec.Command("umount", "-f", brickPath).Output(); err != nil {
+			return err
+		}
+		if err := os.RemoveAll(brickPath); err != nil {
+			return err
+		}
+		if _, err := exec.Command("/sbin/vgremove", "-f", vg).Output(); err != nil {
+			return err
+		}
+
+	}
+	return nil
+
+}
+
+//create given number of virtual hard disk
+func deleteVHD(num int) error {
+
+	for i := 1; i <= num; i++ {
+		vhdPath := pathPrefix[i-1] + "_vhd"
+		devicePath := pathPrefix[i-1] + "_loop"
+		_, err := exec.Command("losetup", "-d", devicePath).Output()
+		if err != nil {
+			return err
+		}
+		if err := os.RemoveAll(vhdPath); err != nil {
+			return err
+		}
+		if err := os.RemoveAll(devicePath); err != nil {
+			return err
+		}
+
+	}
+	return nil
+}
+
+//create given number of virtual hard disk
+func createVHD(num int, size string) error {
+
+	for i := 1; i <= num; i++ {
+		vhdPath := pathPrefix[i-1] + "_vhd"
+		devicePath := pathPrefix[i-1] + "_loop"
+		//TODO replace exec command with syscall.Fallocate
+		_, err := exec.Command("/usr/bin/fallocate", "-l", size, vhdPath).Output()
+		if err != nil {
+			return err
+		}
+		_, err = exec.Command("/usr/bin/mknod", "-m", "660", devicePath, "b", "7", strconv.Itoa(i+8)).Output()
+		loosetupCmd := exec.Command("losetup", devicePath, vhdPath)
+		_, err = loosetupCmd.Output()
+		if err != nil {
+			return err
+
+		}
+
+	}
+	return nil
+}
+
+//CreateLvmBricks provides an lvm mount point created using loop back devices
+func CreateLvmBricks(num int) ([]string, error) {
+	var brickPath []string
+	var err error
+	if lvm := verifyLVM(); lvm == false {
+		return brickPath, errors.New("lvm or thinlv is not available on the machine")
+	}
+
+	brickPath, err = createBrickPath(num)
+	if err != nil {
+		return brickPath, err
+	}
+	err = createVHD(num, "300M")
+	if err != nil {
+		return brickPath, err
+	}
+
+	err = createLV(num, "200M", "150M")
+	if err != nil {
+		return brickPath, err
+	}
+	return brickPath, nil
+
+}
+
+//CleanupLvmBricks provides an lvm mount point created using loop back devices
+func CleanupLvmBricks(num int) error {
+	var err error
+	if lvm := verifyLVM(); lvm == false {
+		return errors.New("lvm or thinlv is not available on the machine")
+	}
+
+	err = deleteVHD(num)
+	if err != nil {
+		return err
+	}
+
+	err = deleteLV(num)
+	if err != nil {
+		return err
+	}
+
+	err = os.RemoveAll(brickPrefix)
+	return err
+
+}
--- a/e2e/snapshot_ops_test.go
+++ b/e2e/snapshot_ops_test.go
@@ -3,7 +3,6 @@ package e2e
 import (
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"net"
 	"os"
 	"os/exec"
@@ -34,16 +33,16 @@ func TestSnapshot(t *testing.T) {
 	snapTestName = t.Name()
 	r := require.New(t)
 
-	prefix := fmt.Sprintf("%s/%s/bricks/", baseWorkdir, snapTestName)
-	lvmtest.Cleanup(baseWorkdir, prefix, brickCount)
+	prefix := fmt.Sprintf("%s/%s/bricks/", baseLocalStateDir, snapTestName)
+	lvmtest.Cleanup(baseLocalStateDir, prefix, brickCount)
 	defer func() {
-		lvmtest.Cleanup(baseWorkdir, prefix, brickCount)
+		lvmtest.Cleanup(baseLocalStateDir, prefix, brickCount)
 	}()
-	gds, err = setupCluster("./config/1.toml", "./config/2.toml")
+	tc, err := setupCluster("./config/1.toml", "./config/2.toml")
 	r.Nil(err)
-	defer teardownCluster(gds)
+	defer teardownCluster(tc)
 
-	client = initRestclient(gds[0].ClientAddress)
+	client = initRestclient(tc.gds[0])
 	brickPaths, err = lvmtest.CreateLvmBricks(prefix, brickCount)
 	r.Nil(err)
 	defer func() {
@@ -51,7 +50,7 @@ func TestSnapshot(t *testing.T) {
 	}()
 
 	// Create the volume
-	if err := volumeCreateOnLvm(snapTestName, brickPaths, client); err != nil {
+	if err := volumeCreateOnLvm(snapTestName, brickPaths, client, tc); err != nil {
 		r.Nil(err, "Failed to create the volume")
 	}
 
@@ -72,12 +71,12 @@ func TestSnapshot(t *testing.T) {
 	t.Run("Create", testSnapshotCreate)
 	t.Run("Activate", testSnapshotActivate)
 	t.Run("List", testSnapshotList)
-	t.Run("Mount", testSnapshotMount)
+	t.Run("Mount", tc.wrap(testSnapshotMount))
 	t.Run("StatusAndForceActivate", testSnapshotStatusForceActivate)
 	t.Run("Info", testSnapshotInfo)
 	t.Run("Clone", testSnapshotClone)
 	t.Run("Restore", testSnapshotRestore)
-	t.Run("MountRestoredVolume", testRestoredVolumeMount)
+	t.Run("MountRestoredVolume", tc.wrap(testRestoredVolumeMount))
 	t.Run("Deactivate", testSnapshotDeactivate)
 	t.Run("Delete", testSnapshotDelete)
 
@@ -88,7 +87,7 @@ func TestSnapshot(t *testing.T) {
 
 }
 
-func volumeCreateOnLvm(volName string, brickPaths []string, client *restclient.Client) error {
+func volumeCreateOnLvm(volName string, brickPaths []string, client *restclient.Client, tc *testCluster) error {
 
 	// create 2x2 dist-rep volume
 	createReq := api.VolCreateReq{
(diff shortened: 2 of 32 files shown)


e2e/snapshot_ops_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+	/*
+		TODO:
+		test snapshot on a volume that is expanded or shrinked
+	*/
+
+}
+
+func createLvmBricks(number int) ([]string, error) {
+	/*
```
Madhu-1: remove commented code

e2e/lvmtest/lvmtest.go
```diff
@@ -0,0 +1,188 @@
...
+		return brickPath, err
+	}
+	return brickPath, nil
+
+}
+
+//CleanupLvmBricks provides an lvm mount point created using loop back devices
+func CleanupLvmBricks(num int) error {
+	var err error
+	if lvm := verifyLVM(); lvm == false {
```
Madhu-1: this can be changed like 
```
if !verifyLVM(){
return brickPath, errors.New("lvm or thinlv is not available on the machine")
}
```

e2e/lvmtest/lvmtest.go
```diff
@@ -0,0 +1,188 @@
...
+	thinLV := strings.Contains(string(out), "thin")
+	return thinLV
+}
+
+func createBrickPath(num int) ([]string, error) {
+
+	var brickPath []string
+
+	for i := 1; i <= num; i++ {
+		prefix := brickPrefix + strconv.Itoa(i) + "/" + lvmPrefix
```
Madhu-1: don't use hardcoded path, use t.Name()

e2e/lvmtest/lvmtest.go
```diff
@@ -0,0 +1,188 @@
...
+func createLV(num int, thinpoolSize, virtualSize string) error {
+
+	for i := 1; i <= num; i++ {
+		brickPath := pathPrefix[i-1] + "_mnt"
+		devicePath := pathPrefix[i-1] + "_loop"
+		vg := lvmPrefix + "_vg_" + strconv.Itoa(i)
+		poolPath := "/dev/" + vg + "/thinpool"
+		xfsPath := "/dev/" + vg + "/brick_lvm"
+
+		if _, err := exec.Command("/sbin/pvcreate", devicePath).Output(); err != nil {
```
Madhu-1: all commands can be the constants

```
/sbin/pvcreate
/sbin/vgcreate
/sbin/vgcreate
/usr/sbin/mkfs.xfs
```
please take care of other commands also ,add those as constants

e2e/lvmtest/lvmtest.go
```diff
@@ -0,0 +1,188 @@
...
+
+	}
+	return nil
+}
+
+//CreateLvmBricks provides an lvm mount point created using loop back devices
+func CreateLvmBricks(num int) ([]string, error) {
+	var brickPath []string
+	var err error
+	if lvm := verifyLVM(); lvm == false {
```
Madhu-1: this can be changed as 

```
if !verifyLVM() {
		return brickPath, errors.New("lvm or thinlv is not available on the machine")
	}
```

e2e/snapshot_ops_test.go
```diff
@@ -0,0 +1,234 @@
...
+func TestSnapshot(t *testing.T) {
+	var err error
+	var brickPaths []string
+	r := require.New(t)
+
+	gds, err = setupCluster("./config/1.toml", "./config/2.toml")
+	if err != nil {
+		r.Nil(err)
+		teardownCluster(gds)
+		return
```
Madhu-1: don't return in the test cases, let it continue

e2e/snapshot_ops_test.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	defer func() {
+		err := lvmtest.CleanupLvmBricks(4)
+		r.Nil(err)
+	}()
+
+	// Create the volume
+	if err := volumeCreateOnLvm(brickPaths, client); err != nil {
+		t.Logf("Failed to create the volume")
```
Madhu-1: remove logging and return. if you want to log you can use  
```
r.Nil(err,"message to print if it fails")
```
take care  of this in rest of the code

e2e/snapshot_ops_test.go
```diff
@@ -0,0 +1,234 @@
...
+		r.Nil(err)
+		return
+	}
+
+	defer func() {
+		err := volumeStop(client)
+		r.Nil(err)
+	}()
+
+	t.Run("Create", testSnapshotCreate)
```
Madhu-1: use snapshot as bigging of the string 

``` CreateSnapshot```

File: e2e/snapshot_ops_test.go (hunk 6 above)
Comment (rafikc30) [reply e2e/snapshot_ops_test.go]: Why do we need to continue in cases where the configuration itself failed. If we continue we may encounter null pointer exceptions and proper cleanup won't happen. I didn't understand why we need to continue here.

File: e2e/snapshot_ops_test.go [hunk 9]
```diff
@@ -0,0 +1,234 @@
...
+				},
+			},
+		},
+		Force: true,
+	}
+	_, err := client.VolumeCreate(createReq)
+	return err
+}
+
+func volumeDelete(client *restclient.Client) error {
```
Madhu-1: no need to add an extra function, you  can call this inside your test cases

e2e/snapshot_ops_test.go
```diff
@@ -0,0 +1,234 @@
...
+		TimeStamp:   true,
+		Description: "Snapshot for testing, timestamp and force flags are enabled",
+		Force:       true,
+	}
+	_, err = client.SnapshotCreate(snapshotCreateReq)
+	r.Nil(err, "snapshot create failed")
+
+}
+
+func volumeStart(client *restclient.Client) error {
```
Madhu-1: no need to add an extra function, you can call this inside your test cases

e2e/snapshot_ops_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+	return client.VolumeStart(volname, true)
+}
+
+func volumeStop(client *restclient.Client) error {
```
Madhu-1: no need to add an extra function, you can call this inside your test cases

e2e/snapshot_ops_test.go
```diff
@@ -0,0 +1,234 @@
...
+	return client.VolumeStop(volname)
+}
+
+func testSnapshotList(t *testing.T) {
+	r := require.New(t)
+
+	snaps, err := client.SnapshotList("")
+	r.Nil(err)
+	r.Len(snaps[0].SnapName, 2)
+	snaps, err = client.SnapshotList(volname)
```
Madhu-1: add a new line between new tests
```
r.Len(snaps[0].SnapName, 2)

snaps, err = client.SnapshotList(volname)
```

e2e/snapshot_ops_test.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+func volumeDelete(client *restclient.Client) error {
+	return client.VolumeDelete(volname)
+}
+
+func testSnapshotCreate(t *testing.T) {
+	r := require.New(t)
+	snapshotCreateReq := api.SnapCreateReq{
+		VolName:  volname,
```
Madhu-1: use test case name t.Name() as volume name instead of volname

e2e/snapshot_ops_test.go
```diff
@@ -0,0 +1,234 @@
...
+		SnapName: snapname,
+	}
+	_, err := client.SnapshotCreate(snapshotCreateReq)
+	r.Nil(err, "snapshot create failed")
+
+	snapshotCreateReq = api.SnapCreateReq{
+		VolName:     volname,
+		SnapName:    snapname,
```
Madhu-1: use test case name t.Name() for volName and Snapname

e2e/snapshot_ops_test.go
Madhu-1: check other test cases. this is how the test framework as been created

e2e/snapshot_ops_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+	vols, err := client.SnapshotList(volname)
+	r.Nil(err)
+
+	for _, snaps := range vols {
+		for _, snapName := range snaps.SnapName {
+			err = client.SnapshotActivate(snapshotActivateReq, snapName)
+			r.Nil(err)
+
+			snapshotActivateReq.Force = true
```
Madhu-1: is this required here, or can we move it to beginning of the test case

e2e/snapshot_ops_test.go
```diff
@@ -0,0 +1,234 @@
...
+	var snapshotActivateReq api.SnapActivateReq
+	r := require.New(t)
+	vols, err := client.SnapshotList(volname)
+	r.Nil(err)
+	for _, snaps := range vols {
+		for _, snapName := range snaps.SnapName {
+			err = client.SnapshotDeactivate(snapName)
+			r.Nil(err)
+
+			snapshotActivateReq.Force = true
```
Madhu-1: snapshotActivateReq.Force = true is this required here, or can we move it to begin of the test case

glusterd2/commands/snapshot/snapshot-create.go
```diff
@@ -290,43 +292,38 @@ func populateBrickMountData(volinfo *volume.Volinfo, snapName string) (map[strin
...
-				MountDir:   mountDir,
-				DevicePath: devicePath,
-				FsType:     mntInfo.MntType,
-				MntOpts:    updateMntOps(mntInfo.MntType, mntInfo.MntOpts),
-			}
-			// Store the results in transaction context. This will be consumed by
-			// the node that initiated the transaction.
+			return nil, err
+		}
+		devicePath := "/dev/" + vG + "/" + snapName + "_" + strconv.Itoa(brickCount)
```
Madhu-1: use fmt.Sprintf()

glusterd2/commands/snapshot/snapshot-create.go
```diff
@@ -408,11 +405,26 @@ func takeBrickSnapshots(c transaction.TxnCtx) error {
...
+				"Path":         b.Path,
+			}).Debug("Running snapshot create command")
+
+			if err := lvm.LVSnapshot(mntInfo.FsName, mountData.DevicePath); err != nil {
 				c.Logger().WithError(err).WithField(
 					"brick", b.Path).Debug("Snapshot failed")
 				return err
 			}
+
+			err = lvm.UpdateFsLabel(mountData.DevicePath, mountData.FsType)
```
Madhu-1: missing returning error if  err!= nil

glusterd2/commands/snapshot/snapshot-delete.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	snapVol := snapinfo.SnapVolinfo
+	for _, b := range snapVol.GetLocalBricks() {
+		if snapVol.State == volume.VolStarted {
+			if err := snapshot.StopBrick(b); err != nil {
+				c.Logger().WithError(err).WithField(
+					"brick", b.Path).Warning("Failed to cleanup the brick.Earlier it might have stopped abnormaly")
+
+			}
```
Madhu-1: missing returning error?

glusterd2/snapshot/lvm/lvm.go
```diff
@@ -94,44 +83,29 @@ func GetVgName(mountDevice string) (string, error) {
...
 
-	cmd := exec.Command(CreateCommand, "-s", mntInfo.FsName, "--setactivationskip", "n", "--name", mountData.DevicePath)
-	err = cmd.Start()
+	cmd := exec.Command(CreateCommand, "-s", originDevice, "--setactivationskip", "n", "--name", DevicePath)
+	err := cmd.Start()
 	if err != nil {
 		return err
 	}
 
 	if true {
```
Madhu-1: not able to get this check
``` if true {```
if true is always true, why we need this?

File: e2e/snapshot_ops_test.go (hunk 6 above)
Comment (Madhu-1) [reply e2e/snapshot_ops_test.go]: @prashanthpai  can you answer @rafikc30  comment, I went through other e2e test cases also, we are not returning an error if setup cluster fails.

File: e2e/lvmtest/lvmtest.go (hunk 3 above)
Comment (rafikc30) [reply e2e/lvmtest/lvmtest.go]: I didn't understand this command, you mean we should avoid paths like brickPrefix = "/d/backends/" and use t.Name() instead ?

Comment (Madhu-1) [reply e2e/lvmtest/lvmtest.go]: brickPrefix should be replaced with t.Name() and use fmt.Sprintf()

File: e2e/snapshot_ops_test.go (hunk 14 above)
Comment (rafikc30) [reply e2e/snapshot_ops_test.go]: we could only use for one, both cannot exist with the same name

File: e2e/lvmtest/lvmtest.go (hunk 3 above)
Comment (rafikc30) [reply e2e/lvmtest/lvmtest.go]: This is the same path name used from the gd1 testing. I kept the same path in case if we are using  our .t fails replacing with glustercli in glusterd2, then it would be easier to port. I guess it is fine to use the brick path, as it just mount it on the location and does a RemoveAll at the end

Comment (Madhu-1) [reply e2e/lvmtest/lvmtest.go]: all glusterd2 test cases should use the /tmp/gd2_func_test directory for testing if we create any directory for testing it should be under this path only.

you can use below code to form brickPrefix
```
brickPrefix=baseWorkdir+t.Name()
```

File: e2e/snapshot_ops_test.go (hunk 1 above)
Comment (rafikc30) [reply e2e/snapshot_ops_test.go]: done

File: e2e/lvmtest/lvmtest.go (hunk 2 above)
Comment (rafikc30) [reply e2e/lvmtest/lvmtest.go]: done

File: e2e/lvmtest/lvmtest.go (hunk 3 above)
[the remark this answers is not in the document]
Comment (rafikc30) [reply e2e/lvmtest/lvmtest.go]: done

File: e2e/snapshot_ops_test.go (hunk 8 above)
Comment (Madhu-1) [reply e2e/snapshot_ops_test.go]: this will be redundant , you can discard this comment

File: e2e/snapshot_ops_test.go [hunk 21]
```diff
@@ -0,0 +1,239 @@
...
+	snapname     = "snaptest"
+	snapTestName string
+)
+
+// TestSnapshot creates a volume and snapshots, runs further tests on it and
+// finally deletes the volume
+func TestSnapshot(t *testing.T) {
+	var err error
+	var brickPaths []string
+	snapTestName = t.Name()
```
Madhu-1: you can use t.Name() directly, no need to assign it to a variable

rafikc30: That is not possible, t.Name() will change the name, when executing sub tests

e2e/snapshot_ops_test.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	_, err := client.VolumeCreate(createReq)
+	return err
+}
+
+func testSnapshotCreate(t *testing.T) {
+	r := require.New(t)
+	snapshotCreateReq := api.SnapCreateReq{
+		VolName:  snapTestName,
+		SnapName: snapname,
```
Madhu-1: as snapTestName is assigned with  TestSnapshot function name, for snapname you can use t.Name()

e2e/snapshot_ops_test.go
```diff
@@ -0,0 +1,239 @@
...
+	var snapshotActivateReq api.SnapActivateReq
+	var result api.SnapStatusResp
+	r := require.New(t)
+
+	vols, err := client.SnapshotList(snapTestName)
+	r.Nil(err)
+
+	snapName := vols[0].SnapName[0]
+	result, err = client.SnapshotStatus(snapName)
+	r.Nil(err)
```
Madhu-1: add a new line after ```r.Nil(err)```

e2e/snapshot_ops_test.go
rafikc30: still we need to have name, because t.Name() is not constant

glusterd2/commands/snapshot/snapshot-create.go
```diff
@@ -408,11 +404,33 @@ func takeBrickSnapshots(c transaction.TxnCtx) error {
 			*/
 			snapBrick := snapVol.Subvols[subvolCount].Bricks[count]
 			mountData := snapBrick.MountInfo
-			if err := lvm.BrickSnapshot(mountData, b.Path); err != nil {
+			length := len(b.Path) - len(mountData.Mountdir)
+			mountRoot := b.Path[:length]
+			mntInfo, err := volume.GetBrickMountInfo(mountRoot)
+			if err != nil {
+				return err
```
Madhu-1: logging is missing here

glusterd2/commands/snapshot/snapshot-delete.go
```diff
@@ -0,0 +1,141 @@
...
+
+	"github.com/gorilla/mux"
+	"github.com/pborman/uuid"
+)
+
+func snapshotDelete(c transaction.TxnCtx) error {
+
+	var snapinfo snapshot.Snapinfo
+	if err := c.Get("snapinfo", &snapinfo); err != nil {
+		return err
```
Madhu-1: add error logging here

glusterd2/commands/snapshot/snapshot-delete.go
```diff
@@ -0,0 +1,141 @@
...
+func snapshotDeleteHandler(w http.ResponseWriter, r *http.Request) {
+
+	ctx := r.Context()
+	logger := gdctx.GetReqLogger(ctx)
+	snapname := mux.Vars(r)["snapname"]
+	//Fetching snapinfo to get the parent volume name. Parent volume has to be locked
+	snapinfo, err := snapshot.GetSnapshot(snapname)
+	if err != nil {
+		if err == errors.ErrSnapNotFound {
+			restutils.SendHTTPError(ctx, w, http.StatusNotFound, errors.ErrVolNotFound)
```
Madhu-1: the error should be ```errors.ErrVolNotFound``` or snapshot not found?

glusterd2/commands/snapshot/snapshot-delete.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+		return
+	}
+	defer txn.Done()
+
+	//Fetching snapinfo again, but this time inside a lock
+	snapinfo, err = snapshot.GetSnapshot(snapname)
+	if err != nil {
+		if err == errors.ErrSnapNotFound {
+			restutils.SendHTTPError(ctx, w, http.StatusNotFound, errors.ErrVolNotFound)
```
Madhu-1: same here, the error should be ```errors.ErrVolNotFound``` or snapshot not found?

glusterd2/commands/snapshot/snapshot-delete.go
```diff
@@ -0,0 +1,141 @@
...
+	if err := txn.Do(); err != nil {
+		logger.WithError(err).WithField(
+			"snapname", snapname).Error("transaction to delete snapshot failed")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+
+	logger.WithField("Snapshot-name", snapname).Info("snapshot deleted")
+
+	restutils.SendHTTPResponse(ctx, w, http.StatusOK, "Snapshot deleted successfully")
```
Madhu-1: for delete, response should be like below
```
restutils.SendHTTPResponse(ctx, w, http.StatusNoContent, nil)
```

glusterd2/commands/snapshot/snapshot-status.go
```diff
@@ -0,0 +1,178 @@
...
+
+const (
+	brickStatusTxnKey string = "snapshotBrickstatuses"
+)
+
+func createSnapshotStatusResp(brickStatuses []brick.Brickstatus) []*api.SnapBrickStatus {
+	var statusesRsp []*api.SnapBrickStatus
+	var s api.SnapBrickStatus
+	for _, status := range brickStatuses {
+
```
Madhu-1: initialize ```var s api.SnapBrickStatus ``` inside for loop, it may contain old data for each for loop

glusterd2/commands/snapshot/snapshot-status.go
```diff
@@ -0,0 +1,178 @@
...
+		errMsg := "Failed to aggregate snapshot status results from multiple nodes."
+		logger.WithField("error", err.Error()).Error("BrickStatusHandler:" + errMsg)
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, errMsg)
+		return
+	}
+
+	restutils.SendHTTPResponse(ctx, w, http.StatusOK, resp)
+}
+
+func createSnapshotStatusesResp(ctx transaction.TxnCtx, snap *snapshot.Snapinfo) (*api.SnapStatusResp, error) {
```
Madhu-1: I don't think you need to return an error here. as you are always returning nil

glusterd2/daemon/daemon.go
```diff
@@ -132,6 +132,25 @@ func Start(d Daemon, wait bool) error {
...
+// When force == false, a SIGTERM signal is sent to the daemon.
+// When force == true, a SIGKILL signal is sent to the daemon.
+func Kill(pid int, force bool) error {
+
+	process, err := GetProcess(pid)
+	if err != nil {
+		return err
+	}
+
+	if force == true {
```
Madhu-1: this can be simplified as 

```
if force{
err = process.Kill()
}else{
err = process.Signal(syscall.SIGTERM)
}

```

glusterd2/snapshot/lvm/lvm.go
```diff
@@ -4,21 +4,36 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strconv"
 	"strings"
 
 	"github.com/gluster/glusterd2/glusterd2/brick"
 	"github.com/gluster/glusterd2/glusterd2/volume"
+	"github.com/gluster/glusterd2/pkg/api"
 	"github.com/pborman/uuid"
```
Madhu-1: add a new line here

glusterd2/snapshot/lvm/lvm.go
```diff
@@ -4,21 +4,36 @@ import (
...
-	"github.com/sirupsen/logrus"
 )
 
-var (
+const (
 	//CreateCommand is path to lvm create
```
Madhu-1: I think you need to change the comments here, all are referring to lvm create

glusterd2/snapshot/snapshot-utils.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/gluster/glusterd2/glusterd2/snapshot/lvm"
 	"github.com/gluster/glusterd2/glusterd2/volume"
 	"github.com/pborman/uuid"
```
Madhu-1: add a new line to adhere coding guidelines, this should be like below 
```
        "github.com/gluster/glusterd2/glusterd2/brick"
	"github.com/gluster/glusterd2/glusterd2/snapshot/lvm"
	
        "github.com/pborman/uuid"
        "golang.org/x/sys/unix"
```

glusterd2/snapshot/store-utils.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"strings"
```
Madhu-1: this import should be changed as below to adhere coding guidelines [guidelines](https://github.com/gluster/glusterd2/blob/master/doc/coding.md)
```
      "context"
      "encoding/json"
      "errors"
      "strings"

       gdstore "github.com/gluster/glusterd2/glusterd2/store"

       "github.com/coreos/etcd/clientv3"
```

glusterd2/snapshot/store-utils.go
```diff
@@ -112,7 +113,27 @@ func GetSnapshot(name string) (*Snapinfo, error) {
 
 //DeleteSnapshot passes the snap path to store to delete the snap object
 func DeleteSnapshot(snapInfo *Snapinfo) error {
+	var vol *volume.Volinfo
 	_, e := gdstore.Store.Delete(context.TODO(), GetStorePath(snapInfo))
```
Madhu-1: store wrapper has been introduced in PR #857 
the code should be changed as 
```
_, e := gdstore.Delete(context.TODO(), GetStorePath(snapInfo))
```

pkg/restclient/snapshot.go
```diff
@@ -42,3 +47,19 @@ func (c *Client) SnapshotInfo(snapname string) (api.SnapGetResp, error) {
 	err := c.get(url, nil, http.StatusOK, &snap)
 	return snap, err
 }
+
+// SnapshotDelete will delete Gluster Snapshot and respective lv
+func (c *Client) SnapshotDelete(snapname string) error {
+	url := fmt.Sprintf("/v1/snapshot/%s", snapname)
+	err := c.del(url, nil, http.StatusOK, nil)
```
Madhu-1: this should be changed as
```
```
we should expect HTTP status code 204 for a delete operation

e2e/snapshot_ops_test.go
rafikc30: done,

e2e/snapshot_ops_test.go
```diff
@@ -0,0 +1,239 @@
...
+	err = client.SnapshotActivate(snapshotActivateReq, snapName)
+	r.Nil(err)
+
+	retries := 4
+	waitTime := 6000
+	err = errors.New("Snapshot failed to activate forcefully")
+	for i := 0; i < retries; i++ {
+		// opposite of exponential backoff
+		time.Sleep(time.Duration(waitTime) * time.Millisecond)
+		result, err = client.SnapshotStatus(snapName)
```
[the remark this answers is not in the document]
rafikc30: done

glusterd2/snapshot/store-utils.go
rafikc30: done

glusterd2/commands/snapshot/snapshot-delete.go
```diff
@@ -0,0 +1,149 @@
...
+
+func snapshotDeleteHandler(w http.ResponseWriter, r *http.Request) {
+
+	ctx := r.Context()
+	logger := gdctx.GetReqLogger(ctx)
+	snapname := mux.Vars(r)["snapname"]
+	//Fetching snapinfo to get the parent volume name. Parent volume has to be locked
+	snapinfo, err := snapshot.GetSnapshot(snapname)
+	if err != nil {
+		if err == errors.ErrSnapNotFound {
```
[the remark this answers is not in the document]
rafikc30: done

glusterd2/commands/snapshot/snapshot-delete.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	go func() {
+		wg.Wait()
+		close(errCh)
+	}()
+	//var err error
+	err := error(nil)
+	for i := range errCh {
+		if i != nil && err == nil {
+			//Return the first error from goroutines
```
aravindavk: Not returned any error

glusterd2/commands/snapshot/snapshot-delete.go
```diff
@@ -0,0 +1,167 @@
...
+	log "github.com/sirupsen/logrus"
+
+	"github.com/gorilla/mux"
+	"github.com/pborman/uuid"
+)
+
+func snapshotBrickDelete(errCh chan error, wg *sync.WaitGroup, snapVol volume.Volinfo, b brick.Brickinfo, logger log.FieldLogger) {
+	defer wg.Done()
+
+	if snapVol.State == volume.VolStarted {
```
aravindavk: Snapshot delete allowed if Snapshot Volume is running?

glusterd2/commands/snapshot/snapshot-restore.go
```diff
@@ -0,0 +1,375 @@
...
+	"github.com/gluster/glusterd2/pkg/errors"
+
+	"github.com/gorilla/mux"
+	"github.com/pborman/uuid"
+	"golang.org/x/sys/unix"
+)
+
+const volumeIDXattrKey = "trusted.glusterfs.volume-id"
+
+func snapRestore(c transaction.TxnCtx) error {
```
aravindavk: Any solution to keep the same brick path as original brick? Unmount original brick and mount Snapbrick to original path?

glusterd2/commands/snapshot/snapshot-restore.go
```diff
@@ -0,0 +1,375 @@
...
+
+	err = txn.Do()
+	if err != nil {
+		logger.WithError(err).Error("snapshot restore transaction failed")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+
+	msg := fmt.Sprintf("Snapshot %s restored to volume %s", snapvolinfo.Name, vol.Name)
+	txn.Ctx.Logger().WithField("Snap Name", snapname).Info(msg)
```
aravindavk: Avoid space in key name. Change this to "snap-name"

e2e/snapshot_ops_test.go
```diff
@@ -0,0 +1,348 @@
...
+	mntCmd := exec.Command("mount", "-t", "glusterfs", volID, mntPath)
+	umntCmd := exec.Command("umount", mntPath)
+
+	err = mntCmd.Run()
+	r.Nil(err, fmt.Sprintf("mount failed: %s", err))
+
+	newDir := mntPath + "/Dir"
+	err = syscall.Mkdir(newDir, 0755)
+	if err == nil {
+		r.Nil(errors.New("Snapshot volume is Read Only File System"))
```
aravindavk: Start error string with lower case https://github.com/gluster/glusterd2/blob/master/doc/coding.md

glusterd2/commands/snapshot/snapshot-restore.go
rafikc30: yes.

rafikc30: Please ignore the previos command.
When we restore, it is not sure the whther the old directory is exist or not. Let's say we had done a remove brick after taking snapshot, then we cannot assure about the existance. In this case it is also possible that the directory might have used for some other mount or as a non-gluster mount.

glusterd2/commands/snapshot/snapshot-delete.go
rafikc30: done

glusterd2/commands/snapshot/snapshot-create.go
```diff
@@ -390,50 +390,73 @@ func validateSnapCreate(c transaction.TxnCtx) error {
...
 
-			log.WithFields(log.Fields{
-				"mount device": mntInfo.FsName,
-				"devicePath":   mountData.DevicePath,
-				"Path":         b.Path,
-			}).Debug("Running snapshot create command")
+		}
+	}
+	go func() {
+		wg.Wait()
```
kshlm: Waiting on goroutines by launching another goroutine does not seem like a good idea. You should wait in a goroutine that you have control over.

I suggest that you collect the errors in a different goroutine (the loop below) and wait for the snapshot goroutines and the err collect goroutine in this function.

glusterd2/commands/snapshot/snapshot-create.go
```diff
@@ -390,50 +390,73 @@ func validateSnapCreate(c transaction.TxnCtx) error {
...
 
-			if err = lvm.UpdateFsLabel(mountData.DevicePath, mountData.FsType); err != nil {
-				log.WithFields(log.Fields{
-					"FsType": mountData.FsType,
-					"Path":   b.Path,
-				}).Error("Failed to update the label")
-				return err
-			}
+	log.WithFields(log.Fields{
+		"mount device": mntInfo.FsName,
```
kshlm: Log field keys shouldn't contain spaces.

glusterd2/commands/snapshot/snapshot-create.go
```diff
@@ -390,50 +390,73 @@ func validateSnapCreate(c transaction.TxnCtx) error {
...
+		"devicePath":   mountData.DevicePath,
+		"Path":         b.Path,
+	}).Debug("Running  snapshot create command")
```
kshlm: There is an extra space in this log message.

glusterd2/commands/snapshot/snapshot-delete.go
```diff
@@ -4,41 +4,69 @@ import (
...
-		if err := lvm.RemoveBrickSnapshot(b.MountInfo.DevicePath); err != nil {
-			c.Logger().WithError(err).WithField(
-				"brick", b.Path).Debug("Failed to remove snapshotted LVM")
-			return err
+	for _, b := range snapVol.GetLocalBricks() {
+		wg.Add(1)
+		go snapshotBrickDelete(errCh, &wg, snapVol, b, c.Logger())
+	}
+	go func() {
+		wg.Wait()
```
kshlm: Same comment as earlier.

glusterd2/snapshot/store-utils.go
```diff
@@ -78,13 +78,13 @@ func GetSnapshotVolumes() ([]*volume.Volinfo, error) {
 func AddOrUpdateSnap(snapInfo *Snapinfo) error {
 	json, e := json.Marshal(snapInfo)
 	if e != nil {
-		log.WithField("error", e).Error("Failed to marshal the volinfo object")
+		log.WithField("error", e).Error("Failed to marshal the snapinfo object")
```
kshlm: This should be log.WithError

e2e/lvmtest/lvmtest.go
```diff
@@ -7,8 +7,10 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/gluster/glusterd2/glusterd2/snapshot/lvm"
+	config "github.com/spf13/viper"
```
kshlm: Have an empty line before importing viper. The convention for imports is to group stdlib packages first, followed by project packages, and then external packages.

glusterd2/commands/snapshot/snapshot-activate.go
```diff
@@ -41,16 +41,11 @@ func validateSnapActivate(c transaction.TxnCtx) error {
 		}
 		fallthrough
 	case false:
-		brickStatuses, err := volume.CheckBricksStatus(vol)
+		brickinfos, err = snapshot.GetOfflineBricks(vol)
 		if err != nil {
+			log.WithError(err).Error("failed to get Brick status")
```
kshlm: Fix the error message to indicate correct failure ("failed to get offline bricks").

glusterd2/commands/snapshot/snapshot-create.go
```diff
@@ -591,11 +597,9 @@ func duplicateVolinfo(vol, v *volume.Volinfo) {
 	 */
 	return
 }
-func snapshotBrickCreate(snapVolinfo, volinfo *volume.Volinfo, brickinfo *brick.Brickinfo, brickCount int) string {
-	mountData := brickinfo.MountInfo
-	nodeID := strings.Replace(brickinfo.ID.String(), "-", "", -1)
-	SnapDirPrefix := config.GetString("rundir") + "/snaps/"
-	brickPath := fmt.Sprintf("%s%s/%s/%s/brick%d%s", SnapDirPrefix, volinfo.Name, nodeID, snapVolinfo.Name, brickCount, mountData.Mountdir)
+func snapshotBrickCreate(snapName, volName, nodeID, mountDir string, brickCount int) string {
```
kshlm: Needs an empty line above. Also rename to `snapshotBrickPath` or something else that describes this function better.

glusterd2/commands/snapshot/snapshot-restore.go
```diff
@@ -0,0 +1,375 @@
...
+	txn, err := transaction.NewTxnWithLocks(ctx, snapname, snapinfo.ParentVolume)
+	if err != nil {
+		status, err := restutils.ErrToStatusCode(err)
+		restutils.SendHTTPError(ctx, w, status, err)
+		return
+	}
+	defer txn.Done()
+
+	//Fetching snapinfo again, but this time inside a lock
+	snapinfo, err = snapshot.GetSnapshot(snapname)
```
kshlm: Snapinfo is a modifiable object right? Then wouldn't you need a lock on the snapname before fetching and using it? Maybe take a lock on "snap-"+snapname.

glusterd2/commands/snapshot/snapshot-restore.go
kshlm: You've used "snapshot" elsewhere as the key for snapname, so use the same here.

glusterd2/snapshot/lvm/lvm.go
```diff
@@ -10,19 +10,20 @@ import (
 	"github.com/gluster/glusterd2/glusterd2/brick"
 	"github.com/gluster/glusterd2/glusterd2/volume"
 	"github.com/gluster/glusterd2/pkg/api"
+
 	"github.com/pborman/uuid"
 )
 
 const (
-	//CreateCommand is path to lvm create
+	//CreateCommand is path to lvcreate
```
kshlm: For these binaries use the `exec.LookPath` instead of hardcoding their paths. It may not be the same everywhere. You may need to change these to be variables rather than constants for that to work.

e2e/lvmtest/lvmtest.go
```diff
@@ -0,0 +1,237 @@
...
+	"strings"
+	"time"
+
+	"github.com/gluster/glusterd2/glusterd2/snapshot/lvm"
+	"github.com/gluster/glusterd2/glusterd2/volume"
+)
+
+const (
+	lvmPrefix    string = "patchy_snap"
+	xfsFormat    string = "/usr/sbin/mkfs.xfs"
```
kshlm: Use `exec.LookupPath` for binary paths. Don't hardcode binary paths.

kshlm:
Bah, turns out reviewing commits one-by-one is perfect. 😠

glusterd2/snapshot/lvm/lvm.go
```diff
@@ -4,21 +4,37 @@ import (
...
 )
 
-var (
-	//CreateCommand is path to lvm create
-	CreateCommand = "/sbin/lvcreate"
-	//RemoveCommand is path to lvm create
-	RemoveCommand = "/sbin/lvremove"
+const (
+	//CreateCommand is path to lvcreate
+	CreateCommand string = "/sbin/lvcreate"
```
kshlm: Use `exec.LookupPath` to set these. You may need to change them to be variables.

glusterd2/commands/snapshot/snapshot-create.go
```diff
@@ -378,53 +381,106 @@ func validateSnapCreate(c transaction.TxnCtx) error {
...
+				continue
+			}
+			wg.Add(1)
+			snapBrick := newVol.Subvols[subvolCount].Bricks[count]
+			go brickSnapshot(errCh, &wg, snapBrick, b)
+
+		}
+	}
+	go func() {
+		wg.Wait()
```
kshlm: Waiting on goroutines by launching another goroutine does not seem like a good idea. You should wait in a goroutine that you have control over.

I suggest that you collect the errors in a different goroutine (the loop below) and wait for the snapshot goroutines and the err collect goroutine in this function.

glusterd2/commands/snapshot/snapshot-delete.go
rafikc30: This err will be returned from the main return, we cannot just add a return as there are goroutines running.

glusterd2/commands/snapshot/snapshot-create.go
rafikc30: done

e2e/lvmtest/lvmtest.go
```diff
@@ -186,13 +187,32 @@ func Cleanup(prefix string, brickCount int) {
 
 	brickPrefix = prefix
 	exec.Command("pkill", "gluster").Output()
-	time.Sleep(2 * time.Second)
+	time.Sleep(3 * time.Second)
+
+	snapDirPrefix := config.GetString("rundir") + "/snaps/"
```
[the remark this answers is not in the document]
rafikc30: Already changed in the later patch sets

e2e/lvmtest/lvmtest.go
rafikc30: Changed in later patchsets

glusterd2/commands/snapshot/snapshot-restore.go
rafikc30: done

glusterd2/snapshot/lvm/lvm.go
rafikc30: done

glustercli/cmd/snapshot-delete.go
```diff
@@ -0,0 +1,43 @@
...
+	snapshotDeleteHelpShort = "Delete a Gluster Snapshot"
+	snapshotDeleteHelpLong  = "Delete a Gluster snapshot of the given name."
+)
+
+var (
+	snapshotDeleteCmd = &cobra.Command{
+		Use:   "delete <snapname>",
+		Short: snapshotDeleteHelpShort,
+		Long:  snapshotDeleteHelpLong,
+		Args:  cobra.MinimumNArgs(1),
```
Madhu-1: suggestion:

instead of ```cobra.MinimumNArgs(1)``` can you use ```cobra.ExactArgs(1)```

glustercli/cmd/snapshot-clone.go
```diff
@@ -0,0 +1,52 @@
...
+	snapshotCloneHelpShort = "Clone a Gluster Snapshot"
+	snapshotCloneHelpLong  = "Clone a Gluster snapshot into a new volume. This new volume will be a writable copy of snapshot."
+)
+
+var (
+	snapshotCloneCmd = &cobra.Command{
+		Use:   "clone <clonename> <snapname>",
+		Short: snapshotCloneHelpShort,
+		Long:  snapshotCloneHelpLong,
+		Args:  cobra.MinimumNArgs(2),
```
Madhu-1: suggestion:

instead of ```cobra.MinimumNArgs(2)``` can you use ```cobra.ExactArgs(2)```

glusterd2/commands/snapshot/snapshot-restore.go
```diff
@@ -0,0 +1,375 @@
...
+			}
+
+		}
+	}
+
+	return nil
+}
+
+func createVolumeBrickFromSnap(bricks []brick.Brickinfo, vol *volume.Volinfo) []brick.Brickinfo {
+	var newBricks []brick.Brickinfo
```
Madhu-1: suggestion: append may consume more memory (if the length exceeds it increments the capacity by twice ), I would suggest making slice capacity as a length of brickstatus

newBricks =make ([]brick.Brickinfo, 0, len(bricks)

glusterd2/snapshot/lvm/lvm.go
```diff
@@ -86,53 +98,36 @@ func MountSnapshotDirectory(mountPath string, mountData brick.MountInfo) error {
...
-		}).Debug("Child exited")
-
-		if errStatus != nil {
-			// Child exited with error
-			return errStatus
-		}
-		err = UpdateFsLabel(mountData.DevicePath, mountData.FsType)
-
+	// Wait for the child to exit
+	errStatus := cmd.Wait()
```
Madhu-1: you can return ```errStatus```  as you are not logging

glusterd2/snapshot/store-utils.go
```diff
@@ -92,27 +94,47 @@ func AddOrUpdateSnap(snapInfo *Snapinfo) error {
...
+		return e
+	}
+
+	vol, e = volume.GetVolume(snapInfo.ParentVolume)
+	if e != nil {
+		return e
+	}
+
+	//TODO change this when label based snapshots are in.
+	e = errors.New("Snap is not found in the volinfo")
```
Madhu-1: error message should start with smaller case

Madhu-1:
@rafikc30 please update ```endpoints.md``` file

rafikc30:
Thanks for the reviews, let me know once you finish. I will update the
patch sets.


Rafi KC

glusterd2/commands/snapshot/snapshot-activate.go
rafikc30: done

glusterd2/commands/snapshot/snapshot-clone.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+
+	// TODO Delete volfile from etcd properly
+	volume.DeleteVolume(vol.Name)
+
+	return nil
+}
+
+func storeSnapshotClone(c transaction.TxnCtx) error {
+
```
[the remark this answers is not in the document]
rafikc30: done

glusterd2/commands/snapshot/snapshot-clone.go
```diff
@@ -0,0 +1,387 @@
...
+		mountRoot, err := volume.GetBrickMountRoot(b.Path)
+		if err != nil {
+			return nil, err
+		}
+		mountDir := b.Path[len(mountRoot):]
+		mntInfo, err := volume.GetBrickMountInfo(mountRoot)
+		if err != nil {
+			log.WithError(err).WithField(
+				"brick", b.Path,
+			).Error("Failed to mount information")
```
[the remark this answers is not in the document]
rafikc30: done

glusterd2/commands/snapshot/snapshot-clone.go
```diff
@@ -0,0 +1,387 @@
...
+		nodeID := strings.Replace(b.ID.String(), "-", "", -1)
+
+		nodeData[b.String()] = snapshot.BrickMountData{
+			MountDir:   mountDir,
+			DevicePath: devicePath,
+			FsType:     mntInfo.MntType,
+			MntOpts:    updateMntOps(mntInfo.MntType, mntInfo.MntOpts),
+			Path:       snapshotCloneBrickCreate(volinfo.Name, name, nodeID, mountDir, brickCount),
+		}
+		// Store the results in transaction context. This will be consumed by
```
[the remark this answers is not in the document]
rafikc30: done

glusterd2/commands/snapshot/snapshot-clone.go
```diff
@@ -0,0 +1,387 @@
...
+		// the node that initiated the transaction.
+
+	}
+	return nodeData, nil
+}
+
+func validateSnapClone(c transaction.TxnCtx) error {
+	var statusStr []string
```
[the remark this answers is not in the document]
rafikc30: done

glusterd2/commands/snapshot/snapshot-clone.go
```diff
@@ -0,0 +1,387 @@
...
+		).Error("Bricks are offline")
+
+		return errors.New("one or more brick is offline")
+	}
+
+	if nodeData, err = populateCloneBrickMountData(volinfo, clonename); err != nil {
+		return err
+	}
+	c.SetNodeResult(gdctx.MyUUID, snapshot.NodeDataTxnKey, &nodeData)
+	//Quorum check ?
```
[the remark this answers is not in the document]
rafikc30: done

glusterd2/commands/snapshot/snapshot-clone.go
```diff
@@ -0,0 +1,387 @@
...
+	if err != nil {
+		log.WithFields(log.Fields{
+			"snapshot":    snapname,
+			"volume name": clonename,
+		}).Error("Failed to create clone volinfo")
+
+		return err
+	}
+
+	err = c.Set("volinfo", newVol)
```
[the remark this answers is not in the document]
rafikc30: done

glusterd2/commands/snapshot/snapshot-clone.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+}
+
+func snapshotCloneHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	req := new(api.SnapCloneReq)
+	logger := gdctx.GetReqLogger(ctx)
+
+	snapname := mux.Vars(r)["snapname"]
+	if snapname == "" {
```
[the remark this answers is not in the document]
rafikc30: That goes same with volume, I have added the code for validity check. It was not there when I last checked on volume create code.Please note that cloned snapshot will be a regular volume

glusterd2/commands/snapshot/snapshot-clone.go
```diff
@@ -0,0 +1,387 @@
...
+		restutils.SendHTTPError(ctx, w, http.StatusUnprocessableEntity, errors.New("Snapshot name should not be empty"))
+		return
+	}
+
+	if err := restutils.UnmarshalRequest(r, &req); err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusUnprocessableEntity, err)
```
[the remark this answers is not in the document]
rafikc30: done

glusterd2/commands/snapshot/snapshot-clone.go
```diff
@@ -0,0 +1,387 @@
...
+
+	txn, err := transaction.NewTxnWithLocks(ctx, req.CloneName, snapname)
+	if err != nil {
+		status, err := restutils.ErrToStatusCode(err)
+		restutils.SendHTTPError(ctx, w, status, err)
+		return
+	}
+	defer txn.Done()
+
+	snapinfo, err := snapshot.GetSnapshot(snapname)
```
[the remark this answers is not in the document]
rafikc30: No Validation should happen within a lock. Let's say there is a parallel volume-create happens with the same name. Then there is a chance for a race in volume.Exists(req.CloneName)

glusterd2/commands/snapshot/snapshot-clone.go
```diff
@@ -0,0 +1,387 @@
...
+		return
+	}
+	err = txn.Ctx.Set("clonename", &req.CloneName)
+	if err != nil {
+		logger.WithError(err).Error("failed to set request in transaction context")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+
+	err = txn.Do()
```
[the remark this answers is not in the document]
rafikc30: done

glustercli/cmd/snapshot-clone.go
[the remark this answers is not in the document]
rafikc30: done

glustercli/cmd/snapshot-delete.go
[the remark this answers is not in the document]
rafikc30: done

glustercli/cmd/snapshot-status.go
```diff
@@ -0,0 +1,84 @@
...
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+const (
+	helpSnapshotStatusCmd = "Status of all bricks of a Snapshot"
+)
+
+func init() {
```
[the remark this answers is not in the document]
rafikc30: done

glusterd2/brick/glusterfsd.go
```diff
@@ -286,3 +286,22 @@ func CreateSizeInfo(fstat *syscall.Statfs_t) *SizeInfo {
 	}
 	return &s
 }
+
+//CreateBrickStatusRsp creates response related to brick process
+func CreateBrickStatusRsp(brickStatuses []Brickstatus) []*api.BrickStatus {
+	var brickStatusesRsp []*api.BrickStatus
```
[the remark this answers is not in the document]
rafikc30: This was the case in most implementation in glusterd2.  Give me more time. I need to see how it differ. I would suggest to leave it like this and will change with on pr in all other places

glusterd2/commands/snapshot/snapshot-restore.go
rafikc30: same as above command

glusterd2/snapshot/lvm/lvm.go
rafikc30: done

glusterd2/snapshot/store-utils.go
rafikc30: done

e2e/lvmtest/lvmtest.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+//CleanupLvmBricks provides an lvm mount point created using loop back devices
+func CleanupLvmBricks(prefix string, num int) error {
+	var err error
+	brickPrefix = prefix
+	if !verifyLVM() {
+		return errors.New("lvm or thinlv is not available on the machine")
+	}
+	err = deleteVHD(num, false)
```
kshlm: For cases where a function returns error and you're checking only the error, (like here)., following the below pattern will be good.

```go
if err := function(); err != nil {
  // Do something
}
```

glustercli/cmd/snapshot-list.go
```diff
@@ -21,21 +21,21 @@ func init() {
 
 func snapshotListHandler(cmd *cobra.Command) error {
```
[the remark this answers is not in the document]
Madhu-1: sorry, when I searched didn't get the function call.

pkg/errors/error.go
```diff
@@ -48,8 +48,10 @@ var (
 	ErrEmptySnapName           = errors.New("snapshot name is empty")
 	ErrSnapExists              = errors.New("snapshot already exists")
 	ErrSnapNotFound            = errors.New("snapshot not found")
+	ErrSnapNotActivated        = errors.New("Snapshot not activated")
```
Madhu-1: an error message should start with a smaller case

glusterd2/commands/snapshot/snapshot-create.go
```diff
@@ -378,53 +381,109 @@ func validateSnapCreate(c transaction.TxnCtx) error {
...
+
 		return err
 	}
 	c.SetNodeResult(gdctx.MyUUID, snapshot.NodeDataTxnKey, &nodeData)
-	//Quorum check ?
+	//TODO Quorum check has to be implemented once we implement highly available snapshot
 	return nil
 }
+func takeVolumeSnapshots(newVol, oldVol *volume.Volinfo) error {
+	var wg sync.WaitGroup
```
kshlm: Heketi has a nice utility `github.com/heketi/heketi/pkg/utils.StatusGroup` which is a wrapper aroud `sync.Waitgroup` that does exactly what this pattern does. We should maybe create a common utils repo with such utilities and have both heketi and gd2 use them.

kshlm:
@rafikc30 There is an issue with tests after rebasing online. Can you please rebase manually and update the PR?

atinmu:
I see some test is failing here. Also I am not a very big fan of addressing any major comments in a different follow up PR. If it does take time to address them in the same PR, that's worth for the correctness and state of the implementation.

glusterd2/commands/snapshot/snapshot-create.go
phlogistonjohn: Hah! I kind of dislike that thing. Not that I dislike the intent, but just the implementation seems wonky and error prone to me.
FWIW some of the heketi utilty code's already been split out into a package here: https://github.com/heketi/utils/blob/master/statusgroup.go However I am not sure what versions the actual heketi code uses as some of the code was coped without having the in tree version removed (IIRC).

Madhu-1:
@rafikc30 CI failure looks genuine, please fix it

```
06:29:12 # github.com/gluster/glusterd2/e2e
06:29:12 e2e/snapshot_ops_test.go:41:2: undefined: gds
06:29:12 e2e/snapshot_ops_test.go:43:24: undefined: gds
06:29:12 e2e/snapshot_ops_test.go:45:26: undefined: gds
06:29:12 e2e/snapshot_ops_test.go:100:15: undefined: gds
06:29:12 e2e/snapshot_ops_test.go:101:15: undefined: gds
06:29:12 e2e/snapshot_ops_test.go:108:15: undefined: gds
06:29:12 e2e/snapshot_ops_test.go:109:15: undefined: gds
06:29:12 e2e/snapshot_ops_test.go:311:34: undefined: gds
06:29:12 e2e/snapshot_ops_test.go:328:34: undefined: gds
06:29:12 FAIL	github.com/gluster/glusterd2/e2e [build failed]
06:29:12 make: *** [functest] Error 2
```

Issue #850, opened 2018-06-04
gd2 is not updating the arbiter-count in shd and fuse volfiles


itisravi:
`option arbiter-count 1` is missing in AFR's xlator option in the mount and selfheal daemon volfiles. Without that, things won't work properly in AFR for arbiter volume. 
Example:
```

Volume Name: testvol
Type: Replicate
Volume ID: 71edcce5-a755-452f-999f-e68ae8cd3680
State: Started
Transport-type: tcp
Options:
    replicate.self-heal-daemon: on
Number of Bricks: 3
Brick1: 192.168.122.207:/bricks/brick1
Brick2: 192.168.122.207:/bricks/brick2
Brick3: 192.168.122.207:/bricks/brick3 (arbiter)
```

```
root@gd22 ~]# ETCDCTL_API=3 etcdctl --endpoints=192.168.122.207:2379 get gluster-731499d5-b8de-4923-a48a-5cb46c4d4d70/volfiles/testvol|grep arbiter|wc
      0       0       0
[root@gd22 ~]# ETCDCTL_API=3 etcdctl --endpoints=192.168.122.207:2379 get gluster-731499d5-b8de-4923-a48a-5cb46c4d4d70/volfiles/gluster/glustershd|grep arbiter|wc
      0       0       0
```

itisravi:
@aravindavk @vpandey-RH @prashanthpai  @karthik-us

PR #854, opened 2018-06-04
Set Arbiter count option on Arbiter Volume creation

aravindavk:
Fixes: #850


itisravi:
@aravindavk I had a doubt: I noticed that the   `option thin-arbiter <IP>:</brickpath>` for thin-arbiter volumes works without a similar fix. Is it handled in a different part of the code?

PR #857, opened 2018-06-06
added wrapper functions for store operations

### LINKED ISSUE #856: glusterd2 is not responding if ETCD server is down
even if  ETCD service is down, glusterd2 should return an internal server error(500)

**steps to reproduce issue** 

1)create a volume
2)stop external ETCD server
3get volume list

**Expected result:**
Internal Server Error (500)

**Actual result:**
1)For curl request, there is no response from glusterd2
2)Response from rest client

```
Error: Post http://10.70.43.235:24007/v1/volumes: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

Madhu-1:
Fixes #856 

### DIFF
--- a/glusterd2/peer/store-utils.go
+++ b/glusterd2/peer/store-utils.go
@@ -53,7 +53,7 @@ func AddOrUpdatePeer(p *Peer) error {
 
 	idStr := p.ID.String()
 
-	if _, err := store.Store.Put(context.TODO(), peerPrefix+idStr, string(json)); err != nil {
+	if _, err := store.Put(context.TODO(), peerPrefix+idStr, string(json)); err != nil {
 		return err
 	}
 
@@ -62,7 +62,7 @@ func AddOrUpdatePeer(p *Peer) error {
 
 // GetPeer returns specified peer from the store
 func GetPeer(id string) (*Peer, error) {
-	resp, err := store.Store.Get(context.TODO(), peerPrefix+id)
+	resp, err := store.Get(context.TODO(), peerPrefix+id)
 	if err != nil {
 		return nil, err
 	}
@@ -114,7 +114,7 @@ func getFilterType(filterParams map[string]string) metadataFilter {
 
 // GetPeers returns all available peers in the store
 func GetPeers(filterParams ...map[string]string) ([]*Peer, error) {
-	resp, err := store.Store.Get(context.TODO(), peerPrefix, clientv3.WithPrefix())
+	resp, err := store.Get(context.TODO(), peerPrefix, clientv3.WithPrefix())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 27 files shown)


Madhu-1:
@prashanthpai @aravindavk @kshlm  please review

glusterd2/peer/store-utils.go
```diff
@@ -257,7 +257,10 @@ func GetPeerByAddrs(addrs []string) (*Peer, error) {
 func GetPeerIDByAddr(addr string) (uuid.UUID, error) {
 	p, e := GetPeerByAddrF(addr)
 	if e != nil {
-		return nil, errors.ErrPeerNotFound
+		if e == errors.ErrPeerNotFound {
```
prashanthpai: You can just return error as is here.

```go
if e != nil {
    return nil, e
}
```

glusterd2/store/store.go
```diff
@@ -18,7 +19,10 @@ import (
 )
 
 const (
-	sessionTTL = 30 // used for etcd mutexes and liveness key
+	sessionTTL    = 30              // used for etcd mutexes and liveness key
+	getTimeout    = 5 * time.Second //timeout for store Get operation
```
prashanthpai: The variable names are self-explanatory, you can remove the comments for timeouts.

glusterd2/store/store.go
```diff
@@ -177,3 +181,37 @@ func newNamespacedStore(oc *clientv3.Client, conf *Config) (*GDStore, error) {
 
 	return &GDStore{*conf, kv, lease, watcher, session, oc, nil, namespaceKey}, nil
 }
+
+//Get wrapper function to call store Get operation with timeout if context is context.TODO()
```
prashanthpai: The comment for these can be better:

`Get is a wrapper function that calls clientv3.KV.Get with a default timeout if an empty context is passed`

You can do the same for `Put` and `Delete`

Madhu-1: done

glusterd2/store/store.go
Madhu-1: done

glusterd2/peer/store-utils.go
Madhu-1: done

glusterd2/store/store.go
```diff
@@ -18,7 +19,10 @@ import (
 )
 
 const (
-	sessionTTL = 30 // used for etcd mutexes and liveness key
+	sessionTTL    = 30 // used for etcd mutexes and liveness key
+	getTimeout    = 5 * time.Second
```
prashanthpai: I'm sorry, I should've noticed this before. Define this as constant and move the `time.Second` to place where it's used.

```go
context.WithTimeout(context.Background(), getTimeout * time.Second)
```

prashanthpai:
retest this please

PR #866, opened 2018-06-07
added volume size  command to glustercli

Madhu-1:

### DIFF
--- a/glustercli/cmd/volume.go
+++ b/glustercli/cmd/volume.go
@@ -26,6 +26,7 @@ const (
 	helpVolumeInfoCmd   = "Get Gluster Volume Info"
 	helpVolumeListCmd   = "List all Gluster Volumes"
 	helpVolumeStatusCmd = "Get Gluster Volume Status"
+	helpVolumeSizeCmd   = "Get Gluster Volume Size Usage"
 	helpVolumeExpandCmd = "Expand a Gluster Volume"
 	helpVolumeEditCmd   = "Edit metadata (key-value pairs) of a volume. Glusterd2 will not interpret these key and value in any way"
 )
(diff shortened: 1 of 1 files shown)


Madhu-1:
@prashanthpai please review

aravindavk:
Can we accommodate these details in Volume Status output itself?

prashanthpai:
> Can we accommodate these details in Volume Status output itself?

The size API is an expensive operation as it involves (fuse) mounting a volume. Is that acceptable ?

aravindavk:
> The size API is an expensive operation as it involves (fuse) mounting a volume. Is that acceptable ?

I thought Volume status command is also calling `client.VolumeStatus`

Madhu-1:
@aravindavk  volume status it calls BricksStatus

prashanthpai:
> @aravindavk volume status in CLI calls BricksStatus api

It's been named so in CLI to preserve backward compatibility

glustercli/cmd/volume.go
```diff
@@ -26,6 +26,7 @@ const (
 	helpVolumeInfoCmd   = "Get Gluster Volume Info"
 	helpVolumeListCmd   = "List all Gluster Volumes"
 	helpVolumeStatusCmd = "Get Gluster Volume Status"
+	helpVolumeSizeCmd   = "Get Gluster Volume Size"
```
prashanthpai: `Get Gluster Volume Size Usage`

glustercli/cmd/volume.go
```diff
@@ -432,6 +434,28 @@ var volumeStatusCmd = &cobra.Command{
...
+		vol, err := client.VolumeStatus(volname)
+		if err != nil {
+			if verbose {
+				log.WithFields(log.Fields{
+					"error": err.Error(),
+				}).Error("error getting volume size")
+			}
+			failure("Error getting Volume size", err, 1)
+		}
+		fmt.Println("Volume :", volname)
```
prashanthpai: You can remove the extra space before `:` but retain the one after it.

Here and below.

glustercli/cmd/volume.go
Madhu-1: done

glustercli/cmd/volume.go
Madhu-1: done

prashanthpai:
retest this please

PR #868, opened 2018-06-08
added support for JWT token authentication in glustercli

Madhu-1:

### DIFF
--- a/glustercli/cmd/common.go
+++ b/glustercli/cmd/common.go
@@ -18,8 +18,8 @@ var (
 `
 )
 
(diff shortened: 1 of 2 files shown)


Madhu-1:
@prashanthpai  please review

aravindavk:
Not related to this patch. One more flag is required for CLI `--auth-file` with default value `gd2statedir/auth`. 

Idea is, `glustercli` should get access if run from one of the Cluster node. If CLI is running from external nodes then new user creation is required(External invocation uses `--user` and `--secret`).  Auth file access can be controlled to by having read access to that file only to system user group "gluster"

PR #872, opened 2018-06-11
Fix typo in glustercli/cmd/root.go

rishubhjain:
Typo


prashanthpai:
@rishubhjain `Update root.go` is  a very broad and vague commit message title. The commit message title must describe the fix/solution and not the place/files you had to modify to get there.

Issue #851, opened 2018-06-04
Unified Volume Create API and mount bricks on Glusterd2 start


aravindavk:
Based on the comments in PR #661, listed the fixes required for Smart Volume Create API

- Single REST API `/volumes` and auto provision bricks if size is specified.
- Volume name as optional parameter
- Mount bricks on Glusterd2 start or Volume Start 
- Save device/lv details in `brick.MountInfo`(This structure is added by Snapshot feature, useful here also. We can even skip custom `fstab` logic since this is sufficient for mounting bricks on Glusterd2 start)
- Add `omitempty` for optional API parameters

aravindavk:
I will open new issues for each of the following tasks and move out of `GCS-Sprint1`

> Auto distribute count: If Distribute count is not specified and disk space is not enough to create volume, try again with increased distribute count till all the bricks can be planned(till max distribute count)

> Arbiter brick size calculation: Introduce configurable option for average file size with good default value. Apply the formula while planning arbiter brick

> Global defaults should be configurable: Default replica count, Default Snapshot reserve factor, etc should be configurable via API/Config

> Add support for multiple filesystem types(Default to xfs)

atinmu:
Moving this to Sprint 2 as there's good amount of work required to complete all of the issues referenced.

aravindavk:
> Moving this to Sprint 2 as there's good amount of work required to complete all of the issues referenced.

Patch #973 posted for the tasks other than the new issues opened. It can be merged if somebody reviews it.

aravindavk:
@atinmu updated status in first comment.

aravindavk:
Updated title and description of issue

aravindavk:
Closing this issue since #973 is merged.

PR #876, opened 2018-06-11
Auto generate Volume Name if not specified in Request

aravindavk:
Updates: #851


Madhu-1:
@aravindavk is it possible to add a test case for this.

glustercli/cmd/volume-create.go
```diff
@@ -45,12 +46,13 @@ var (
 		Use:   "create <volname> [<brick> [<brick>]...|--size <size>]",
```
Madhu-1: need to change this use

aravindavk: updated. Thanks

aravindavk:
> LGTM, only concern is in case of volume name is empty we are logging it in CLI.

That should be fine I guess, it will reflect the user input.

aravindavk:
> @aravindavk not related to this. Just a question
we have Volume struct and VolCreateReq struct in req_resp.go and VolCreateReq struct in volume_req.go. are we using all these structures for volume create?

We are using all of the struct, but we need to unify these structs. Once we decide on API endpoint for smartvol and regular volume create, we can start unifying them

glustercli/cmd/volume-create.go
```diff
@@ -133,17 +133,16 @@ func volumeCreateCmdRun(cmd *cobra.Command, args []string) {
 		return
 	}
 
-	if len(args) < 2 {
+	if len(args) < 1 {
```
raghavendra-talur: what happens when name is specified but bricks are omitted?

aravindavk: Converted name as flag(`--name=<name>`). args will contain bricks list only.

aravindavk:
I realized that, I need to update documentation and args upgrade script. Do not merge.

I will also add test case as @Madhu-1 suggested

glustercli/cmd/volume-create.go
raghavendra-talur: thanks!

aravindavk:
Added tests and updated documentation. Please review

PR #893, opened 2018-06-18
volume options reset: set volume options to default values.

Madhu-1:


Madhu-1:

Madhu-1:
@aravindavk Thanks for bringing this, I will add it in upcoming PR.

glusterd2/commands/volumes/volume-option.go
```diff
@@ -66,24 +66,47 @@ type txnOpType uint8
...
 
 // This function can be reused when volume reset operation is implemented.
 // However, volume reset can be also be treated logically as volume set but
 // with the value set to default value.
-func xlatorAction(c transaction.TxnCtx, txnOp txnOpType) error {
-
-	var req api.VolOptionReq
-	if err := c.Get("req", &req); err != nil {
-		return err
+func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, txnType txnOpType) error {
```
prashanthpai: The volume set or reset should be a different type altogether. One is user initiated op, the other is internal to txn framework. See earlier comment.

glusterd2/commands/volumes/volume-reset.go
```diff
@@ -89,17 +90,32 @@ func volumeResetHandler(w http.ResponseWriter, r *http.Request) {
...
 		{
-			DoFunc: "vol-option.UpdateVolinfo",
-			Nodes:  []uuid.UUID{gdctx.MyUUID},
+			DoFunc:   "vol-option.XlatorActionDoReSet",
+			UndoFunc: "vol-option.XlatorActionUndoReSet",
+			Nodes:    volinfo.Nodes(),
+		},
+		{
+			DoFunc:   "vol-option.UpdateVolinfo",
+			UndoFunc: "vol-option.UpdateVolinfo.Undo",
```
prashanthpai: Thanks for fixing the undo part :+1:

glusterd2/commands/volumes/volume-reset.go
```diff
@@ -89,17 +90,32 @@ func volumeResetHandler(w http.ResponseWriter, r *http.Request) {
 		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
 		return
 	}
-
+	//save volume information for transaction failure scenario
+	if err := txn.Ctx.Set("oldvolinfo", oldvolinfo); err != nil {
+		logger.WithError(err).Error("failed to set oldvolinfo in transaction context")
```
prashanthpai: You can remove this log. Or..at least maintain uniformity with `txn.Ctx.Set` done on line 115 below.

glusterd2/commands/volumes/volume-option.go
```diff
@@ -139,6 +169,8 @@ func registerVolOptionStepFuncs() {
 		{"vol-option.Validate", optionSetValidate},
 		{"vol-option.XlatorActionDoSet", xlatorActionDoSet},
 		{"vol-option.XlatorActionUndoSet", xlatorActionUndoSet},
+		{"vol-option.XlatorActionDoReSet", xlatorActionDoReSet},
```
prashanthpai: Please move registering these to `glusterd2/commands/volumes/volume-reset.go` file.

glusterd2/commands/volumes/volume-option.go
```diff
@@ -175,6 +207,12 @@ func volumeOptionsHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	//save volume information for transaction failure scenario
+	if err := txn.Ctx.Set("oldvolinfo", volinfo); err != nil {
+		logger.WithError(err).Error("failed to set oldvolinfo in transaction context")
```
prashanthpai: You don't need the log here. It's logged at `Error` log level inside 

```go
func (c *Tctx) Set(key string, value interface{}) error {

        b, err := json.Marshal(value)
        if err != nil {
                c.logger.WithError(err).WithField("key", key).Error("failed to marshal value")
                return err
        }
```

glusterd2/commands/volumes/volume-reset.go
```diff
@@ -89,17 +90,32 @@ func volumeResetHandler(w http.ResponseWriter, r *http.Request) {
...
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
 	txn.Steps = []*transaction.Step{
 		{
-			DoFunc: "vol-option.UpdateVolinfo",
-			Nodes:  []uuid.UUID{gdctx.MyUUID},
+			DoFunc:   "vol-option.XlatorActionDoReSet",
+			UndoFunc: "vol-option.XlatorActionUndoReSet",
+			Nodes:    volinfo.Nodes(),
```
prashanthpai: This also needs the skip optimization:

```go
Skip:     !isActionStepRequired(&req)
```

You can move implementation of `isActionStepRequired(&req)` to `glusterd2/commands/volumes/common.go`

You will have to make  `isActionStepRequired` take a generic map. That will allow you to reuse it.

glusterd2/commands/volumes/volume-option.go
```diff
@@ -107,6 +130,13 @@ func xlatorAction(c transaction.TxnCtx, txnOp txnOpType) error {
 			} else {
 				fn = xltr.Actor.Undo
 			}
+			if txnType == txnReset {
```
prashanthpai: Can you do this above in the first section ?


```go
if txnType == txnSet {
} else {
    var req api.VolOptionResetReq
    ...
    // here
}
```

glusterd2/commands/volumes/volume-reset.go
Madhu-1: done

glusterd2/commands/volumes/volume-option.go
Madhu-1: done

glusterd2/commands/volumes/volume-reset.go
```diff
@@ -89,17 +103,36 @@ func volumeResetHandler(w http.ResponseWriter, r *http.Request) {
 		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
 		return
 	}
-
+	//save volume information for transaction failure scenario
+	if err := txn.Ctx.Set("oldvolinfo", oldvolinfo); err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+	opt := make(map[string]string)
```
prashanthpai: I"m sorry @Madhu-1. I didn't realise that - one request was of map type and the other one was a slice. hence this doesn't look as clean. Duplicating `isActionStepRequired` might be cleaner. I'll let this one go though.

Madhu-1: if necessary will do clean up in upcoming PR.

glusterd2/commands/volumes/volume-option.go
prashanthpai: We can :+1: Feel free to do so.

Usually the caller should (optionally) log and inner function shouldn't. That's a good general practice. But in this specific case, it's beneficial to do it once internally because:

* the pattern repeats several times and these double logs clutter the code very much
* developers may forget to log in the caller
* developers may not use the right structured log format to log
* what is logged inside the function is same as what is logged outside - including the log level

prashanthpai:
retest this please

PR #894, opened 2018-06-18
volume-option-reset: return 204 status code on successful resetting of volume option.

Madhu-1:
instead of returning the remaining volume options left on the volume
for resetting volume options (delete operation). 
return 204 status code with empty HTTP response.

### DIFF
--- a/glusterd2/commands/volumes/volume-reset.go
+++ b/glusterd2/commands/volumes/volume-reset.go
@@ -114,5 +114,5 @@ func volumeResetHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	restutils.SendHTTPResponse(ctx, w, http.StatusOK, volinfo.Options)
+	restutils.SendHTTPResponse(ctx, w, http.StatusNoContent, nil)
 }


Madhu-1:

glusterd2/commands/volumes/volume-reset.go
```diff
@@ -114,5 +114,5 @@ func volumeResetHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	restutils.SendHTTPResponse(ctx, w, http.StatusOK, volinfo.Options)
+	restutils.SendHTTPResponse(ctx, w, http.StatusNoContent, nil)
```
rishubhjain: Shouldn't we send a success code and return the updated volume option status?

Madhu-1: I don't see any use in returning the remaining volume options. if we have only one option we will be returning nil to the user. will wait for others thought on this.

Madhu-1:
retest this please

Madhu-1:
retest this please

glusterd2/commands/volumes/volume-reset.go
prashanthpai: 204 is sent for as response to `DELETE` operation. We can stick to that convention everywhere DELETE is used.

In this case, we can return the entire updated `Volinfo` with `http.StatusOK`. We can't return 204 response because 204 responses shouldn't have a body

This response body can be used by clients (including e2e tests) to confirm that a volume reset actually removed it from the volume options without having to send another request for confirmation.

Madhu-1:
@prashanthpai Thanks, closing this PR

prashanthpai:
> @prashanthpai Thanks, closing this PR

You can actually make it return `Volinfo` instead of `Volinfo.Options` :)

Issue #890, opened 2018-06-15
SHD is not healing the files


karthik-us:
Even though the shd is started and connected to all the bricks, it is not healing the files which are pending heal.

karthik-us:
@aravindavk, @vpandey-RH can any one of you take a look at this?

itisravi:
@karthik-us you might want to explain the imshd part.

karthik-us:
In the gd1 shd graph for every subvol in a replicated/distributed-replicated volume the option
"option iam-self-heal-daemon yes"
will be set, which is missing in the gd2 shd graph. That might be causing the issue.

aravindavk:
I will add that missing option, it should be added under `type cluster/replicate` right?

itisravi:
It applies for both EC and AFR.

aravindavk:
@itisravi thanks. Is it possible to paste the sample glustershd volfile from glusterd1?

itisravi:
http://termbin.com/htqq has the sample output.  If not already there, all the below volume options need to be enabled when shd volfile is created:

```
volume testvol-replicate-0
    type cluster/replicate
    option iam-self-heal-daemon yes
    option use-compound-fops off
    option self-heal-daemon enable
    option entry-self-heal on
    option data-self-heal on
    option metadata-self-heal on
    option background-self-heal-count 0
    option afr-pending-xattr testvol-client-0,testvol-client-1,testvol-client-2
    subvolumes testvol-client-0 testvol-client-1 testvol-client-2
end-volume
```

aravindavk:
One more issue I need to address. Reopened the issue for the same

PR #865, opened 2018-06-07
Index/full heal API

vpandey-RH:
Fixes [#972](https://github.com/gluster/glusterd2/issues/972)
Updates [#427](https://github.com/gluster/glusterd2/issues/427)
Add Index and Full Heal API and cli support.

PR #896, opened 2018-06-18
volgen: Added new option for glustershd

aravindavk:
Following options are added to `cluster/replicate` and `cluster/disperse`
in self heal daemon volfile

    option iam-self-heal-daemon yes
    option afr-pending-xattr <xattr names>

Fixes: #890


aravindavk:
@karthik-us, @itisravi please review

itisravi:
@aravindavk Why are the other options that I listed in #890 not present?
Even with this change applied , I see that healing is not happening . You can try the below steps to verify the fix:

```
glustercli volume create testvol 192.168.122.207:/bricks/brick1 192.168.122.207:/bricks/brick2 --replica 2
glustercli volume start testvol
mount -t glusterfs 127.0.0.2:testvol /mnt/fuse_mnt
cd /mnt/fuse_mnt/
echo hello>>file
kill one brick
echo append>>file
glustercli volume stop testvol 
glustercli volume start testvol
glustercli volume set testvol replicate.heal-timeout 10 --advanced
```
After 10 seconds, heal should happen and file content must be same on both bricks

aravindavk:
> @aravindavk Why are the other options that I listed in #890 not present?

Other options can be set using `volume set` command.(Should be set during Volume create or using volume set command). I added those options which can't be set using volume set.

glusterd2/volgen2/clustergraph.go
```diff
@@ -73,6 +75,9 @@ func clusterGraph(volfile *Volfile, dht *Entry, vol *volume.Volinfo, nodeid uuid
 
 		if parent != nil {
 			for brickIdx, b := range subvol.Bricks {
+				afrPendingXattr = append(afrPendingXattr, fmt.Sprintf("%s-client-%d", vol.Name, clientIdx))
+				clientIdx++
```
itisravi: Will clientIdx++ always yield the same value when you do add-brick etc? The value has to be persistent. For example, If I create a 1x3 and then remove the middle brick to make it 1x2, the value must still be
`option afr-pending-xattr testvol-client-0,testvol-client-2` and not
`option afr-pending-xattr testvol-client-0,testvol-client-1`

Not sure if persistent-xattrs for AFR are taken care of in gd2. See https://review.gluster.org/#/c/7155/ 's commit message.

aravindavk:
Is self heal daemon started? If not, try setting below volume option to start the shd(after volume start)

    glustercli volume set testvol replicate.self-heal-daemon on

glusterd2/volgen2/clustergraph.go
aravindavk: This is not yet considered in gd2, I will fix it. Since this issue can happen only when add/remove brick. Can this be addressed in next PR?

itisravi: I guess that should be fine.

prashanthpai:
> @itisravi There is a PR #865 dealing with the above error. Its not been merged yet but you can for now work by adding a && instead of || in plugins/glustershd/actor.go line number 35.

@vpandey-RH Can you send that small fix as a separate PR ? I'd want to get that in first

PR #898, opened 2018-06-18
Removing duplicate structures by merging them at one place

rishubhjain:


prashanthpai:
@rishubhjain The duplicates are there for a reason. We may not return what we store. Besides, the structs in the core packages can have methods on them which has application specific logic.

PR #904, opened 2018-06-19
e2e: a step toward running in limited envionment

phlogistonjohn:
This small PR is a few steps toward making the e2e tests work successfully in a limited environment, such as a container.

There are more tests that need similar changes but I'd figure I'd float what I had so far past the team.


aravindavk:
retest this please

PR #905, opened 2018-06-20
use testcase Name as volume name  in e2e test cases

Madhu-1:

### DIFF
--- a/e2e/smartvol_ops_test.go
+++ b/e2e/smartvol_ops_test.go
@@ -30,7 +30,7 @@ func brickSizeTest(brickpath string, min uint64, max uint64) error {
 
 func testSmartVolumeDistribute(t *testing.T) {
 	r := require.New(t)
-	smartvolname := "sv1"
+	smartvolname := formatVolName(t.Name())
 	// create Distribute 3 Volume
 	createReq := smartvolapi.VolCreateReq{
 		Name:            smartvolname,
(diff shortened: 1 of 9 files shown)


Madhu-1:
@prashanthpai @aravindavk @rishubhjain  please review

e2e/smartvol_ops_test.go
```diff
@@ -55,7 +56,7 @@ func testSmartVolumeDistribute(t *testing.T) {
 
 func testSmartVolumeReplicate2(t *testing.T) {
 	r := require.New(t)
-	smartvolname := "sv2"
+	smartvolname := strings.Replace(t.Name(), "/", "-", 1)
```
prashanthpai: Why do you need the `strings.Replace` here ?

Madhu-1: Test name will become ```TestSmartVolume/Smartvol_Distributed_Volume```

we cannot use / in volume names. need to replace / with -

phlogistonjohn: What might be nice is a short helper function, something like testVolName(t) to make it easier for future test writers to do the right thing.

Madhu-1: done

Madhu-1:
resolved merged conflict. @aravindavk can you please review this.

PR #906, opened 2018-06-20
volgen: Include Option default values in volfiles

aravindavk:
If default value in xlators option table is not empty then
it will be added in volfiles.

Also added `afr-pending-xattr` option to client volfile.

Fixes: #890


vpandey-RH:
@aravindavk  @itisravi @karthik-us  Should we add " background-self-heal-count " option as well ?

aravindavk:
> @aravindavk @itisravi @karthik-us Should we add " background-self-heal-count " option as well ?

Check this PR, this includes all the default values in volfiles. I am not sure why the default value for `background-self-heal-count` is 8, but it is now included in volfile.

aravindavk:
@vpandey-RH, @itisravi, @karthik-us  please review

vpandey-RH:
LGTM.

Tested on a multinode setup as well.

aravindavk:
@itisravi please ack if looks good

aravindavk:
@prashanthpai please review

PR #907, opened 2018-06-20
Resolve process already running error with SHD

vpandey-RH:
Resolve process already running error with SHD


plugins/glustershd/actor.go
```diff
@@ -32,7 +32,7 @@ func (actor *shdActor) Do(v *volume.Volinfo, key string, value string, logger lo
 	case "on":
 		if isHealEnabled(v) {
 			err = daemon.Start(glustershDaemon, true, logger)
-			if err != gderrors.ErrProcessAlreadyRunning || err != nil {
+			if err != gderrors.ErrProcessAlreadyRunning && err != nil {
```
prashanthpai: While this okay, for slightly better readablity, can you please check for nil first and then the specific error ?

Below too

vpandey-RH: sure

PR #910, opened 2018-06-23
Using peerid instead of nodeid

rishubhjain:
Issue: https://github.com/gluster/glusterd2/issues/576

### DIFF
--- a/doc/xlator.md
+++ b/doc/xlator.md
@@ -22,7 +22,7 @@ volume.auth.password
 ```
 brick.id
 brick.hostname
-brick.nodeid
+brick.peerid
 brick.path
 brick.volumename
 brick.volumeid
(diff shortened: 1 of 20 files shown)


Madhu-1:
retest this please

doc/xlator.md
```diff
@@ -22,7 +22,7 @@ volume.auth.password
 ```
 brick.id
 brick.hostname
-brick.nodeid
+brick.peerid
```
aravindavk: This change is a problem if glusterfs xlator code is not changed to use `peerid`. (I just noticed `$SRC/glusterd2/brick/types.go` is already modified, we need to change the glusterfs xlator option table if `{{ brick.nodeid }}` is used

PR #918, opened 2018-06-25
Initialize slice to avoid null  in HTTP response

Madhu-1:

### DIFF
--- a/glusterd2/commands/volumes/volume-create-txn.go
+++ b/glusterd2/commands/volumes/volume-create-txn.go
@@ -130,6 +130,7 @@ func newVolinfo(req *api.VolCreateReq) (*volume.Volinfo, error) {
 		State:     volume.VolCreated,
 		Type:      voltypeFromSubvols(req),
 		DistCount: len(req.Subvols),
+		SnapList:  []string{},
 		Auth: volume.VolAuth{
 			Username: uuid.NewRandom().String(),
 			Password: uuid.NewRandom().String(),


Madhu-1:
@aravindavk please review

glusterd2/volume/volume-utils.go
```diff
@@ -191,6 +191,10 @@ func CreateVolumeInfoResp(v *Volinfo) *api.VolumeInfo {
 		Options:   v.Options,
 		Subvols:   CreateSubvolInfo(&v.Subvols),
 		Metadata:  v.Metadata,
-		SnapList:  v.SnapList,
+		SnapList:  make([]string, len(v.SnapList)),
 	}
+	for i, snap := range v.SnapList {
```
aravindavk: Instead of iterating, how about checking the `null` itself?

    if volinfo.SnapList == nil {
        volinfo.SnapList = []string{}
    }

Madhu-1: yeah, This will be simpler.

glusterd2/volume/volume-utils.go
```diff
@@ -198,4 +198,9 @@ func CreateVolumeInfoResp(v *Volinfo) *api.VolumeInfo {
 		Metadata:  v.Metadata,
 		SnapList:  v.SnapList,
 	}
+	if len(volinfo.SnapList) == 0 {
+		volinfo.SnapList = []string{}
```
prashanthpai: Well, this isn't neat although it fixes the problem.

What do you think of using `omitempty` on `snap-list` instead ?

Madhu-1: if we want to use ```omitempty``` for this. a volume may have empty metadata and options as well.same should be used  for ```options``` and ```metadata```.
let me know your thought on this.

prashanthpai: Okay.

prashanthpai: Wait, a better place to do this is at `newVolinfo` in `glusterd2/commands/volumes/volume-create-txn.go`. Let me know if that's an acceptable approach.

Madhu-1: I think this is the best place to do. as ```CreateVolumeInfoResp``` will be called by other functions for preparing volume response ex:- snapshot, VolumeExpand etc

aravindavk: @prashanthpai's suggestion is to initialize when Volume is created, so that you need not worry formatting while sending response.

Issue #819, opened 2018-05-29
Add-brick changes the replica count for the first replica set


nbalacha:
If add-brick is run without specifying the replica count, the newly added bricks are added to the first replica set instead of forming a new replica set.

On performing an add-brick , the newly added bricks end up in the middle of the list of existing bricks:

192.168.122.7:/bricks/brick2/repl-1 and 192.168.122.7:/bricks/brick2/repl-2 were added to the volume.
#>glustercli volume add-brick gd2vol 192.168.122.7:/bricks/brick2/repl-1 192.168.122.7:/bricks/brick2/repl-2

Volume Name: gd2vol
Type: Distributed-Replicate
Volume ID: bb724279-73d6-4b29-b1e0-cc368d1f92b5
State: Started
Transport-type: tcp
Options:
Number of Bricks: 2 x 2 = 6  <--- Incorrect
Brick1: 192.168.122.6:/bricks/brick2/repl-1
Brick2: 192.168.122.6:/bricks/brick2/repl-2
Brick3: 192.168.122.7:/bricks/brick2/repl-1
Brick4: 192.168.122.7:/bricks/brick2/repl-2
Brick2: 192.168.122.6:/bricks/brick2/repl-3
Brick3: 192.168.122.6:/bricks/brick2/repl-4

and has converted the first replica set to a replica 4. (I had used replica 2 while creating the volume but had not specified the replica count during the add-brick as GD1 uses the existing replica count for the volume).

 53: volume gd2vol-replicate-0  
 54:     type cluster/replicate 
 55:     subvolumes gd2vol-replicate-0-client-0 gd2vol-replicate-0-client-1 gd2vol-replicate-0-client-2 gd2vol-replicate-0-client-3
 56: end-volume

PR #928, opened 2018-06-27
add-brick: Fix issue while add brick without changing replica count

aravindavk:
If Replica count is increased during add brick then new bricks will
be added to existing sub volumes, If replica count is not changed then
new bricks should be added as new sub volume. Default behavior was
adding bricks to existing sub volumes. With this PR default behavior is
changed to add new sub volume.

Fixes: #819


aravindavk:
@Madhu-1 added tests as suggested.

aravindavk:
@prashanthpai please review

Issue #923, opened 2018-06-26
Trigger actors on volume start/stop


Madhu-1:
if the options are set during volume creation, Trigger actors to start daemons on volume Start and to stop daemons on volume stop.

aravindavk:
Also required during stop to stop the daemons when Volume is stopped

PR #939, opened 2018-06-29
Start/Stop daemons during volume start and stop 

Madhu-1:
Fixes: #923 

### DIFF
--- a/glusterd2/commands/volumes/volume-option.go
+++ b/glusterd2/commands/volumes/volume-option.go
@@ -62,37 +62,58 @@ func optionSetValidate(c transaction.TxnCtx) error {
 }
 
 type txnOpType uint8
-type volumeOpType uint8
 
 const (
 	txnDo txnOpType = iota
 	txnUndo
-	volumeSet volumeOpType = iota
-	volumeReset
 )
 
 func xlatorActionDoSet(c transaction.TxnCtx) error {
-	return xlatorAction(c, txnDo, volumeSet)
+	return xlatorAction(c, txnDo, xlator.VolumeSet)
 }
 
 func xlatorActionUndoSet(c transaction.TxnCtx) error {
-	return xlatorAction(c, txnUndo, volumeSet)
+	return xlatorAction(c, txnUndo, xlator.VolumeSet)
 }
 
 func xlatorActionDoReset(c transaction.TxnCtx) error {
-	return xlatorAction(c, txnDo, volumeReset)
+	return xlatorAction(c, txnDo, xlator.VolumeReset)
 }
 
 func xlatorActionUndoReset(c transaction.TxnCtx) error {
-	return xlatorAction(c, txnUndo, volumeReset)
+	return xlatorAction(c, txnUndo, xlator.VolumeReset)
+}
+
+func xlatorActionDoVolumeStart(c transaction.TxnCtx) error {
+	return xlatorAction(c, txnDo, xlator.VolumeStart)
+}
+
+func xlatorActionUndoVolumeStart(c transaction.TxnCtx) error {
+	return xlatorAction(c, txnUndo, xlator.VolumeStart)
+}
+
+func xlatorActionDoVolumeStop(c transaction.TxnCtx) error {
+	return xlatorAction(c, txnDo, xlator.VolumeStop)
+}
+
+func xlatorActionUndoVolumeStop(c transaction.TxnCtx) error {
+	return xlatorAction(c, txnUndo, xlator.VolumeStop)
 }
 
 // This function can be reused when volume reset operation is implemented.
 // However, volume reset can be also be treated logically as volume set but
 // with the value set to default value.
-func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, volOp volumeOpType) error {
+func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, volOp xlator.VolumeOpType) error {
+
+	var volinfo volume.Volinfo
+	if err := c.Get("volinfo", &volinfo); err != nil {
+		return err
+	}
+
 	reqOptions := make(map[string]string)
-	if volOp == volumeSet {
+	var fn func(*volume.Volinfo, string, string, xlator.VolumeOpType, log.FieldLogger) error
+	switch volOp {
+	case xlator.VolumeSet:
 		var req api.VolOptionReq
 		if err := c.Get("req", &req); err != nil {
 			return err
@@ -100,7 +121,7 @@ func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, volOp volumeOpType) err
 		for key, value := range req.Options {
 			reqOptions[key] = value
 		}
-	} else {
+	case xlator.VolumeReset:
 		var req api.VolOptionResetReq
 		if err := c.Get("req", &req); err != nil {
 			return err
@@ -113,13 +134,23 @@ func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, volOp volumeOpType) err
 			reqOptions[key] = op.DefaultValue
 		}
 
-	}
-	var volinfo volume.Volinfo
-	if err := c.Get("volinfo", &volinfo); err != nil {
-		return err
-	}
+	case xlator.VolumeStart:
+		fallthrough
 
-	var fn func(*volume.Volinfo, string, string, log.FieldLogger) error
+	case xlator.VolumeStop:
+		for _, actor := range xlator.GetOptActors() {
+			if txnOp == txnDo {
+				fn = actor.Do
+			} else {
+				fn = actor.Undo
+			}
+			if err := fn(&volinfo, "", "", volOp, c.Logger()); err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+	//The code below applies only for xlator.VolumeSet and xlator.VolumeReset operations.
 	for k, v := range reqOptions {
 		_, xl, key, err := options.SplitKey(k)
 		if err != nil {
@@ -135,7 +166,7 @@ func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, volOp volumeOpType) err
 			} else {
 				fn = xltr.Actor.Undo
 			}
-			if err := fn(&volinfo, key, v, c.Logger()); err != nil {
+			if err := fn(&volinfo, key, v, volOp, c.Logger()); err != nil {
 				return err
 			}
 		}
--- a/glusterd2/commands/volumes/volume-start.go
+++ b/glusterd2/commands/volumes/volume-start.go
@@ -64,10 +64,21 @@ func stopAllBricks(c transaction.TxnCtx) error {
 }
 
 func registerVolStartStepFuncs() {
-	transaction.RegisterStepFunc(startAllBricks, "vol-start.StartBricks")
-	transaction.RegisterStepFunc(stopAllBricks, "vol-start.StartBricksUndo")
-	transaction.RegisterStepFunc(storeVolume, "vol-start.UpdateVolinfo")
-	transaction.RegisterStepFunc(undoStoreVolume, "vol-start.UpdateVolinfo.Undo")
+	var sfs = []struct {
+		name string
+		sf   transaction.StepFunc
+	}{
+		{"vol-start.StartBricks", startAllBricks},
+		{"vol-start.StartBricksUndo", stopAllBricks},
+		{"vol-start.XlatorActionDoVolumeStart", xlatorActionDoVolumeStart},
+		{"vol-start.XlatorActionUndoVolumeStart", xlatorActionUndoVolumeStart},
+		{"vol-start.UpdateVolinfo", storeVolume},
+		{"vol-start.UpdateVolinfo.Undo", undoStoreVolume},
+	}
+	for _, sf := range sfs {
+		transaction.RegisterStepFunc(sf.sf, sf.name)
+	}
+
 }
 
 func volumeStartHandler(w http.ResponseWriter, r *http.Request) {
@@ -113,6 +124,11 @@ func volumeStartHandler(w http.ResponseWriter, r *http.Request) {
 			DoFunc:   "vol-start.UpdateVolinfo",
 			UndoFunc: "vol-start.UpdateVolinfo.Undo",
 			Nodes:    []uuid.UUID{gdctx.MyUUID},
+		},
+		{
+			DoFunc:   "vol-start.XlatorActionDoVolumeStart",
+			UndoFunc: "vol-start.XlatorActionUndoVolumeStart",
+			Nodes:    volinfo.Nodes(),
 		},
 	}
 
--- a/glusterd2/volume/store-utils.go
+++ b/glusterd2/volume/store-utils.go
@@ -182,15 +182,18 @@ func GetAllBricksInCluster() ([]brick.Brickinfo, error) {
 	return bricks, nil
 }
 
-// AreReplicateVolumesRunning retrieves the volinfo objects from GetVolumes() function
-// and checks if all replicate, Disperse volumes are stopped before
-// stopping the self heal daemon.
-func AreReplicateVolumesRunning() (bool, error) {
+// AreReplicateVolumesRunning checks if all replicate and disperse volumes are stopped.
+// The volume being acted upon is excluded from this check and
+// the volume ID of that volume needs to be volume passed as an argument.
+func AreReplicateVolumesRunning(skipVolID uuid.UUID) (bool, error) {
 	volumes, e := GetVolumes()
 	if e != nil {
 		return false, e
 	}
 	for _, v := range volumes {
+		if uuid.Equal(v.ID, skipVolID) {
+			continue
+		}
 		if (v.Type == Replicate || v.Type == Disperse || v.Type == DistReplicate || v.Type == DistDisperse) && v.State == VolStarted {
 			return true, nil
 		}
--- a/glusterd2/xlator/actors.go
+++ b/glusterd2/xlator/actors.go
@@ -8,17 +8,33 @@ import (
 
 var optionActors = make(map[string]OptionActor)
 
+//GetOptActors returns a mapping of xlator ID to its action interface implementation
+func GetOptActors() map[string]OptionActor {
+	return optionActors
+}
+
+//VolumeOpType type associated with volume operation
+type VolumeOpType uint8
+
+//constants definded to represent volume operation types
+const (
+	VolumeSet VolumeOpType = iota
+	VolumeReset
+	VolumeStart
+	VolumeStop
+)
+
 // OptionActor is an interface that contains Do and Undo methods. These methods
 // are invoked during volume set on ALL nodes that make up the volume.
 // Each plugin or xlator can provide a type satisying OptionActor interface to
 // have the xlator/feature specific logic executed during volume set. An example
 // of such logic is the task of starting and stopping daemon.
 type OptionActor interface {
-	// Do function takes volinfo, option key, option value, logger.
-	Do(*volume.Volinfo, string, string, log.FieldLogger) error
-	// Undo function takes volinfo, option key, option value and logger. The returned
+	// Do function takes volinfo, option key, option value,VolumeOpType, logger.
+	Do(*volume.Volinfo, string, string, VolumeOpType, log.FieldLogger) error
+	// Undo function takes volinfo, option key, option value,VolumeOpType, logger. The returned
 	// error is currently ignored.
-	Undo(*volume.Volinfo, string, string, log.FieldLogger) error
+	Undo(*volume.Volinfo, string, string, VolumeOpType, log.FieldLogger) error
 }
 
 // RegisterOptionActor registers a xlator's type implementing OptionActor
--- a/pkg/api/volume_req.go
+++ b/pkg/api/volume_req.go
@@ -41,11 +41,7 @@ type VolCreateReq struct {
 	Name                    string            `json:"name,omitempty"`
 	Transport               string            `json:"transport,omitempty"`
 	Subvols                 []SubvolReq       `json:"subvols"`
-	Options                 map[string]string `json:"options,omitempty"`
 	Force                   bool              `json:"force,omitempty"`
-	Advanced                bool              `json:"advanced,omitempty"`
-	Experimental            bool              `json:"experimental,omitempty"`
-	Deprecated              bool              `json:"deprecated,omitempty"`
 	Metadata                map[string]string `json:"metadata,omitempty"`
 	Flags                   map[string]bool   `json:"flags,omitempty"`
 	Size                    uint64            `json:"size"`
@@ -63,6 +59,7 @@ type VolCreateReq struct {
 	ExcludeZones            []string          `json:"exclude-zones,omitempty"`
 	SubvolZonesOverlap      bool              `json:"subvolume-zones-overlap,omitempty"`
 	SubvolType              string            `json:"subvolume-type,omitempty"`
+	VolOptionReq
 }
 
 // VolOptionReq represents an incoming request to set volume options
--- a/plugins/glustershd/actor.go
+++ b/plugins/glustershd/actor.go
@@ -2,6 +2,8 @@ package glustershd
 
 import (
 	"errors"
+	"fmt"
+	"os"
 
 	"github.com/gluster/glusterd2/glusterd2/daemon"
 	"github.com/gluster/glusterd2/glusterd2/volume"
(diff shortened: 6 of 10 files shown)


Madhu-1:

prashanthpai:
@Madhu-1 I haevn't reviewed code from this PR. The PR title is slightly misleading though.

I can accept the 7 commits style only if you have ensured that unit and functional test on each commit separately successfully passes all tests.

It's also making it hard to review commit messages such as these which need fixing:

```
check any replicate volumes are running except the passed one

pass volume ID to AreReplicateVolumesRunning  function
and check any other replicate volumes are running
or not.
```

```
export contants

export contants from actor, which will
be used by other  registered actor to
act based on the volume state
```

The embedding struct is a separate commit. I like that. But there are other problems:

* there's a commit that adds `VolumeOpType` but it's not used in that commit but in a later commit. That's a problem.
* If you found a problem with your earlier commit, you would fix it in a later commit which is bad.
* The commit "daemon management based on volume start and stop" depends on previous two commits which makes it not self-contained. I cannot revert just one and expect things to work.
* If the `check any replicate volumes are running except the passed one` commit is fixing a separate bug, then the commit should describe what it's fixing and independently pass test.

If you really want all this to be not a single commit, I'd like to see 3 separate but self-contained commits to:

* Fix self heal bug
* Start and stop daemons during volume start and stop
* Embed struct

Madhu-1:
@prashanthpai merged multiple commits.

Madhu-1:
retest this please, CI failure is not related to this PR

Madhu-1:
retest this please

glusterd2/commands/volumes/volume-option.go
```diff
@@ -62,45 +62,66 @@ func optionSetValidate(c transaction.TxnCtx) error {
...
-	return xlatorAction(c, txnDo, volumeReset)
+	return xlatorAction(c, txnDo, xlator.VolumeReset)
 }
 
 func xlatorActionUndoReset(c transaction.TxnCtx) error {
-	return xlatorAction(c, txnUndo, volumeReset)
+	return xlatorAction(c, txnUndo, xlator.VolumeReset)
+}
+
+func xlatorActionDoStart(c transaction.TxnCtx) error {
```
prashanthpai: The name is misleading without any reference to volume. I suggest `xlatorActionDoVolumeStart`. Follow the same for all 4 functions you've added.

glusterd2/commands/volumes/volume-option.go
```diff
@@ -113,13 +134,25 @@ func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, volOp volumeOpType) err
...
 		}
 
-	}
-	var volinfo volume.Volinfo
-	if err := c.Get("volinfo", &volinfo); err != nil {
-		return err
+	case xlator.VolumeStart:
+		fallthrough
+	case xlator.VolumeStop:
+		//start the daemon, if daemon default values is "on",
```
prashanthpai: `value` (singular)

glusterd2/commands/volumes/volume-option.go
```diff
@@ -113,13 +134,25 @@ func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, volOp volumeOpType) err
...
+		//or if particular  daemon  options are set to "on" in volume option
```
prashanthpai: `option` (singular)

there might be extra spaces between words

glusterd2/commands/volumes/volume-start.go
```diff
@@ -98,14 +114,42 @@ func volumeStartHandler(w http.ResponseWriter, r *http.Request) {
...
 		},
-	}
+		{
+			DoFunc:   "vol-start.XlatorActionDoStart",
+			UndoFunc: "vol-start.XlatorActionUndoStart",
+			Nodes:    volinfo.Nodes(),
+		},
 
+		{
+			DoFunc:   "vol-start.UpdateVolinfo",
```
prashanthpai: `glusterd2/commands/volumes/volume-option.go` has the order steps reversed but I think this should be okay as volfiles are regenrated anyway during volume set on stopped volume and daemons will get the new volfiles.

glusterd2/volume/store-utils.go
```diff
@@ -189,14 +189,17 @@ func GetAllBricksInCluster() ([]brick.Brickinfo, error) {
 }
 
 // AreReplicateVolumesRunning retrieves the volinfo objects from GetVolumes() function
-// and checks if all replicate, Disperse volumes are stopped before
-// stopping the self heal daemon.
-func AreReplicateVolumesRunning() (bool, error) {
+// and checks if all replicate, Disperse volumes are stopped except input volume ID
+//before stopping the self heal daemon.
+func AreReplicateVolumesRunning(ID uuid.UUID) (bool, error) {
```
prashanthpai: This variable can be named better - `skipPeer` or `ignorePeer`

Can you describe why the peer is skipped ? I'd like to understand this better

glusterd2/xlator/actors.go
```diff
@@ -6,24 +6,36 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-var optionActors = make(map[string]OptionActor)
+//OptionActors in-memory to store registered xlator type and OptionActor
```
prashanthpai: Do not export this. Write a function that returns the map.

plugins/quota/actor.go
```diff
@@ -37,7 +37,7 @@ func isQuotadStopRequired(volumes []*volume.Volinfo) bool {
 	return true
 }
 
-func (actor *quotadActor) Do(v *volume.Volinfo, key, value string, logger log.FieldLogger) error {
+func (actor *quotadActor) Do(v *volume.Volinfo, key, value string, volopt xlator.VolumeOpType, logger log.FieldLogger) error {
```
prashanthpai: The variable name `volopt` often means 'volume option. Please use `volOp` to clearly indicate that it's a volume operation. Same comment below.

plugins/glustershd/actor.go
```diff
@@ -19,14 +21,52 @@ const (
...
+	if opt == xlator.VolumeStart {
+		for k, v := range v.Options {
+			_, _, key, err := options.SplitKey(k)
+			if err != nil {
+				return err
+			}
+			if key == selfHealKey && v == "off" {
+				return nil
+			}
+
```
prashanthpai: extra line

plugins/glustershd/actor.go
```diff
@@ -19,14 +21,52 @@ const (
...
+	if voltype != "Replicate" && voltype != "Disperse" {
 		return nil
 	}
+
 	glustershDaemon, err := newGlustershd()
 	if err != nil {
 		return err
 	}
+	if opt == xlator.VolumeStart {
```
prashanthpai: I'd like to see explicit switch case for each of the four operations to make it more readable. Currently, it is implicitly implied that if it's not start or stop, rest of the code is for set or reset

plugins/glustershd/actor.go
```diff
@@ -38,27 +78,63 @@ func (actor *shdActor) Do(v *volume.Volinfo, key string, value string, logger lo
...
+
+	if voltype != "Replicate" && voltype != "Disperse" {
 		return nil
 	}
 
 	glustershDaemon, err := newGlustershd()
 	if err != nil {
 		return err
 	}
+	if opt == xlator.VolumeStart {
```
prashanthpai: Same comment here

plugins/glustershd/actor.go
```diff
@@ -38,27 +78,63 @@ func (actor *shdActor) Do(v *volume.Volinfo, key string, value string, logger lo
...
+		for k, v := range v.Options {
+			_, _, key, err := options.SplitKey(k)
+			if err != nil {
+				return err
+			}
+			if key == selfHealKey && v == "off" {
+				return nil
+			}
+
```
prashanthpai: empty line

prashanthpai:
@vpandey-RH Please review the self heal parts of change

glusterd2/commands/volumes/volume-option.go
Madhu-1: done

glusterd2/commands/volumes/volume-option.go
Madhu-1: done

glusterd2/commands/volumes/volume-option.go
Madhu-1: done

glusterd2/volume/store-utils.go
Madhu-1: This is volume ID, not the peerID.  to check any replicate volumes are running we get volumes from the store, as this volume is not yet updated in the store, need to skip this volume.

glusterd2/xlator/actors.go
Madhu-1: done

plugins/quota/actor.go
Madhu-1: done

plugins/glustershd/actor.go
Madhu-1: removed

plugins/glustershd/actor.go
Madhu-1: done

plugins/glustershd/actor.go
Madhu-1: done

plugins/glustershd/actor.go
Madhu-1: removed

plugins/glustershd/actor.go
```diff
@@ -19,69 +21,148 @@ const (
...
 	}
-
-	switch value {
-	case "on":
-		if isHealEnabled(v) {
-			err = daemon.Start(glustershDaemon, true, logger)
-			if err != nil && err != gderrors.ErrProcessAlreadyRunning {
+	switch volOp {
+	case xlator.VolumeStart:
+		for k, v := range v.Options {
```
prashanthpai: I didn't catch this earlier. You don't have to loop over all options here as it's a map. Specifically look for `replicate.self-heal-daemon` or `afr.self-heal-daemon`

plugins/glustershd/actor.go
```diff
@@ -19,69 +21,148 @@ const (
...
-	case "off":
-		if !isHealEnabled(v) {
-			if v.State != volume.VolStarted {
-				return errors.New("volume should be in started state")
-			}
-			err = daemon.Start(glustershDaemon, true, logger)
-			if err != nil && err != gderrors.ErrProcessAlreadyRunning {
+	switch volOp {
+	case xlator.VolumeStart:
+		for k, v := range v.Options {
```
prashanthpai: Same comment here

plugins/glustershd/actor.go
```diff
@@ -19,69 +21,148 @@ const (
...
+		if err != nil && err != gderrors.ErrProcessAlreadyRunning {
+			return err
+		}
+
+	case xlator.VolumeStop:
+		isVolRunning, err := volume.AreReplicateVolumesRunning(v.ID)
+		if err != nil {
+			return err
+		}
+
```
prashanthpai: There is one SHD per node. When multiple volumes are present, we'll have to restart SHD when any one volume is stopped. And that restart should fetch the new volfile which doesn't contain the stopped volume details. This needs more thinking.

@aravindavk Thoughts ?

aravindavk: today I opened an issue for the same. We need to regenerate cluster volfiles on volume start and stop(Notify not required). when these daemons restarted they will get latest volfiles automatically.

https://github.com/gluster/glusterd2/issues/1005

aravindavk:
> But from what I do see I think it makes me like my daemon management proposal even more... this code could worry less about started/stopped state and concentrate on simply communicating the desired state to the daemon manager.

Problem is not about bringing daemon to desired state, problem here is finding that desired state based on many other factors. 

The issue here is detecting the volfile changes scenario and restarting the cluster level processes. For example, Self Heal daemon needs to be restarted when volfile changes. Volfile can change due to one of the following,

- A new Volume is created and started
- A volume is stopped
- Volume set
- Self-heal disabled for one Volume
- Self-heal enabled for one Volume

plugins/glustershd/actor.go
Madhu-1: (this is not only for the self-heal daemon).This is intended to leave the action to registered actors. whether they need to start or stopped based on the default options.

plugins/glustershd/actor.go
Madhu-1: v.Options may contain the (self-heal)key if the user gave it during volume create. or else start the daemon based on the default option.

plugins/glustershd/actor.go
prashanthpai: I'm gonna take the liberty move this out of the current sprint until volgen changes get in.

aravindavk: PRs are not related. Side effect of not having volgen patch and having only this patch is cluster level daemons will do extra work even for the stopped volumes. (Like crawl and other book keeping activities). We can review this PR and take it up since no changes required to this part of code when #1005 is addressed.

glusterd2/commands/volumes/volume-start.go
```diff
@@ -98,14 +114,42 @@ func volumeStartHandler(w http.ResponseWriter, r *http.Request) {
...
+			UndoFunc: "vol-start.UpdateVolinfo.Undo",
+			Nodes:    []uuid.UUID{gdctx.MyUUID},
+		},
+		{
+			DoFunc: "vol-start.NotifyVolfileChange",
```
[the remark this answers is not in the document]
Madhu-1: @aravindavk  just a question, don't we need to NotifyVolfile change, if the daemon is already started.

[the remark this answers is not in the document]
aravindavk: - Volume is just now started, so no mount/client exists yet.
- glusterfsd also started now itself, they get latest volfiles
- Cluster level daemons will get restarted with Actor model, so they will get latest volfiles

aravindavk:
Please resolve the conflicts.

Any more changes expected to this patch? PR related to triggering volgen on Volume start and Stop is merged https://github.com/gluster/glusterd2/pull/1015

pkg/api/volume_req.go
```diff
@@ -27,16 +27,13 @@ type SubvolReq struct {
...
-	Force        bool              `json:"force,omitempty"`
-	Advanced     bool              `json:"advanced,omitempty"`
-	Experimental bool              `json:"experimental,omitempty"`
-	Deprecated   bool              `json:"deprecated,omitempty"`
-	Metadata     map[string]string `json:"metadata,omitempty"`
-	Flags        map[string]bool   `json:"flags,omitempty"`
+	Name      string      `json:"name,omitempty"`
+	Transport string      `json:"transport,omitempty"`
+	Subvols   []SubvolReq `json:"subvols"`
+	VolOptionReq
```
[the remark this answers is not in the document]
Madhu-1: @phlogistonjohn yes, during the volume creation we can set the options on the volume.
if a volume options changes we need to change it in volume create also.

Madhu-1:
@aravindavk resolved merge conflicts, and made changes please review

plugins/glustershd/actor.go
aravindavk: this is only for glustershd right? other actors will look for there keys to stop and start the daemons.

Madhu-1: @aravindavk  yes, its actors work to start/stop daemons. I think I need to edit my previous comment.
and update the code as per @prashanthpai comment

plugins/glustershd/actor.go
Madhu-1: sure i will update this also

glusterd2/commands/volumes/volume-option.go
```diff
@@ -62,45 +62,66 @@ func optionSetValidate(c transaction.TxnCtx) error {
...
 }
 
 // This function can be reused when volume reset operation is implemented.
 // However, volume reset can be also be treated logically as volume set but
 // with the value set to default value.
-func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, volOp volumeOpType) error {
+func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, volOp xlator.VolumeOpType) error {
 	reqOptions := make(map[string]string)
-	if volOp == volumeSet {
+	var volinfo volume.Volinfo
```
prashanthpai: Can you move this line to the place just before where its unmarshalled ?

```go
var volinfo volume.Volinfo
if err := c.Get("volinfo", &volinfo); err != nil {
```

glusterd2/commands/volumes/volume-option.go
```diff
@@ -113,13 +134,25 @@ func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, volOp volumeOpType) err
...
 		}
 
-	}
-	var volinfo volume.Volinfo
-	if err := c.Get("volinfo", &volinfo); err != nil {
-		return err
+	case xlator.VolumeStart:
+		fallthrough
+	case xlator.VolumeStop:
+		//start the daemon, if daemon option default value is "on",
```
prashanthpai: The comment is incorrect. The actor framework is generic and is not limited to volume starting or stopping. Please remove this comment.

glusterd2/commands/volumes/volume-option.go
```diff
@@ -113,13 +134,25 @@ func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, volOp volumeOpType) err
...
+		for _, actor := range xlator.GetOptActors() {
+			if txnOp == txnDo {
+				fn = actor.Do
+			} else {
+				fn = actor.Undo
+			}
+			if err := fn(&volinfo, "", "", volOp, c.Logger()); err != nil {
+				return err
+			}
+		}
```
prashanthpai: Shouldn't you return here ? 

```go
return fn(&volinfo, "", "", volOp, c.Logger()
```
The `fn()` is called again below. And that is only for `VolumeSet` and `VolumeReset` cases

glusterd2/volume/store-utils.go
```diff
@@ -189,14 +189,18 @@ func GetAllBricksInCluster() ([]brick.Brickinfo, error) {
 }
 
 // AreReplicateVolumesRunning retrieves the volinfo objects from GetVolumes() function
-// and checks if all replicate, Disperse volumes are stopped before
-// stopping the self heal daemon.
-func AreReplicateVolumesRunning() (bool, error) {
+// and checks if all replicate, Disperse volumes are stopped except input volume ID
+//before stopping the self heal daemon.
```
prashanthpai: * needs space after `//` for these two line 193 and 194
* s/Disperse/disperse

The comment can be improved as follows:

```
AreReplicateVolumesRunning checks if all replicate and disperse volumes are stopped.
The volume being acted upon is excluded from this check and the volume ID of that volume
needs to be volume passed as an argument.
```

glusterd2/xlator/actors.go
```diff
@@ -6,19 +6,36 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+//optionActors in-memory to store registered xlator type and OptionActor
```
prashanthpai: It is implicit that this is in-memory. A comment on a map should describe what the key and values are.

glusterd2/xlator/actors.go
```diff
@@ -6,19 +6,36 @@ import (
...
 var optionActors = make(map[string]OptionActor)
 
+//GetOptActors returns unexported optionActors
```
prashanthpai: You don't have to mention that this is unexported. Just this should be good:

```
GetOptActors returns a mapping of xlator ID to its action interface implementation
```

plugins/glustershd/actor.go
```diff
@@ -13,75 +14,149 @@ import (
 
 var names = [...]string{"replicate", "afr"}
 
+var selfhealKeys = []string{"replicate.self-heal-daemon", "afr.self-heal-daemon"}
```
prashanthpai: `selfhealKeys` can be derived from `names` and `selfHealKey` variables.

plugins/glustershd/actor.go
```diff
@@ -13,75 +14,149 @@ import (
...
 	selfHealKey = "self-heal-daemon"
 )
 
 type shdActor struct{}
 
-func (actor *shdActor) Do(v *volume.Volinfo, key string, value string, logger log.FieldLogger) error {
-	if key != selfHealKey {
+func (actor *shdActor) Do(v *volume.Volinfo, key string, value string, volOp xlator.VolumeOpType, logger log.FieldLogger) error {
+
+	voltype := v.Type.String()
```
prashanthpai: Use the volume type constants from volume package directly. Do not convert it to strings

plugins/glustershd/actor.go
```diff
@@ -13,75 +14,149 @@ import (
...
 		}
 	}
 	return nil
 }
 
-func (actor *shdActor) Undo(v *volume.Volinfo, key string, value string, logger log.FieldLogger) error {
-	if key != selfHealKey {
+func (actor *shdActor) Undo(v *volume.Volinfo, key string, value string, volOp xlator.VolumeOpType, logger log.FieldLogger) error {
+
+	voltype := v.Type.String()
```
prashanthpai: Same comment here

plugins/glustershd/actor.go
```diff
@@ -13,75 +14,149 @@ import (
...
+
 const (
 	selfHealKey = "self-heal-daemon"
 )
 
 type shdActor struct{}
 
-func (actor *shdActor) Do(v *volume.Volinfo, key string, value string, logger log.FieldLogger) error {
-	if key != selfHealKey {
+func (actor *shdActor) Do(v *volume.Volinfo, key string, value string, volOp xlator.VolumeOpType, logger log.FieldLogger) error {
```
prashanthpai: @vpandey-RH Can you review this file ?

vpandey-RH: @prashanthpai I have tested the glustershd behaviour with this patch. Seems to work fine. 
Cannot comment on the code as I am not very aware of the code architecture being followed here.

plugins/glustershd/actor.go
```diff
@@ -13,75 +14,149 @@ import (
...
 	}
+
 	glustershDaemon, err := newGlustershd()
 	if err != nil {
 		return err
 	}
+	switch volOp {
+	case xlator.VolumeStart:
+		for _, key := range selfhealKeys {
+			if v, ok := v.Options[key]; ok && v == "off" {
```
vpandey-RH: v here is used to store volinfo as well as value for a xlator option. Can we change one of them maybe, to avoid any kind of confusion ?

Madhu-1:
@prashanthpai @vpandey-RH  addressed review comments please review

glusterd2/commands/volumes/volume-option.go
```diff
@@ -62,45 +62,66 @@ func optionSetValidate(c transaction.TxnCtx) error {
...
 // This function can be reused when volume reset operation is implemented.
 // However, volume reset can be also be treated logically as volume set but
 // with the value set to default value.
-func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, volOp volumeOpType) error {
+func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, volOp xlator.VolumeOpType) error {
 	reqOptions := make(map[string]string)
-	if volOp == volumeSet {
+	var fn func(*volume.Volinfo, string, string, xlator.VolumeOpType, log.FieldLogger) error
+	var volinfo volume.Volinfo
+
```
prashanthpai: nit: Declare variables right before they are unmarshalled. Let's focus on better readability and leave the pre-declaration optimisation to the compiler.

```go
var volinfo volume.Volinfo
if err := c.Get("volinfo", &volinfo); err != nil {
....
```

glusterd2/commands/volumes/volume-option.go
```diff
@@ -113,13 +134,23 @@ func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, volOp volumeOpType) err
...
+		fallthrough
+
+	case xlator.VolumeStop:
+		for _, actor := range xlator.GetOptActors() {
+			if txnOp == txnDo {
+				fn = actor.Do
+			} else {
+				fn = actor.Undo
+			}
+			if err := fn(&volinfo, "", "", volOp, c.Logger()); err != nil {
```
prashanthpai: Concern unrelated to PR: If one of the actors fail, the undo step will run all of them. This requires undo functions to be written in certain way. The actor framework function implementations has gotten complicated over time. May be declarative daemon management will help. I don't know.

glusterd2/commands/volumes/volume-option.go
```diff
@@ -62,45 +62,66 @@ func optionSetValidate(c transaction.TxnCtx) error {
...
+func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, volOp xlator.VolumeOpType) error {
 	reqOptions := make(map[string]string)
-	if volOp == volumeSet {
+	var fn func(*volume.Volinfo, string, string, xlator.VolumeOpType, log.FieldLogger) error
+	var volinfo volume.Volinfo
+
+	if err := c.Get("volinfo", &volinfo); err != nil {
+		return err
+	}
+
```
prashanthpai: Same comment. Declare other two variables here.

glusterd2/commands/volumes/volume-option.go
```diff
@@ -113,13 +134,23 @@ func xlatorAction(c transaction.TxnCtx, txnOp txnOpType, volOp volumeOpType) err
...
+				fn = actor.Do
+			} else {
+				fn = actor.Undo
+			}
+			if err := fn(&volinfo, "", "", volOp, c.Logger()); err != nil {
+				return err
+			}
+		}
+		return nil
 	}
```
prashanthpai: Please add comment here that the code that follows applies only to `xlator.VolumeSet` and `xlator.VolumeReset` operations only.

Madhu-1:
@prashanthpai addressed review comments

Madhu-1:
@prashanthpai added e2e test cases for glustershd

Issue #921, opened 2018-06-26
Events API enhancements


aravindavk:
- [ ]  Add support for events listener port for glusterfs daemons like `glusterfsd`, `glustershd`, `gsyncd`, `quotad` etc.
- [ ] Webhook Test API  (Say `POST /webhook/test`), From all peer nodes send a dummy HTTP POST message to webhook URL. return success if Webhook is reachable
- [ ] All event names to new format(Ex: `VOLUME_CREATE` to `volume.created`)
- [ ] On receive of local event from other gluster daemon/client, push global event in etcd
- [ ] Add Events listener to listen to events originated from Glusterd2 and push it to webhook handler.

aravindavk:
> Add support for events listener port for glusterfs daemons like glusterfsd, glustershd, gsyncd, quotad etc.

Currently in Gd1, Events Port is defaulted to `#define EVENT_PORT 24009`, This should change to accept event-port as parameter. The changes are expected in two places, 1) Glusterfs source to add this new argument and use it whenever required. 2) Modify Glusterd2, daemon management code to include this argument while spawning every daemon.

For example, to enable brick process to send events, glusterfsd can accept event port as `glusterfsd --event-port=%d`

atinmu:
@Madhu-1 Please open a new issue tracking the pending work and close it

Madhu-1:
@aravindavk #942 is merged now. I have opened a new issue to track pending tasks, can we close this one?

PR #942, opened 2018-06-29
Events API enhancements

Madhu-1:
Partial #921 


Madhu-1:

Madhu-1:
@prashanthpai @aravindavk @kshlm @rishubhjain    please review

glusterd2/events/eventhandler.go
```diff
@@ -154,3 +162,66 @@ func (h *handler) Handle(e *api.Event) {
...
+		return ""
+	}
+	return ss
+}
+
+//WebhookPublish sends event to registered webhooks
+func WebhookPublish(webhook *eventsapi.Webhook, e *api.Event) error {
+	message, err := json.Marshal(e)
+	if err != nil {
+		log.WithError(err).Error("failed to marshal event")
```
rishubhjain: We can also log the event here, just the event name. log with field.

glusterd2/events/eventhandler.go
```diff
@@ -154,3 +162,66 @@ func (h *handler) Handle(e *api.Event) {
...
+	if err != nil {
+		log.WithError(err).Error("error forming the request object")
+		return err
+	}
+	req.Header.Set("Content-Type", "application/json")
+	if webhook.Token != "" {
+		req.Header.Set("Authorization", "bearer "+webhook.Token)
+	}
+
+	if webhook.Secret != "" {
```
rishubhjain: Is it possible that use can provide both Token and Secret key?

plugins/events/rest.go
```diff
@@ -122,3 +125,72 @@ func eventsListHandler(w http.ResponseWriter, r *http.Request) {
 
 	restutils.SendHTTPResponse(ctx, w, http.StatusOK, events)
 }
+
+func checkConnection(c transaction.TxnCtx) error {
+	var req eventsapi.Webhook
+
```
rishubhjain: Unwanted line

plugins/events/rest.go
```diff
@@ -122,3 +125,72 @@ func eventsListHandler(w http.ResponseWriter, r *http.Request) {
...
+	if err := restutils.UnmarshalRequest(r, &req); err != nil {
+		restutils.SendHTTPError(
+			ctx, w, http.StatusBadRequest,
+			errors.ErrJSONParsingFailed)
+		return
+	}
+
+	if req.URL == "" {
```
rishubhjain: TRIVIAL: Something I read recently, may be usefull, instead of comparing string like this its faster to compare the length of string, though internally this comparison happens (implicit). Also we can trim the strings before comparing,

plugins/events/rest.go
```diff
@@ -122,3 +125,72 @@ func eventsListHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	if err := txn.Ctx.Set("req", &req); err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+
+	if err := txn.Do(); err != nil {
+		logger.WithError(err).Error("webhook-test transaction failed")
+		status, err := restutils.ErrToStatusCode(err)
```
rishubhjain: This can be used at other places as well right?, where ever the status is generic

rishubhjain:
@Madhu-1 Also if possible log with field whenever possible, it helps in debugging.

plugins/events/rest.go
Madhu-1: I just used the code from [here](https://github.com/gluster/glusterd2/blob/master/plugins/events/rest.go#L27 ). Its a matter of taste,  both will generate the same code [check  here](https://groups.google.com/forum/#!topic/golang-nuts/5bxsM_sPQyI)

glusterd2/events/eventhandler.go
Madhu-1: done

plugins/events/rest.go
Madhu-1: This function can't be used for system generated error msgs like failed unmarshal.

plugins/events/hooks.go
```diff
@@ -0,0 +1,53 @@
...
+	gd2events "github.com/gluster/glusterd2/glusterd2/events"
+	"github.com/gluster/glusterd2/glusterd2/gdctx"
+	"github.com/gluster/glusterd2/pkg/api"
+	eventsapi "github.com/gluster/glusterd2/plugins/events/api"
+	"github.com/pborman/uuid"
+
+	log "github.com/sirupsen/logrus"
+)
+
+type hooks struct{}
```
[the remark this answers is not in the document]
Madhu-1: done

glusterd2/events/eventhandler.go
```diff
@@ -154,3 +162,69 @@ func (h *handler) Handle(e *api.Event) {
...
+	}
+
+	body := strings.NewReader(string(message))
+
+	req, err := http.NewRequest("POST", webhook.URL, body)
+	if err != nil {
+		log.WithError(err).Error("error forming the request object")
+		return err
+	}
+	req.Header.Set("Content-Type", "application/json")
```
phlogistonjohn: It seems a bit odd to me that this api is assuming the other side uses json. Perhaps we should consider allowing per api serialization so that the caller can provide this with whatever body the api on the other side actually expects (perhaps the remote api works with xml for example)

glusterd2/events/eventhandler.go
```diff
@@ -154,3 +162,69 @@ func (h *handler) Handle(e *api.Event) {
...
+	}
+
+	if webhook.Secret != "" {
+		token := getJWTToken(e.Name, webhook.Secret)
+		req.Header.Set("Authorization", "bearer "+token)
+	}
+
+	tr := &http.Transport{
+		DisableCompression:    true,
+		DisableKeepAlives:     true,
```
phlogistonjohn: Are these good defaults? It seems odd to me to see these hard-coded here, although I'll admit to not entirely understanding this subsystem.
At the very least I think comments explaining why these settings were chosen would be helpful.

plugins/events/init.go
```diff
@@ -53,5 +60,6 @@ func (p *Plugin) RestRoutes() route.Routes {
 // RegisterStepFuncs registers transaction step functions with
 // Glusterd Transaction framework
 func (p *Plugin) RegisterStepFuncs() {
+	registerWebhookTestStepFuncs()
```
phlogistonjohn: This confuses me. Why does webhooks support need transaction steps?

plugins/events/init.go
```diff
@@ -26,6 +26,13 @@ func (p *Plugin) RestRoutes() route.Routes {
 			Version:     1,
 			RequestType: utils.GetTypeString((*events.Webhook)(nil)),
 			HandlerFunc: webhookAddHandler},
+		route.Route{
+			Name:        "WebhookTest",
+			Method:      "POST",
+			Pattern:     "/events/webhook/test",
+			Version:     1,
+			RequestType: utils.GetTypeString((*events.Webhook)(nil)),
+			HandlerFunc: webhookTestHandler},
```
phlogistonjohn: Hmm. WebhookTest sounds like this is for test code but it looks like its a permanent part of the server. What's this actually do?

glusterd2/events/eventhandler.go
aravindavk: JSON is the only supported format now.

plugins/events/init.go
aravindavk: Transaction steps is required since this needs to be run in multiple nodes

plugins/events/init.go
aravindavk: This API is to test that if the webhook can be reached from all gluster nodes or not

aravindavk:
> I'll admit to not fully understanding what the end goal of these changes are and unfortunately the commit messages don't clear things up too much for me.

@Madhu-1 please update the commit message with description. 

@phlogistonjohn this feature is ported from glusterd1, no change wrt design. Documentation is available here https://docs.gluster.org/en/latest/Administrator%20Guide/Events%20APIs/

aravindavk:
@Madhu-1 I just checked the commit messages, description is available in commit messages.

Issue #835, opened 2018-06-01
Volume delete looks for volfile on disk


prashanthpai:
...but it's in store.

```log
WARN[2018-06-01 10:32:58.139284] failed to delete brick volfile                brick=/export/brick1/data error="remove /tmp/gd2_func_test/w1/vols/test/test.1f982c00-55af-40ab-825d-399476e41abb.export-brick1-data.vol: no such file or directory" reqid=ec269225-b37f-4de2-897b-6eb2ec1edbe4 source="[volume-delete.go:33:volumes.deleteVolfiles]" txnid=c6f72e00-dcee-42a6-8d51-57efcafa7155
```

prashanthpai:
The volfile isn't deleted from store:

```sh
[ppai@gd2-1 glusterd2]$ ETCDCTL_API=3 etcdctl get --keys-only=true --prefix=true "gluster-e60591eb-c6a6-4e21-ad47-eb3dd667775f/volfiles" --endpoints=[127.0.0.1:2379]
gluster-e60591eb-c6a6-4e21-ad47-eb3dd667775f/volfiles/1f982c00-55af-40ab-825d-399476e41abb-gluster/bitd

gluster-e60591eb-c6a6-4e21-ad47-eb3dd667775f/volfiles/1f982c00-55af-40ab-825d-399476e41abb-gluster/scrub

gluster-e60591eb-c6a6-4e21-ad47-eb3dd667775f/volfiles/gfproxy-client/test

gluster-e60591eb-c6a6-4e21-ad47-eb3dd667775f/volfiles/gfproxyd/test

gluster-e60591eb-c6a6-4e21-ad47-eb3dd667775f/volfiles/gluster/glustershd

gluster-e60591eb-c6a6-4e21-ad47-eb3dd667775f/volfiles/gluster/quotad

gluster-e60591eb-c6a6-4e21-ad47-eb3dd667775f/volfiles/rebalance/test

gluster-e60591eb-c6a6-4e21-ad47-eb3dd667775f/volfiles/test.1f982c00-55af-40ab-825d-399476e41abb.export-brick1-data
```

prashanthpai:
@Madhu-1 pointed out that the volume delete code uses older volgen code. There are multiple places in store where a volume volfile@ is stored in etcd - gfproxy etc. @aravindavk What's a clean way to do it in volgen2 ? Considering that rebalance, gfproxy, quotad volfile etc will have to change.

aravindavk:
[...]

PR #949, opened 2018-07-03
Volgen volume delete

aravindavk:
Delete Volfiles belonging to a Volume on Volume delete.

Fixes: #835


aravindavk:
Note to reviewers: First commit contains all the changes required to fix the issue. Second commit is only rename of `volgen2`=>`volgen` and respective import.

aravindavk:
Create Two Volumes
    
    # glustercli volume create --name gv1 gluster1.example.com:/bricks/b1 gluster1.example.com:/bricks/b2 --allow-root-dir --create-brick-dir --reuse-bricks
    # glustercli volume create --name gv2 gluster1.example.com:/bricks/b3 gluster1.example.com:/bricks/b4 --allow-root-dir --create-brick-dir --reuse-bricks

Get List of Volfiles:

    # curl -XGET http://localhost:24007/v1/volfiles | python -m json.tool
    [
        "f2005c6e-1f58-49cc-a26f-7dc5ef0d8114-gluster/bitd",
        "f2005c6e-1f58-49cc-a26f-7dc5ef0d8114-gluster/scrub",
        "gfproxy-client/gv1",
        "gfproxy-client/gv2",
        "gfproxyd/gv1",
        "gfproxyd/gv2",
        "gluster/glustershd",
        "gluster/quotad",
        "gv1",
        "gv1.f2005c6e-1f58-49cc-a26f-7dc5ef0d8114.bricks-b1",
        "gv1.f2005c6e-1f58-49cc-a26f-7dc5ef0d8114.bricks-b2",
        "gv2",
        "gv2.f2005c6e-1f58-49cc-a26f-7dc5ef0d8114.bricks-b3",
        "gv2.f2005c6e-1f58-49cc-a26f-7dc5ef0d8114.bricks-b4",
        "rebalance/gv1",
        "rebalance/gv2"
    ]
    
Delete one Volume:

    # glustercli volume delete gv1                                                                                                                                              
    Volume gv1 deleted successfully

Get List of Volfiles again:

    # curl -XGET http://localhost:24007/v1/volfiles | python -m json.tool
    [
        "f2005c6e-1f58-49cc-a26f-7dc5ef0d8114-gluster/bitd",
        "f2005c6e-1f58-49cc-a26f-7dc5ef0d8114-gluster/scrub",
        "gfproxy-client/gv2",
        "gfproxyd/gv2",
        "gluster/glustershd",
        "gluster/quotad",
        "gv2",
        "gv2.f2005c6e-1f58-49cc-a26f-7dc5ef0d8114.bricks-b3",
        "gv2.f2005c6e-1f58-49cc-a26f-7dc5ef0d8114.bricks-b4",
        "rebalance/gv2"
    ]

glusterd2/volgen2/store-utils.go
```diff
@@ -46,5 +58,34 @@ func GetVolfile(volfileID string) ([]byte, error) {
...
+	if e = json.Unmarshal(resp.Kvs[0].Value, &v); e != nil {
+		log.WithError(e).Error("Failed to unmarshal the data into volfiledata object")
+		return nil, e
+	}
+	return []byte(v.Content), nil
+}
+
+// DeleteVolumeVolfiles deletes all the Volfiles belonging to a Volume
+func DeleteVolumeVolfiles(volname string) error {
+	resp, e := store.Get(context.TODO(), volfilePrefix, clientv3.WithPrefix())
```
rafikc30: Snapshot client volfiles will be under volfilePrefix/snaps/<volname>.  Does store.Get returns the files by recursive crawling?

glusterd2/volgen2/generate.go
```diff
@@ -37,7 +37,7 @@ func generateClusterLevelVolfiles(clusterinfo []*volume.Volinfo, xopts *map[stri
 				if err != nil {
 					return err
 				}
-				err = save(peerid.String()+"-"+volfile.FileName, volfiledata)
+				err = save(peerid.String()+"-"+volfile.FileName, volfiledata, "-")
```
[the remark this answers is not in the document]
aravindavk: Got it. I need to regenerate Cluster Volfiles

glusterd2/volgen2/store-utils.go
aravindavk: Mostly yes. I will verify this use case.

aravindavk:
> Not related to this patch, When we generate cluster level volfiles, I think we don't need to do for stopped volumes.

Thanks, if that is the case then Volume Start will trigger cluster volfiles re-generate. Need to enhance the notification mechanism as well to notify only Cluster level daemons.

glusterd2/volgen2/store-utils.go
aravindavk: `clientv3.WithPrefix()` will include all keys with the prefix

aravindavk:
retest this please

aravindavk:
Added volfile re-generation after Volume delete. Please review

glusterd2/volgen2/generate.go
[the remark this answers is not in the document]
aravindavk: After creating new issue https://github.com/gluster/glusterd2/issues/1005, I realized that we don't need to re-generate cluster level volfiles since volfiles are re-generated after every volume stop.

@rafikc30 Do you see any issues here?

glusterd2/volgen2/store-utils.go
prashanthpai: Suggestion: If all volfiles belonging to a volume are under one prefix - `volfiles/<volname>, you can delete all with one op using `WithPrefix()`

glusterd2/volgen/generate.go
```diff
@@ -49,7 +49,7 @@ func generateClusterLevelVolfiles(clusterinfo []*volume.Volinfo, xopts *map[stri
 			if err != nil {
 				return err
 			}
-			err = save(volfile.FileName, volfiledata)
+			err = save(volfile.FileName, volfiledata, "-")
```
prashanthpai: Passing `-` might be slightly confusing here.

You can have volfile type (as enums) and pass that may be ?

Suggestion:
Also, except for user mounting a volume usecase, we have tight control over volfile ID. This is because we pass `volfile-id` as an arg to all daemons. So we can come up with a common convention to name volfiles as we wish to make code simpler.

glusterd2/volgen/store-utils.go
```diff
@@ -0,0 +1,91 @@
...
+	return []byte(v.Content), nil
+}
+
+// DeleteVolumeVolfiles deletes all the Volfiles belonging to a Volume
+func DeleteVolumeVolfiles(volname string) error {
+	resp, e := store.Get(context.TODO(), volfilePrefix, clientv3.WithPrefix())
+	if e != nil {
+		return e
+	}
+
```
prashanthpai: Suggestion: If all volfiles belonging to a volume are under one prefix - `volfiles/<volname>`, you can delete all with one op usingWithPrefix()`

glusterd2/volgen/store-utils.go
```diff
@@ -0,0 +1,91 @@
...
+		volFile := strings.TrimPrefix(string(kv.Key), "volfiles/")
+		volfiles[i] = volFile
+	}
+
+	return volfiles, nil
+}
+
+//GetVolfile return particular volfile info
+func GetVolfile(volfileID string) ([]byte, error) {
+	volfile := volfilePrefix + volfileID
```
prashanthpai: nit: Please use `err` as variable name if possible. I understand that this code may have come from somewhere else where it's used as `e`

glusterd2/volgen/store-utils.go
aravindavk: good idea, rebalance(`rebalance/<volname>`) and gfproxy volfiles(`gfproxy-client/<volname>` and `gfproxyd/<volname>`) names are not in this format. I will change those daemons as well to use `<volname>...` as volfile id.

glusterd2/volgen/generate.go
aravindavk: This change is not necessary if I add `<volname>` prefix to all volfiles.

glusterd2/volgen/store-utils.go
```diff
@@ -0,0 +1,91 @@
...
+
+type volfileData struct {
+	VolumeName string
+	Content    string
+}
+
+func save(name string, content string, volname string) error {
+	jsondata, err := json.Marshal(volfileData{VolumeName: volname, Content: content})
+	if err != nil {
+		log.WithField("error", err).Error("Failed to marshal the volfile data")
```
rishubhjain: You can log the error using WithError(). 
log.WithError(err)...

glusterd2/volgen/store-utils.go
rishubhjain: @prashanthpai I think using short names is better if the variable is used very close to its declaration/initialization. Just read it https://talks.golang.org/2014/names.slide#4. If its a convention in our source code then its better to use error.

aravindavk:
@prashanthpai updated the code as suggested. All Volume volfiles are renamed to have volume name as prefix and deleting volfiles based on prefix `/volfiles/<volname>` or `/volfiles/snaps/<volname>`

glusterd2/volgen/store-utils.go
```diff
@@ -48,3 +48,35 @@ func GetVolfile(volfileID string) ([]byte, error) {
 	}
 	return resp.Kvs[0].Value, nil
 }
+
+// DeleteVolumeVolfiles deletes all the Volfiles belonging to a Volume
+func DeleteVolumeVolfiles(volname string) error {
+	resp, err := store.Get(context.TODO(), volfilePrefix+volname, clientv3.WithPrefix())
```
kshlm: You could directly do a `store.Delete(ctx, prefix, clientv3.WithPrefix())` to delete everything under the prefix. No need to fetch each entry individually. Same for the snaps volfiles below.

Tie both together using etcd transaction to ensure all happen with a single call.
```go
resp, err = store.Store.Txn(ctx).If().
  Then(
    clientv3.OpDel(volprefix, clientvt.WithPrefix()),
    clientv3.OpDel(volpreifx),
    clientv3.OpDel(snapPrefix, clientv3.WithPrefix()),
    clientv3.OpDel(snapPrefix)
  ).Else().Commit()
```

This is what I believe @prashanthpai commented about before.

glusterd2/volgen/store-utils.go
prashanthpai: I was hinting at just client and brick volfiles (per volume ones). The cluster volfiles aren't on per volume basis right ? It'll be global, isn't it ?

glusterd2/volgen/store-utils.go
aravindavk: good one. thanks for the suggestion.

glusterd2/volgen/store-utils.go
aravindavk: yes. cluster volfiles are prefixed as `volfiles/gluster/`. Rebalance and gfproxy are also volume level volfiles(one per volume)

aravindavk:
@kshlm updated as suggested.

glusterd2/volgen/store-utils.go
```diff
@@ -48,3 +48,14 @@ func GetVolfile(volfileID string) ([]byte, error) {
 	}
 	return resp.Kvs[0].Value, nil
 }
+
+// DeleteVolumeVolfiles deletes all the Volfiles belonging to a Volume
+func DeleteVolumeVolfiles(volname string) error {
+	_, err := store.Store.Txn(context.TODO()).If().
+		Then(
+			clientv3.OpDelete(volfilePrefix+volname, clientv3.WithPrefix()),
+			clientv3.OpDelete(volfilePrefix+"snaps/"+volname, clientv3.WithPrefix()),
```
prashanthpai: This code is okay but just a question: Volumes cannot be deleted without deleting snapshots first. If that's the case, snapshot volfiles should be deleted when snapshot itself is deleted, right ?

aravindavk: yeah, you are right. I will remove `snaps/<volname>` from here.

aravindavk:
retest this please

glusterd2/volgen2/generate.go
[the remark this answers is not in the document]
rafikc30: Yea, Once we modify volfile regenerate code for volume stop, then we don't need to worry about deleting the cluster level volfiles. We have to make sure that we do the modification for the last volume correctly, either we can delete the volfile here, or we can keep the first xlator.

rafikc30:
There is a problem with snapshot brick volfile delete, I will send a PR to fix the issue #1013 .

PR #952, opened 2018-07-03
Extract master volume ID and remote volume ID from GeorepStatus API

Madhu-1:
currently, we are  making separate calls to fetch
remote volume ID and master volume ID,

call GeorepStatus API to get the list of geo-replication
sessions and fetch master volume ID and remote volume
ID from the response


Madhu-1:
@aravindavk , @prashanthpai  please review

Issue #606, opened 2018-03-20
Enable REST API authentication by default


aravindavk:
- [ ] REST API authentication to be enabled by default.
- [ ] Modify tests to use REST API Auth
- [ ] New argument to CLI to accept Auth file path(For local usage, to run Gluster CLI from one of the node of Cluster).
- [ ] Protect Auth file to "gluster" group only(Read permission)

Note: User management yet to be implemented, Using Gluster CLI from outside the Cluster is not yet possible.

PR #960, opened 2018-07-04
 Enable Rest Authentication by default

Madhu-1:
* Enable Rest authentication in glusterd2 by default
* updated test cases to use auth by default
* Protect Auth file to "gluster" group only

Fixes: #606 

### DIFF
--- a/e2e/config/3.toml
+++ b/e2e/config/3.toml
@@ -7,3 +7,5 @@ peeraddress = "127.0.0.1:22008"
 clientaddress = "127.0.0.1:22007"
 etcdcurls = "http://127.0.0.1:2279"
 etcdpurls = "http://127.0.0.1:2280"
+#restauth should be set to false to disable REST authentication in glusterd2
+restauth = true
--- a/e2e/utils_test.go
+++ b/e2e/utils_test.go
@@ -27,6 +27,7 @@ type gdProcess struct {
 	ClientAddress string `toml:"clientaddress"`
 	PeerAddress   string `toml:"peeraddress"`
 	LocalStateDir string `toml:"localstatedir"`
+	RestAuth      bool   `toml:"restauth"`
 	Rundir        string `toml:"rundir"`
 	uuid          string
 }
@@ -191,7 +192,7 @@ func setupCluster(configFiles ...string) ([]*gdProcess, error) {
 	}
 
 	// restclient instance that will be used for peer operations
-	client := initRestclient(gds[0].ClientAddress)
+	client := initRestclient(gds[0].ClientAddress, gds[0].LocalStateDir)
 
 	// first gd2 instance spawned shall add other glusterd2 instances as its peers
 	for i, gd := range gds {
@@ -237,8 +238,9 @@ func teardownCluster(gds []*gdProcess) error {
 	return nil
 }
 
-func initRestclient(clientAddress string) *restclient.Client {
-	return restclient.New("http://"+clientAddress, "", "", "", false)
+func initRestclient(clientAddress, authPath string) *restclient.Client {
+	secret := getAuth(authPath)
+	return restclient.New("http://"+clientAddress, "glustercli", secret, "", false)
 }
 
 func prepareLoopDevice(devname, loopnum, size string) error {
@@ -392,4 +394,16 @@ func testTempDir(t *testing.T, prefix string) string {
 		panic(err)
 	}
 	return d
+}
+
+func getAuth(path string) string {
+	filepath := path + "/auth"
+	if _, err := os.Stat(filepath); !os.IsNotExist(err) {
+		s, err := ioutil.ReadFile(filepath)
+		if err != nil {
+			panic("unable to read auth file")
+		}
+		return string(s)
+	}
+	return ""
 }
--- a/glustercli/cmd/root.go
+++ b/glustercli/cmd/root.go
@@ -2,9 +2,12 @@ package cmd
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
 
 	"github.com/gluster/glusterd2/pkg/logging"
 
+	log "github.com/sirupsen/logrus"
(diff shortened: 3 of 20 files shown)


Madhu-1:

e2e/config/3.toml
```diff
@@ -7,3 +7,4 @@ peeraddress = "127.0.0.1:22008"
 clientaddress = "127.0.0.1:22007"
 etcdcurls = "http://127.0.0.1:2279"
 etcdpurls = "http://127.0.0.1:2280"
+restauth = true
```
prashanthpai: What's this ?

prashanthpai: Never mind, github is testing some new feature. Ignore the comment.

phlogistonjohn:
I have one general concern about this change, specifically I am concerned that this PR indicates that rest auth is being turned on by default, but the diff makes it appear that the server relies on a "restauth" config parameter being set to true.
I'm OK with the change if the intent is just to do more testing with auth enabled. However, I don't want to say "rest auth is enabled by default" if config file parameter must be set. Rather, if the project advertises authentication is enabled by default it shouldn't require the config parameter to be present (IOW, missing should mean restauth=true) and users should have to do extra work to make it less secure rather than more. I realize that is a bit hair-splitting but I think it is important to be very clear about what we mean when it comes to stuff with a security impact.

Madhu-1:
@phlogistonjohn, the server does not rely on "restauth" key to turn on the authentication, but it does rely on "restauth" key to turn it off.
I have added the key to the config file to make sure the user knows the key to disable the authentication
even if you remove the key from the config file and starts the server, the rest authentication is enabled by default.

glustercli/cmd/root.go
```diff
@@ -17,7 +19,28 @@ var RootCmd = &cobra.Command{
 		if err != nil {
 			fmt.Println("Error initializing log file ", err)
 		}
-		initRESTClient(flagEndpoints[0], flagUser, flagSecret, flagCacert, flagInsecure)
+
+		if flagAuthFile == "" && flagSecret == "" {
+			data, err := ioutil.ReadFile(defaultAuthPath)
+			if err != nil && !os.IsNotExist(err) {
+				failure("failed to read secret", err, 1)
```
aravindavk: Do not fail here(you can add verbose log), try without secret since CLI doesn't know that restauth is enabled or not. set secret to empty string.

aravindavk: This can also happen when glustercli is run from outside the cluster(Example: Run glustercli from laptop to manage remote Gluster clusters)

glusterd2/gdctx/global.go
```diff
@@ -73,6 +75,22 @@ func GenerateLocalAuthToken() error {
 			if errWrite := ioutil.WriteFile(authFile, []byte(LocalAuthToken), 0640); errWrite != nil {
 				return errWrite
 			}
+			usr, err := user.Lookup("gluster")
```
aravindavk: Look for "gluster" group not "gluster" user.

glusterd2/gdctx/global.go
```diff
@@ -73,6 +75,22 @@ func GenerateLocalAuthToken() error {
...
+			}
+			uID, err := strconv.Atoi(usr.Uid)
+			if err != nil {
+				return err
+			}
+			gID, err := strconv.Atoi(usr.Gid)
+			if err != nil {
+				return err
+			}
+			err = os.Chown(authFile, uID, gID)
```
aravindavk: set `uID` of `root` user itself.

glustercli/cmd/root.go
Madhu-1: done

glusterd2/gdctx/global.go
Madhu-1: done

glusterd2/gdctx/global.go
Madhu-1: setting ```uID``` of the current user.

glustercli/cmd/root.go
```diff
@@ -17,7 +20,32 @@ var RootCmd = &cobra.Command{
...
+				}
+			}
+			secret = string(data)
+		}
+
+		if flagAuthFile != "" {
+			data, err := ioutil.ReadFile(flagAuthFile)
+			if err != nil {
+				if verbose {
+					log.WithError(err).Error("failed to read secret")
```
aravindavk: Retain failure here since user passed authFile with invalid path.

Madhu-1: done

Madhu-1:
if ```gluster``` group doesn't exist, glusterd2 will fail with the following error

```group: unknown group gluster```

if the group doesn't exist, do I need to disable ```authfile``` protection, 
or glusterd2 should log error and exit?

**Comment by aravindavk** [comment]
> if gluster group doesn't exist, glusterd2 will fail with the following error

glusterd2 only logs the error don't exit. I think by default user and group is root right? if yes then we are good wrt security. If user is created after glusterd2 service start then auth file permission needs to be set manually  as required or glusterd2 needs restart.

>  should log error and exit?

Log error is fine. But we need answer to this issue in future.  May be we can document it CLI documentation in future.

File: e2e/utils_test.go [hunk 6]
```diff
@@ -191,7 +192,7 @@ func setupCluster(configFiles ...string) ([]*gdProcess, error) {
 	}
 
 	// restclient instance that will be used for peer operations
-	client := initRestclient(gds[0].ClientAddress)
+	client := initRestclient(gds[0].ClientAddress, gds[0].LocalStateDir)
```
prashanthpai: As a cleaner way, you could've passed `*gdProcess` to `initRestclient`.

Madhu-1: as this is merged I will send a patch to pass ```*gdProcess``` to ```initRestclient```

PR #961, opened 2018-07-05
Setup Jaegertracing for volume create command.

### LINKED ISSUE #941: GD2 Tracing: Add Opencensus/Jaeger infrastructure for tracing GD2 management commands.
Add the necessary infrastructure to enable tracing using opencensus and Jaeger.

sseshasa:
This change sets up Jaeger tracing for volume create path. This initial changeset only traces calls locally and not across rpc boundaries. Therefore, the traces are only visible on the node which handled the volume create request.

None of the code changes changes the original functionality and therefore UTs and functional tests are not added. But all existing UTs, functional tests and testing on VM with 1 node succeeded.

github issue #941 tracks this change.

Closes #941

### DIFF
--- a/glusterd2/transaction/step.go
+++ b/glusterd2/transaction/step.go
@@ -1,6 +1,7 @@
 package transaction
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net/http"
@@ -10,6 +11,7 @@ import (
 
 	"github.com/pborman/uuid"
 	log "github.com/sirupsen/logrus"
+	"go.opencensus.io/trace"
 )
 
 // StepFunc is the function that is supposed to be run during a transaction step
@@ -33,14 +35,14 @@ var (
 )
 
 // do runs the DoFunc on the nodes
-func (s *Step) do(ctx TxnCtx) error {
-	return runStepFuncOnNodes(s.DoFunc, ctx, s.Nodes)
+func (s *Step) do(origCtx context.Context, ctx TxnCtx) error {
+	return runStepFuncOnNodes(origCtx, s.DoFunc, ctx, s.Nodes)
 }
 
 // undo runs the UndoFunc on the nodes
 func (s *Step) undo(ctx TxnCtx) error {
 	if s.UndoFunc != "" {
-		return runStepFuncOnNodes(s.UndoFunc, ctx, s.Nodes)
+		return runStepFuncOnNodes(nil, s.UndoFunc, ctx, s.Nodes)
 	}
 	return nil
 }
@@ -88,13 +90,13 @@ func (r stepResp) Status() int {
 	return http.StatusInternalServerError
 }
 
-func runStepFuncOnNodes(stepName string, ctx TxnCtx, nodes []uuid.UUID) error {
+func runStepFuncOnNodes(origCtx context.Context, stepName string, ctx TxnCtx, nodes []uuid.UUID) error {
 
 	respCh := make(chan stepPeerResp, len(nodes))
 	defer close(respCh)
 
 	for _, node := range nodes {
-		go runStepFuncOnNode(stepName, ctx, node, respCh)
+		go runStepFuncOnNode(origCtx, stepName, ctx, node, respCh)
 	}
 
 	// Ideally, we have to cancel the pending go-routines on first error
@@ -126,7 +128,7 @@ func runStepFuncOnNodes(stepName string, ctx TxnCtx, nodes []uuid.UUID) error {
 	return nil
 }
 
-func runStepFuncOnNode(stepName string, ctx TxnCtx, node uuid.UUID, respCh chan<- stepPeerResp) {
+func runStepFuncOnNode(origCtx context.Context, stepName string, ctx TxnCtx, node uuid.UUID, respCh chan<- stepPeerResp) {
 
 	ctx.Logger().WithFields(log.Fields{
 		"step": stepName, "node": node,
@@ -134,7 +136,7 @@ func runStepFuncOnNode(stepName string, ctx TxnCtx, node uuid.UUID, respCh chan<
 
 	var err error
 	if uuid.Equal(node, gdctx.MyUUID) {
-		err = runStepFuncLocally(stepName, ctx)
+		err = runStepFuncLocally(origCtx, stepName, ctx)
 	} else {
 		// remote node
 		err = runStepOn(stepName, node, ctx)
@@ -143,9 +145,16 @@ func runStepFuncOnNode(stepName string, ctx TxnCtx, node uuid.UUID, respCh chan<
 	respCh <- stepPeerResp{node, err}
 }
 
-func runStepFuncLocally(stepName string, ctx TxnCtx) error {
+func runStepFuncLocally(origCtx context.Context, stepName string, ctx TxnCtx) error {
 
 	var err error
+
+	if origCtx != nil {
+		reqID := ctx.GetTxnReqID()
+		spanName := stepName + " ReqID:" + reqID
+		_, span := trace.StartSpan(origCtx, spanName)
+		defer span.End()
+	}
 
 	stepFunc, ok := getStepFunc(stepName)
 	if ok {
--- a/glusterd2/transaction/transaction.go
+++ b/glusterd2/transaction/transaction.go
@@ -132,7 +132,7 @@ func (t *Txn) Do() error {
 			continue
 		}
 
-		if err := s.do(t.Ctx); err != nil {
+		if err := s.do(t.OrigCtx, t.Ctx); err != nil {
 			expTxn.Add("initiated_txn_failure", 1)
 			if !t.DisableRollback {
 				t.Ctx.Logger().WithError(err).Error("Transaction failed, rolling back changes")
--- a/glusterd2/commands/volumes/volume-create.go
+++ b/glusterd2/commands/volumes/volume-create.go
@@ -16,6 +16,7 @@ import (
 	gderrors "github.com/gluster/glusterd2/pkg/errors"
 
 	"github.com/pborman/uuid"
+	"go.opencensus.io/trace"
 )
 
 const (
@@ -99,6 +100,9 @@ func registerVolCreateStepFuncs() {
 func volumeCreateHandler(w http.ResponseWriter, r *http.Request) {
 
 	ctx := r.Context()
+	ctx, span := trace.StartSpan(ctx, "/volumeCreateHandler")
+	defer span.End()
+
 	logger := gdctx.GetReqLogger(ctx)
 	var err error
 


sseshasa:
This is a new PR for tracing changes as the git history of the old PR got messed up.

Request @aravindavk @prashanthpai @kshlm @rishubhjain to review this and add other folks as well.

glusterd2/main.go
```diff
@@ -119,6 +121,29 @@ func main() {
...
+
+	// Create the Opencensus Jaeger exporter endpoints to log traces and spans
+	exporter, err := jaeger.NewExporter(jaeger.Options{
+		Endpoint:      jaegerEndpoint,
+		AgentEndpoint: jaegerAgentEndpoint,
+		ServiceName:   gdctx.HostName,
+	})
+	if err != nil {
+		log.WithFields(log.Fields{
+			"error": err.Error(),
```
rishubhjain: IMO: field should include the component name, or the object details etc, Error msg must be logged as an error message or a warning message.

rishubhjain: @prashanthpai Please suggest the best way to log this.

glusterd2/transaction/context.go
```diff
@@ -165,6 +167,12 @@ func (c *Tctx) GetNodeResult(peerID uuid.UUID, key string, value interface{}) er
 	return c.Get(storeKey, value)
 }
 
+// GetTxnReqID gets the reqID string saved within the txnCtxConfig.
+func (c *Tctx) GetTxnReqID() string {
+	logFields := c.config.LogFields
```
rishubhjain: TRIVIAL: Is this temporary variable really needed

glusterd2/main.go
rishubhjain: @sseshasa IMO: the field should have a component name, object name etc, to make debugging easy,   error message must be provided in the WithError() or Warning section.

rishubhjain: @prashanthpai Can you suggest the best way to log error here.

glusterd2/main.go
```diff
@@ -119,6 +121,29 @@ func main() {
 		log.WithError(err).Fatal("Failed to generate local auth token")
 	}
 
+	// Get the Jaeger endpoints from config file
```
kshlm: Please move all the jaeger initialization into a seperate package or a seperate file in the main package if you feel creating a new package is a bit much.
I'd prefer a seperate package, so that you could reuse the configuration and intialization for glustercli as well.

glusterd2/transaction/transaction.go
```diff
@@ -125,12 +127,14 @@ func (t *Txn) Do() error {
 		return err
 	}
 
+	var origCtx []context.Context
```
kshlm: You can do this in one step by `origCtx := []context.Context{t.OrigCtx}`

glusterd2/main.go
kshlm: The consistent way to log errors is to use `log.WithError(err)`. Chain it with `WithFields()` for any other context that needs to be added.

glusterd2/main.go
```diff
@@ -119,6 +121,29 @@ func main() {
...
+		log.WithFields(log.Fields{
+			"error": err.Error(),
+		}).Warning("Unable to create jaeger exporter")
+	} else {
+		defer exporter.Flush()
+		// Register the Jaeger exporter
+		trace.RegisterExporter(exporter)
+		// TODO: Change to probability sampler. Use always sample for now.
+		trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
+		log.Info("Registered opencensus exporter for traces and stats")
```
kshlm: Please log the jaeger endpoints that you've registered with as context with this log.

glusterd2/main.go
```diff
@@ -119,6 +121,29 @@ func main() {
 		log.WithError(err).Fatal("Failed to generate local auth token")
 	}
 
+	// Get the Jaeger endpoints from config file
+	jaegerEndpoint := config.GetString("jaegerendpoint")
+	jaegerAgentEndpoint := config.GetString("jaegeragentendpoint")
```
kshlm: With just this, the endpoints can only be fetched from the config (toml) file. We prefer that all configurable options be configurable as CLI options as well. Please refer to how the store package initalizes its flags and config (refer `store.InitFlags`).

glusterd2/main.go
```diff
@@ -119,6 +121,29 @@ func main() {
...
+
+	// Create the Opencensus Jaeger exporter endpoints to log traces and spans
+	exporter, err := jaeger.NewExporter(jaeger.Options{
```
kshlm: Have this be a function `InitTracing` that returns an exporter. The GD2 main function will call this function, and then call `defer exporter.Flush()` if a non-nil return happens.

glusterd2/transaction/step.go
```diff
@@ -143,10 +145,18 @@ func runStepFuncOnNode(stepName string, ctx TxnCtx, node uuid.UUID, respCh chan<
 	respCh <- stepPeerResp{node, err}
 }
 
-func runStepFuncLocally(stepName string, ctx TxnCtx) error {
+func runStepFuncLocally(stepName string, ctx TxnCtx, origCtx ...context.Context) error {
```
kshlm: Do you expect a case in the future where you may have multiple origCtx's? If not, then it's better that this (and the other occurances above) be simple function rather than variadic functions (...origCtx -> origCtx).

glusterd2/main.go
sseshasa: I will use log.WithError() and chain it with WithFields() as suggested. Please check latest evidence.

glusterd2/transaction/context.go
sseshasa: I will remove the temp variable.

glusterd2/main.go
sseshasa: Agreed. I have moved the jaeger initialization to a separate package called 'tracing' within glusterd2/pkg. Please check latest evidence.

glusterd2/transaction/transaction.go
sseshasa: With the latest change, this change is no longer necessary.

glusterd2/main.go
sseshasa: Done. Please check latest evidence.

glusterd2/main.go
sseshasa: I have introduced tracing.InitFlags() to initialize the jaeger endpoints using CLI as suggested.

glusterd2/main.go
sseshasa: The exporter initialization is moved to tracing.InitJaegerExporter() which is specific to Jaeger. If we support tracing using some other mechanism/endpoint(s) in the future, a new function can be defined as needed.

glusterd2/transaction/step.go
sseshasa: No, I don't expect multiple origCtx's. I have converted all the functions to simple functions.

My initial thought was to make the functions variadic so that this parameter need not be specified at all if tracing is not required for certain transactions. But looking closer, we needn't make them variadic. If the context parameter is nil, the transaction and associated step function(s) won't be traced.

glusterd2/main.go
```diff
@@ -119,6 +120,36 @@ func main() {
...
+	if exporter != nil {
+		defer exporter.Flush()
+		jaegerEndpoints[tracing.JaegerEndpointOpt] = jaegerEndpoint
+		jaegerEndpoints[tracing.JaegerAgentEndpointOpt] = jaegerAgentEndpoint
+		// Register the Jaeger exporter
+		tracing.RegisterJaegerExporter(exporter, jaegerEndpoints)
+	}
+
+	// Get the Jaeger endpoints from config file if provided
+	jaegerEndpoint2 := config.GetString(tracing.JaegerEndpointOpt)
```
kshlm: Just this second initialization from the values obtained from `config` package is enough. `config.Get*` functions fetch configured with the following priority, CLI > Config file > default values. You do not need to do a seperate fetch.

glusterd2/main.go
```diff
@@ -119,6 +120,36 @@ func main() {
...
+	jaegerAgentEndpoint2 := config.GetString(tracing.JaegerAgentEndpointOpt)
+	// Create the Opencensus Jaeger exporter with info from config file
+	if jaegerEndpoint != jaegerEndpoint2 || jaegerAgentEndpoint != jaegerAgentEndpoint2 {
+		exporter := tracing.InitJaegerExporter(jaegerEndpoint2, jaegerAgentEndpoint2)
```
kshlm: You could actually do all this within the tracing package. The `InitJaegerExporter` function can fetch the configured options and if required create the exporter and register the endpoints. It would then return just the exporter (or nil) so it could be closed when GD2 exits.

This way much of the control over tracing initialization is within the tracing package, and the main package and function remain simple.

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,57 @@
...
+
+	log "github.com/sirupsen/logrus"
+	flag "github.com/spf13/pflag"
+	"go.opencensus.io/exporter/jaeger"
+	"go.opencensus.io/trace"
+)
+
+// Commandline options for exporter endpoints
+const (
+	JaegerEndpointOpt      = "jaegerendpoint"
```
kshlm: Would be better to use '-' as seperator ("jaeger-endpoint" "jaegar-agent-endpoint"). It would make the options much more readable.

Also, with the suggested changes, you don't need to export these constants out of this package.

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,57 @@
...
+	flag.String(JaegerEndpointOpt, "", "Jaeger collector endpoint that accepts spans from Jaeger agent.")
+	flag.String(JaegerAgentEndpointOpt, "", "Jaeger agent endpoint that the Jaeger client sends spans to.")
+}
+
+// InitJaegerExporter initializes the jaeger exporter as the tracing endpoint
+// This should be called early when a process starts.
+// This creates and returns an exporter if successful.
+// Otherwise, a warning is logged and a 'nil' exporter is returned.
+func InitJaegerExporter(jaegerEndpoint string, jaegerAgentEndpoint string) *jaeger.Exporter {
+	// Create the Opencensus Jaeger exporter
```
kshlm: As mentioned in my earlier comment, you can fetch the configured options here directly, and you need to do it once. Just  `import config "github.com/spf13/viper` and use the config package.

Create and register the exporter only if both options have been set (`!= ""). Return a nil otherwise.

glusterd2/main.go
sseshasa: This makes things much simpler! Thanks for pointing this out. I saw the way logging related options were being reinitialized and followed it.

pkg/tracing/tracing.go
sseshasa: Sure, I have changed the options as suggested and made the constants private to this package.

glusterd2/main.go
kshlm: Logging is a special case because. We need logging to be available before we read-in the config file.

sseshasa:
Closes #941

PR #947, opened 2018-07-01
e2e: establish a common approach to creating per-test temp dirs

phlogistonjohn:
Add a helper function for tests that need unique test dirs that
creates test directories like /[base]/[testname]/[unique].

Fixes #927

Signed-off-by: John Mulligan <jmulligan@redhat.com>

PR #959, opened 2018-07-04
e2e: Ensure setupCluster() returns error on failure

prashanthpai:
Also:
* Make setupCluster() use the restclient to do peer operations
* Bump up default wait time to check if gd2 is alive from 1.5s to 2s
* Ensure cleanup happens in all failure cases

Closes #956 
Signed-off-by: Prashanth Pai <ppai@redhat.com>

PR #970, opened 2018-07-07
e2e: use an external etcd during tests

### LINKED ISSUE #930: CI: Deploy external etcd for e2e tests
As the first and imminent deployment of glusterd2 will be in a K8s environment, glusterd2 shall use an external etcd cluster managed by K8s etcd operator. To focus on testing that focused deployment, I propose we change our e2e tests running on CI test runs to use an external etcd cluster. This external etcd cluster can be a single node cluster for testing purposes.

phlogistonjohn:
This series of changes adds the ability to run the glusterd2 servers under test with or without the embedded etcd enabled. The default is now to run without embedded etcd (aka external etcd). When embedded etcd is disabled the test framework will start a single etcd instance for the tests.

This behavior can be controlled with a new -external-etcd=[true|false] option. When using the external etcd mode a etcd binary must be present on your path. No other setup is required

As part of this change a number of paths in the code needed to be refactored, including some of the sub-test function arguments. I'm not 100% happy with the extra length added by the inline adapter functions, but at least it avoids a sneaky global.

This change is built upon the changes in pr #947 to avoid conflicting with myself. Please review and merge that change first. 

Fixes #930

### DIFF
--- a/glusterd2/xlator/load.go
+++ b/glusterd2/xlator/load.go
@@ -89,7 +89,6 @@ func loadXlator(xlPath string) (*Xlator, error) {
 	defer C.dlclose(handle)
 
 	xl := new(Xlator)
-	xl.ID = strings.TrimSuffix(filepath.Base(xlPath), filepath.Ext(xlPath))
 
 	xlSym := C.CString("xlator_api")
 	defer C.free(unsafe.Pointer(xlSym))
@@ -97,9 +96,7 @@ func loadXlator(xlPath string) (*Xlator, error) {
 	p := C.dlsym(handle, xlSym)
 	if p != nil {
 		xp := (*C.xlator_api_t)(p)
-		// FIXME: It's named "server-protocol" instead of "server" in server.so
-		//        https://review.gluster.org/18879
-		// xl.ID = C.GoString(xp.identifier)
+		xl.ID = C.GoString(xp.identifier)
 		xl.rawID = uint32(xp.xlator_id)
 		xl.Flags = uint32(xp.flags)
 		for _, k := range xp.op_version {
@@ -113,6 +110,14 @@ func loadXlator(xlPath string) (*Xlator, error) {
 		if p == nil {
 			return xl, nil
 		}
+	}
+
+	if xl.ID == "" {
+		// The xlator ID defaults to name of its .so file unless the
+		// xlator defines 'xlator_api_t' structure which has the
+		// 'identifier' field.
+		xl.ID = strings.TrimSuffix(filepath.Base(xlPath),
+			filepath.Ext(xlPath))
 	}
 
 	soOptions := (*[maxOptions]C.volume_option_t)(p)


prashanthpai:
> This change is built upon the changes in pr #947 to avoid conflicting with myself. Please review and merge that change first.

#947 has been merged. I'd also prefer to get #959 in before this so that you can fix all merge conflicts at once. I haven't looked at this PR yet.

Some arbitrary sleeps were added to peer tests as a workaround for elasticetcd issues.

```sh
$ grep "Sleep" e2e/peer_ops_test.go 
	time.Sleep(6 * time.Second)
	time.Sleep(6 * time.Second)
```

I believe we no longer need those when we use external etcd. Please see if the tests consistently pass with those sleeps removed. Users should be able to expect a healthy and fully operable cluster as soon as a peer operation (add or remove) returns a success response.

phlogistonjohn:
> I believe we no longer need those when we use external etcd. Please see if the tests consistently pass with those sleeps removed. Users should be able to expect a healthy and fully operable cluster as soon as a peer operation (add or remove) returns a success response.

Hopefully, you are right. However, I'd like to tackle this in a follow up pr. For now I'd just like to get the infrastructure in place. Later, we can start making changes based on that infrastructure. (Esp. since rebasing this has been quite a handful so far)

prashanthpai:
@phlogistonjohn A general comment on the commit style: It's making my job of reviewing this PR quite hard. For a complete review, I have to look at `Files changed` **and** also each commit independently which is 12 commits with their commit messages! To do that, I have to form some kind of stack mentally to digest it which is additional cognitive burden. This additional burden is not even a trade-off for something good. I'd like to hear your's (and others) opinion on the purpose of this.

I can see that the commits serve as a history of how a contributor has broken the problem into multiple chunks and how he/she has arrived at the final solution over multiple days or iterations. When this is merged, in future, I'd prefer to see a single commit via `git show <commit-id>` to figure out what it took to support external etcd in tests. I do not prefer to be forced to go to github to see that.

I don't have a problem with multiple commits as long as they aren't related/coupled. A general rule of thumb to follow is that - multiple commits are okay if it also makes sense be sent as separate PRs.

I'll try to illustrate this: One commit adds `etcd` as a dependent package, another commit adds a `--external-etcd` flag, another commit adds code to spawn etcd process. All these 3 commits solve one problem **only collectively**. If any one of the commit is reverted or not present, it won't work. It doesn't make sense for these 3 commits to be sent as 3 PRs. 

How do you guys review PRs in heketi ? Do you only look at `Files changed` tab and ignore reviewing the commit messages or do you review both ?

e2e/bitrot_test.go
```diff
@@ -14,20 +14,20 @@ func TestBitrot(t *testing.T) {
...
 	r.Nil(err)
-	defer teardownCluster(gds)
+	defer teardownCluster(tc)
 
-	client = initRestclient(gds[0])
+	client = initRestclient(tc.gds[0])
 
 	// test Bitrot on dist-rep volume
-	t.Run("Replica-volume", testBitrotOnReplicaVolume)
+	t.Run("Replica-volume", func(t *testing.T) { testBitrotOnReplicaVolume(t, tc) })
```
prashanthpai: Suggestion: you can write a wrapper function if this pattern repeats. In pseudo-Go code:

```go
wrapTest(t, testBitrotOnReplicaVolume, tc)
```

e2e/peer_ops_test.go
```diff
@@ -12,6 +12,10 @@ import (
 func TestAddRemovePeer(t *testing.T) {
 	r := require.New(t)
 
+	tc, err := setupCluster()
```
prashanthpai: This is slightly misleading. Why is this required ?

e2e/process_utils_test.go
```diff
@@ -0,0 +1,224 @@
...
+		err  error
+	)
+	if ep.LogPath != "" {
+		logf, err = os.OpenFile(ep.LogPath,
+			os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			return err
+		}
+		ep.Cmd.Stdout = logf
+		defer logf.Close()
```
prashanthpai: Wouldn't this close the file ? Should this statement be moved to the goroutine spawned below which waits for the process to die ?

e2e/utils_test.go
```diff
@@ -14,188 +12,62 @@ import (
...
+		}
+		if err := os.MkdirAll(tc.etcd.DataDir, 0755); err != nil {
+			return nil, err
+		}
+		err := tc.etcd.Spawn()
+		if err != nil {
+			return nil, err
+		}
+	}
+	// exit the function early if not gd2 instances were requested
```
prashanthpai: `no gd2 instances`

e2e/peer_ops_test.go
prashanthpai: I got the `why` part. This should be okay.

e2e/bitrot_test.go
phlogistonjohn: I had thought about that and decided to defer doing that. Since you've suggested it I will add it.

e2e/process_utils_test.go
phlogistonjohn: :-) I'm glad you brought that up, I'll add a comment to clarify. In brief, once the process has been spawned the parent process no longer needs the fd and can close it. defer was used such that the file is closed on all paths.

e2e/utils_test.go
phlogistonjohn: will fix

phlogistonjohn:
> A general comment on the commit style: It's making my job of reviewing this PR quite hard. For a complete review, I have to look at Files changed and also each commit independently which is 12 commits with their commit messages! To do that, I have to form some kind of stack mentally to digest it which is additional cognitive burden. This additional burden is not even a trade-off for something good. I'd like to hear your's (and others) opinion on the purpose of this style

> I can see that the commits serve as a history of how a contributor has broken the problem into multiple chunks and how he/she has arrived at the final solution over multiple days or iterations. How useful is that to reviewers or users of git commands such as git log, git show, git blame ? When this is merged, in future, I'd prefer to see a single commit via git show <commit-id> to figure out what it took to support external etcd in tests. I do not prefer to be forced to go to github to see that.

If you wish me to squash patches together I will do so. Am I correct that you prefer to see all patches that create the support for external etcd squashed? Does that include the preparatory refactoring?

> I don't have a problem with multiple commits as long as they aren't related/coupled. A general rule of thumb to follow is that - multiple commits are okay if it also makes sense be sent as separate PRs.

> I'll try to illustrate this: One commit adds etcd as a dependent package, another commit adds a --external-etcd flag, another commit adds code to spawn etcd process. All these 3 commits solve one problem only collectively. If any one of the commit is reverted or not present, it won't work. It doesn't make sense for these 3 commits to be sent as 3 PRs.

> How do you guys review PRs in heketi ? Do you only look at Files changed tab and ignore reviewing the commits (ordering, title and messages, content) or do you review both ?

I always review the commits (commit by commit) which is actually why I prefer an approach like this. I believe you have had previous discussions with @obnoxxx on this topic and I believe my position is much closer to his. I think he also reviews the patches change-by-change but in additon I don't think he reviews the changes using github itself (github has the big downside of not retaining patch order!). In my mind, we both generally seem to prefer small digestible commits that build up to a more comprehensive whole. I see it like writing helper functions that are dependencies for a higher level function, you build up a set of tools and then make use of it.

In this particular case I'd disagree with squashing the dependency for etcd in the centos ci setup script with the golang changes, as the script is not a hard dependency for the e2e tests themselves but rather only part of the centos ci environment.

e2e/process_utils_test.go
prashanthpai: Indeed!

prashanthpai:
> Am I correct that you prefer to see all patches that create the support for external etcd squashed ?

Yes.

> Does that include the preparatory refactoring?

It's okay if that's a separate commit as long as it doesn't implement parts of "support for external etcd" and something that can be potentially sent as a separate PR.

> I'd disagree with squashing the dependency for etcd in the centos ci setup script with the golang changes, as the script is not a hard dependency for the e2e tests themselves but rather only part of the centos ci environment.

I'm okay with that.

> I see it like writing helper functions that are dependencies for a higher level function, you build up a set of tools and then make use of it.

From code contributor or author's perspective solving a problem, I understand that. As a reviewer and a maintainer, I don't like to see that end up in git log. As a personal preference, I like Gerrit as it  makes things so much easier for everyone - better reviews, cleaner git log, preserves change history and comments etc.

phlogistonjohn:
All right, I'll try to have something that takes in the feedback in place later today.

As for contribution/commit styles, I will aim to not make life too difficult for you where we disagree on style. But, we'll have to agree to disagree for now.... We'll just have to try and do a better job of brainwashing you on this topic in the future. ;-)

phlogistonjohn:
Updates pushed.

e2e/process_utils_test.go
```diff
@@ -0,0 +1,227 @@
...
+
+	return true
+}
+
+// Stop will terminate the associated process. It will attempt a graceful
+// shutdown before killing the process.
+func (tp *testProcess) Stop() error {
+	tp.Cmd.Process.Signal(os.Interrupt) // try shutting down gracefully
+	time.Sleep(500 * time.Millisecond)
+	if tp.IsRunning() {
```
Madhu-1: suggestion:

can we simplify this

```
if tp.IsRunning() {
	time.Sleep(1 * time.Second)
        return tp.Cmd.Process.Kill()
        } 
return nil
	
	
```

phlogistonjohn: I don't see how that's simpler. The current code:
* Signals the process to shut down
* Waits a 1/2 second
* Checks to see if process is still alive
* If it is, tries to kill the process

It seems simple enough to me, and it pretty much has to wait before running the IsRunning test, not after.

Madhu-1: I just wanted to remove else case, it won't affect functionality

phlogistonjohn: Oh! I understand you. Your example moved the sleep and would have resulted in a behavior change, but I get you now. Something for another day. :-)

PR #973, opened 2018-07-09
Mounting auto provisioned bricks on gd2 start and unified Volume create API

aravindavk:
- `Commit 1` Merged Smartvol API into regular Volume Create API
- `Commit 2` Mounting auto provisioned bricks on glusterd2 start

Updates: #851


glusterd2/volume/volume-utils.go
```diff
@@ -199,3 +200,54 @@ func CreateVolumeInfoResp(v *Volinfo) *api.VolumeInfo {
...
+		for _, md := range mountdirs[1:] {
+			mounts[md] = struct{}{}
+		}
+	}
+
+	for _, v := range volumes {
+		for _, b := range v.GetLocalBricks() {
+			// Mount all local Bricks if they are auto provisioned
+			if b.MountInfo.DevicePath != "" {
+				if _, exists := mounts[b.MountInfo.Mountdir]; exists {
```
rafikc30: MountInfo.Mountdir actually contains the directory after mount root, for example let's say mount point is /bricks, but brick path could be /bricks/b1 . in this case MountDir will have /b1. 

The name is bit consfusing, I just took the same name from gd1.

glusterd2/volume/volume-utils.go
```diff
@@ -199,3 +200,54 @@ func CreateVolumeInfoResp(v *Volinfo) *api.VolumeInfo {
 		SnapList:  v.SnapList,
 	}
 }
+
+// MountLocalBricks mounts bricks of auto provisioned volumes
+func MountLocalBricks() error {
```
rafikc30: Can we have this function in common utils, so that the same can be used for snapshot also. Otherwise I have to use another function during glusterd2 init.

glusterd2/volume/volume-utils.go
aravindavk: I added without that directory. See Path and Mountdir here https://github.com/gluster/glusterd2/pull/973/files#diff-3e14b73c75a337a5a43f9719dcd68228R131

glusterd2/volume/volume-utils.go
aravindavk: Since it is not generic function it depends on store utils, can't move to common utils. Let me see

aravindavk:
retest this please

aravindavk:
@kshlm @prashanthpai @Madhu-1 @rafikc30  please review

glusterd2/commands/volumes/volume-create.go
```diff
@@ -18,8 +20,20 @@ import (
 
 const (
 	maxMetadataSizeLimit = 4096
+	minVolumeSize        = 10
```
Madhu-1: question:
want to understand why minimum volume size is 10

glusterd2/commands/volumes/volume-create.go
```diff
@@ -29,7 +43,11 @@ func validateVolCreateReq(req *api.VolCreateReq) error {
 		return errors.New("invalid transport. Supported values: tcp or rdma")
 	}
 
-	if len(req.Subvols) <= 0 {
+	if req.Size > 0 && req.Size < minVolumeSize {
+		return errors.New("invalid Volume Size, Minimum size required is " + strconv.Itoa(minVolumeSize))
```
Madhu-1: suggestion: can you convert the capital case to a smaller case in the error message, wherever its required

glusterd2/commands/volumes/volume-create.go
```diff
@@ -98,9 +118,24 @@ func volumeCreateHandler(w http.ResponseWriter, r *http.Request) {
...
 
-	if err := checkDupBrickEntryVolCreate(req); err != nil {
-		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, err)
-		return
+	if req.Size > 0 {
+		applyDefaults(&req)
+
+		if req.SnapshotReserveFactor < 1 {
+			restutils.SendHTTPError(ctx, w, http.StatusBadRequest,
+				errors.New("invalid Snapshot Reserve Factor"))
```
Madhu-1: same here, check any changes required in  error message case

glusterd2/commands/volumes/volume-smartvol-txn.go
```diff
@@ -1,22 +1,19 @@
...
 )
 
 func txnPrepareBricks(c transaction.TxnCtx) error {
-	var req smartvolapi.Volume
-	if err := c.Get("reqBricksPrepare", &req); err != nil {
+	var req api.VolCreateReq
+	if err := c.Get("req", &req); err != nil {
 		c.Logger().WithError(err).WithField(
-			"key", "reqBricksPrepare").Debug("Failed to get key from store")
+			"key", "req").Debug("Failed to get key from store")
```
Madhu-1: question:
This should be Debug or Error message

glusterd2/commands/volumes/volume-smartvol-txn.go
```diff
@@ -61,20 +56,19 @@ func txnPrepareBricks(c transaction.TxnCtx) error {
...
 			}
 
 			// Mount the Created FS
-			err = deviceutils.BrickMount(dev, brickMountDir)
+			err = deviceutils.BrickMount(b.DevicePath, b.Mountdir)
 			if err != nil {
 				c.Logger().WithError(err).
-					WithField("dev", dev).
-					WithField("path", brickMountDir).
+					WithField("dev", b.DevicePath).
```
Madhu-1: question:

which one is best to use ```WithFields``` or ```WithFiled``` in this case

glusterd2/commands/volumes/volume-smartvol-txn.go
```diff
@@ -90,38 +84,17 @@ func txnPrepareBricks(c transaction.TxnCtx) error {
...
 }
 
 func txnUndoPrepareBricks(c transaction.TxnCtx) error {
-	var req smartvolapi.Volume
-	if err := c.Get("reqBricksPrepare", &req); err != nil {
+	var req api.VolCreateReq
+	if err := c.Get("req", &req); err != nil {
 		c.Logger().WithError(err).WithField(
-			"key", "reqBricksPrepare").Debug("Failed to get key from store")
+			"key", "req").Debug("Failed to get key from store")
```
Madhu-1: same here, ```Debug``` or ```Error```

glusterd2/commands/volumes/volume-create.go
aravindavk: I think we may have to increase this upto 20, this is because `mkfs.xfs` was not working on smaller size.

For now this is moved from `plugin/smartvol` code to here.

glusterd2/volume/volume-utils.go
```diff
@@ -199,3 +200,53 @@ func CreateVolumeInfoResp(v *Volinfo) *api.VolumeInfo {
...
+	volumes, e := GetVolumes()
+	if e != nil || volumes == nil {
+		// In case cluster doesn't have any volumes configured yet,
+		// treat this as success
+		log.Debug("Failed to retrieve volumes")
```
Madhu-1: This should be Error on debug log

glusterd2/volume/volume-utils.go
```diff
@@ -199,3 +200,53 @@ func CreateVolumeInfoResp(v *Volinfo) *api.VolumeInfo {
...
+			// Mount all local Bricks if they are auto provisioned
+			if b.MountInfo.DevicePath != "" {
+				if _, exists := mounts[b.MountInfo.Mountdir]; exists {
+					continue
+				}
+
+				err := utils.ExecuteCommandRun("mount", "-o", b.MountInfo.MntOpts, b.MountInfo.DevicePath, b.MountInfo.Mountdir)
+				if err != nil {
+					log.WithError(err).
+						WithField("volume", v.Name).
```
Madhu-1: same here, ```WithFields``` or ```WithField``` is better in this case

glusterd2/commands/volumes/volume-create.go
aravindavk: ok

glusterd2/commands/volumes/volume-smartvol-txn.go
aravindavk: not changed this part of the code while moving from plugins/smartvol. Will update this as required

glusterd2/commands/volumes/volume-smartvol-txn.go
aravindavk: I will change this to WithFields

glusterd2/volume/volume-utils.go
aravindavk: Not error

glusterd2/volume/volume-utils.go
kshlm: You could create a glusterd2/mountutils or just glusterd2/utils instead of pkg/utils

glusterd2/volume/volume-utils.go
```diff
@@ -199,3 +200,53 @@ func CreateVolumeInfoResp(v *Volinfo) *api.VolumeInfo {
 		SnapList:  v.SnapList,
 	}
 }
+
+// MountLocalBricks mounts bricks of auto provisioned volumes
+func MountLocalBricks() error {
+	volumes, e := GetVolumes()
+	if e != nil || volumes == nil {
```
kshlm: This should be 2 unique cases. You can have an error trying to get volumes, and you can get an empty list of volumes. The error is an error, the empty list is not.

pkg/api/volume_req.go
```diff
@@ -27,16 +38,31 @@ type SubvolReq struct {
...
+	DisperseRedundancyCount int               `json:"disperse-redundancy,omitempty"`
+	DisperseDataCount       int               `json:"disperse-data,omitempty"`
+	SnapshotEnabled         bool              `json:"snapshot,omitempty"`
+	SnapshotReserveFactor   float64           `json:"snapshot-reserve-factor,omitempty"`
+	LimitPeers              []string          `json:"limit-peers,omitempty"`
+	LimitZones              []string          `json:"limit-zones,omitempty"`
+	ExcludePeers            []string          `json:"exclude-peers,omitempty"`
+	ExcludeZones            []string          `json:"exclude-zones,omitempty"`
+	SubvolZonesOverlap      bool              `json:"subvolume-zones-overlap,omitempty"`
+	SubvolType              string            `json:"subvolume-type,omitempty"`
```
kshlm: These two structs are now huge. I think it would be a good idea to group all the IVP related members in to their own structs and add those structs as members of the main structs.

aravindavk: I feel input API will not be intuitive if we make it as sub level.

glusterd2/volume/volume-utils.go
rafikc30: While I agree that Mountdir is not an optimal name, storing both path and actual mount dir seems redundant to me. In most case, path will be equal to mount root. And in almost all cases, directories after mount root will be less in length.
So I think we should store path and directory after mount root, and creates mount root using these two.

aravindavk: makes sense, we can make that optimization. That can be separate PR since that needs to be handled in many places including response. `brick.Path = brick.Mountdir + brick.Dir`

rafikc30: That's fine for me. We should store path and dir and calculate mount root like MountRoot := b.Path[:len(b.Dir)] .

One more thing As soon as this patch get's in, the snapshot functionality will break. Because, if you have any restored volume or cloned volume, then Mountdir is different in both case.

aravindavk: If e2e tests don't fail then I prefer submitting snapshot related changes as separate PR. is that okay?

aravindavk:
Addressed the review comments. Please review.

glusterd2/volume/volume-utils.go
rafikc30: sure. We can create an issue

pkg/api/volume_req.go
kshlm: I think otherwise. But we can finalize the request structure later.

Issue #887, opened 2018-06-14
Capture all incoming commands in a logfile


atinmu:
We need something similar to what GD1 provides in cmd_history.log where all incoming CLI commands on that node are logged.

aravindavk:
incoming API requests are logged(we may have to add parameters as well). CLI logs will not be useful if some functionalities are executed via APIs and some via CLI.

atinmu:
cmd_history.log file is actually populated by glusterd. What I am requesting for a similar placeholder where one can easily figure out what all commands were (un)successfully executed with a timestamp. Having this captured in glusterd log file itself might not satisfy the need. Believe me, this has immensely helped in debugging different types of problems in glusterd.

phlogistonjohn:
As @aravindavk noted it makes more sense to log the API calls rather than the cli commands alone.

However, I do have two mad-science ideas. First, command line tools could send the originating cli arguments as part of the rest call, either in a header or in specialized attributes common across the api. If present, the logging would record the original cli that initiated the api in question.

Next, the HTTP requests and responses could be handled by a separate logger/logging endpoint. This endpoint would only contain the requests and responses (something akin to the access_log in apache)  I quickly skimmed the logrus docs and it looks like this is doable. However, I think this would only apply in configurations where GD2 logs to files. In a containerized scenario the logging is expected to go to stdio.

atinmu:
On Fri, Jun 15, 2018 at 11:36 PM, John Mulligan <notifications@github.com>
wrote:

> As @aravindavk <https://github.com/aravindavk> noted it makes more sense
> to log the API calls rather than the cli commands alone.
>
> However, I do have two mad-science ideas. First, command line tools could
> send the originating cli arguments as part of the rest call, either in a
> header or in specialized attributes common across the api. If present, the
> logging would record the original cli that initiated the api in question.
>
> Next, the HTTP requests and responses could be handled by a separate
> logger/logging endpoint. This endpoint would only contain the requests and
> responses (something akin to the access_log in apache) I quickly skimmed
> the logrus docs and it looks like this is doable. However, I think this
> would only apply in configurations where GD2 logs to files. In a
> containerized scenario the logging is expected to go to stdio.
>

This would satisfy the requirement. The goal is to capture the requests of
the API (with the detailed parameters) and their respective response status
*somewhere* which can be parsed and readable in an easy manner.

amarts:
[...]

PR #978, opened 2018-07-09
[RFC] Rest logging improvements 

phlogistonjohn:
Issue #887 requests a few improvements to the logging, based on that discussion I've done the following:
* Added request logging before an http request is processed, adding more keys to the log lines
* Add more specific user agents and a way for callers of the Go rest client api to set more specific agent info
* Added the ability for cli tools, specifically glustercli, to pass the original command line arguments to the server so that the server can log them.

Still to-do is the ability to log these items to an additional/separate log file.

I figured I'd throw an RFC PR out to get some feedback on the approaches I took here. I do not expect this PR to be merged in this state, just looking for discussion.


prashanthpai:
I like the overall approach.

It'll be great if changes such as this come with a sample output. I pulled this to local repo to see how it looks like:

***curl -s -X GET http://127.0.0.1:24007/v1/peers**
```
INFO[2018-07-13 11:34:52.709635] HTTP Request                                  client=curl/7.53.1 method=GET reqid=80380fc9-c095-4f04-a914-f0ad4a771024 source="[request_logging.go:29:middleware.LogRequest.func1]" subsys=rest url=/v1/peers
INFO[2018-07-13 11:34:52.712011] 127.0.0.1 - - [13/Jul/2018:11:34:52 +0530] "GET /v1/peers HTTP/1.1" 200 224 "" "curl/7.53.1"  reqid=80380fc9-c095-4f04-a914-f0ad4a771024 subsys=rest
```

**glustercli peer list**
```
INFO[2018-07-13 11:35:24.705967] HTTP Request                                  client=Go-http-client/1.1 method=GET reqid=33de6a4a-db30-457f-a626-142e87b5e62e source="[request_logging.go:29:middleware.LogRequest.func1]" subsys=rest url=/v1/peers
INFO[2018-07-13 11:35:24.713151] 127.0.0.1 - - [13/Jul/2018:11:35:24 +0530] "GET /v1/peers HTTP/1.1" 200 224 "" "Go-http-client/1.1"  reqid=33de6a4a-db30-457f-a626-142e87b5e62e subsys=rest
```

Some observations and suggestion:

* We can lose the `subsys` field if it's not deemed useful
* Ensure that `source` field doesn't get logged for this specific use.
* Rephrase `HTTP Request` to clearly indicate that at that point we've only accepted an incoming request but haven't processed it yet.

pkg/restclient/common.go
```diff
@@ -114,6 +123,8 @@ func (c *Client) do(method string, url string, data interface{}, expectStatusCod
 		req.Header.Set("Authorization", "bearer "+getAuthToken(c.username, c.password))
 	}
 
+	c.sendOriginArgs(req)
```
prashanthpai: Suggestion: this can be named to `setOriginArgs`

glusterd2/middleware/request_logging.go
```diff
@@ -14,8 +14,20 @@ import (
 // Apache Common Log Format (CLF)
 func LogRequest(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		entry := log.WithField("reqid", gdctx.GetReqID(r.Context()).String())
+		f := log.Fields{
+			"subsys": "rest",
+			"reqid":  gdctx.GetReqID(r.Context()).String(),
+		}
+		if origin := r.Header.Get("X-Gluster-Origin-Args"); origin != "" {
+			f["origin"] = origin
```
rishubhjain: TRIVIAL: IMO origin seems a bit ambiguous, a better variable name can be chosen, something like orgnArgs.

glustercli/cmd/common.go
```diff
@@ -20,6 +21,8 @@ var (
 
 func initRESTClient(hostname, user, secret, cacert string, insecure bool) {
 	client = restclient.New(hostname, user, secret, cacert, insecure)
+	client.ExtendAgent(
+		fmt.Sprintf("glustercli/%v", version.GlusterdVersion))
```
aravindavk: nit: We can have cli version instead of glusterd version. For now it is same, but it can change in future. @prashanthpai  your thoughts?

pkg/restclient/common.go
```diff
@@ -163,3 +174,34 @@ func (c *Client) do(method string, url string, data interface{}, expectStatusCod
...
+func (c *Client) setAgent(req *http.Request) {
+	req.Header.Set("User-Agent",
+		fmt.Sprintf("%v (Go-http-client/1.1)", c.agent))
+}
+
+// ExtendAgent adds the given string to the client's user agent
+// by prefixing it to the existing agent information.
+// This allows client programs to identify themselves more than
+// just something using the rest client api.
+func (c *Client) ExtendAgent(a string) {
```
aravindavk: please rename "a" to something meaningful.

pkg/restclient/common.go
```diff
@@ -163,3 +174,34 @@ func (c *Client) do(method string, url string, data interface{}, expectStatusCod
...
+		req.Header.Set("X-Gluster-Origin-Args",
+			fmt.Sprintf("1:%#v", strings.Join(c.originArgs, " ")))
+	}
+}
+
+// SetOriginArgs provides a way for tools using this library to
+// inform the server what arguments were provided to generate
+// the api call(s). The contents of the array are meant only for
+// human interpretation but will generally be command line args.
+func (c *Client) SetOriginArgs(a []string) {
```
aravindavk: same here, please rename "a"

pkg/restclient/common.go
```diff
@@ -163,3 +174,34 @@ func (c *Client) do(method string, url string, data interface{}, expectStatusCod
...
+func (c *Client) ExtendAgent(a string) {
+	// new additions to the agent identifier go to the
+	// beginning of the string. It is meant to read like:
+	// foo (based on) bar (based on) baz, etc...
+	c.agent = fmt.Sprintf("%v %v", a, c.agent)
+}
+
+func (c *Client) sendOriginArgs(req *http.Request) {
+	if len(c.originArgs) != 0 {
+		req.Header.Set("X-Gluster-Origin-Args",
```
aravindavk: This is nice. is this logged? I don't see this detail in output @prashanthpai posted https://github.com/gluster/glusterd2/pull/978#issuecomment-404738074

pkg/restclient/common.go
kshlm: We have a `c.SetOriginArgs` which does a different job to `sendOriginArgs`. The suggested rename makes the code more confusing.

I'd suggest that it be renamed to  `addOriginArgs`, but that would take away the consistent naming with the other "set" methods being called. Maybe also rename 'c.setAgent` to `c.addAgent`?

glustercli/cmd/common.go
prashanthpai: IMO, this should be okay. glustercli and glusterd are almost always built together. We can quickly identify if older/outdated clients connect to more recent glusterd2 by looking at the version if they follow same versioning. A separate version for CLI will makes more sense if it's a separate repo/project.

aravindavk: makes sense. we can update later if that happens.

atinmu:
Do we need this to extend to sprint 2? Or we can close it in a day's time?

glusterd2/middleware/request_logging.go
phlogistonjohn: Sounds good.

pkg/restclient/common.go
phlogistonjohn: It should be, but only if the client sends it. I'm not exactly sure exactly how @prashanthpai got his examples, but when I incorporate the other feedback for these changes I'll include examples of the logging as well.

pkg/restclient/common.go
phlogistonjohn: Perhaps we keep SetOriginArgs and rename this to addOriginArgsToHeaders ?

phlogistonjohn:
>     We can lose the subsys field if it's not deemed useful

Subsys was going to be how I redirected some of the logging into an special purpose log file equivalent to what @atinmu orginally requested. Additionally, I actually think the general approach is good; tag subsystems so that you can easily grep a logfile for the subsystem you care about. Note I'm not deeply in love with the 'subsys' name, but I think there's some value here.

>    Ensure that source field doesn't get logged for this specific use. I hope moving delete(entry.Data, logging.SourceField) line before logging entry should do it.

When I was working on this change orginally, I was wondering why this was one of the only places that didn't log the source. I tried to make it so that the logging that was originally there continued to skip the 

If you really want it gone I think we'll need to tweak the hook that generates the source item rather than 
trying to work around it. I will take a look into this - but I may suggest this can go in a separate pr.


>     Rephrase HTTP Request to clearly indicate that at that point we've only accepted an incoming request but haven't processed it yet.

Good point. Will do.

> That is okay for now as it helps in debugging if it's in the same logfile. It's easier to track timelines.

Sure, based on the original issue I was going to add an additional option like `--accesslog [/path/to/file | STDOUT] that would start a seperate "logging stream" for certain log entries (see above). When deployed as a container this could certainly be off by default but in the more traditional deployment the service file might specify --accesslog to create the gd2 equivalent of the command log.

pkg/restclient/common.go
prashanthpai: My bad. I think I did `make glusterd2` as I usually do. I can see it after I recompiled the CLI.

```log
INFO[2018-07-17 10:03:37.209081] HTTP Request                                  client="glustercli/v4.1.0-97.git987fac1 GlusterD2-rest-client/v4.1.0-97.git987fac1 (Go-http-client/1.1)" method=GET origin="1:\"glustercli peer list\"" reqid=817c7683-37dd-4b04-a589-16136deaf766 source="[request_logging.go:29:middleware.LogRequest.func1]" subsys=rest url=/v1/peers
INFO[2018-07-17 10:03:37.209832] 127.0.0.1 - - [17/Jul/2018:10:03:37 +0530] "GET /v1/peers HTTP/1.1" 401 71 "" "glustercli/v4.1.0-97.git987fac1 GlusterD2-rest-client/v4.1.0-97.git987fac1 (Go-http-client/1.1)"  origin="1:\"glustercli peer list\"" reqid=817c7683-37dd-4b04-a589-16136deaf766 subsys=rest
```

aravindavk: nice, thanks for the update

atinmu:
@phlogistonjohn would you mind refreshing this?

phlogistonjohn:
Sorry for the long delay. I will try to get around to this soon.

PR #981, opened 2018-07-10
change group 

Madhu-1:
if the gluster group is not present on the
a system where glusterd2 is supposed to run,
set Current user GID as UID for auth file

### DIFF
--- a/glusterd2/gdctx/global.go
+++ b/glusterd2/gdctx/global.go
@@ -94,18 +94,29 @@ func GenerateLocalAuthToken() error {
 }
 
 func protectAuthFile(authfile string) error {
-	usr, err := user.LookupGroup("gluster")
-	if err != nil {
-		return err
-	}
-	gID, err := strconv.Atoi(usr.Gid)
-	if err != nil {
-		return err
-	}
+	var uGID string
+
 	cuser, err := user.Current()
 	if err != nil {
 		return err
 	}
+
+	uGID = cuser.Gid
+
+	usr, err := user.LookupGroup("gluster")
+	if err != nil {
+		if _, ok := err.(user.UnknownGroupError); !ok {
+			return err
+		}
+	} else {
+		uGID = usr.Gid
+	}
+
+	gID, err := strconv.Atoi(uGID)
+	if err != nil {
+		return err
+	}
+
 	uID, err := strconv.Atoi(cuser.Uid)
 	if err != nil {
 		return err


Madhu-1:
@aravindavk please review

glusterd2/gdctx/global.go
```diff
@@ -94,18 +94,33 @@ func GenerateLocalAuthToken() error {
...
 	usr, err := user.LookupGroup("gluster")
 	if err != nil {
-		return err
+		if _, ok := err.(user.UnknownGroupError); ok {
+			unknownGrp = true
+		} else {
+			return err
+		}
+	}
+	if err == nil {
```
rishubhjain: @Madhu-1 This can be a simple else statement.

glusterd2/gdctx/global.go
```diff
@@ -94,18 +94,35 @@ func GenerateLocalAuthToken() error {
...
 func protectAuthFile(authfile string) error {
+	var (
+		uGID       string
+		unknownGrp bool
+	)
 	usr, err := user.LookupGroup("gluster")
 	if err != nil {
-		return err
+		if _, ok := err.(user.UnknownGroupError); ok {
+			unknownGrp = true
```
aravindavk: You can set `uGID = cuser.Gid` as default and avoid one extra variable(`unknownGrp`). Move `cuser, err := user.Current()` before `LookupGroup()`

Madhu-1: done

aravindavk:
> It doesn't appear that this change (and other changes -- but this on is convenient for me to comment on) has met the 24hr review period guidelines. This is a gentle reminder that we agreed to let changes remain open a while for wider community review and feedback.

24hr review period is only for non-trivial patches. Without this patch, glusterd2 will not start if "gluster" group doesn't exist before start(Regression from one of the patch). This is problem for other integrations who consume nightly builds.

phlogistonjohn:
> 24hr review period is only for non-trivial patches. Without this patch, glusterd2 will not start if "gluster" group doesn't exist before start(Regression from one of the patch). This is problem for other integrations who consume nightly builds.

I thought that in the discussion we had discarded the trivial rule (it was too vague). On the other hand if this is need to prevent test runs from breaking I'm ok that it was "fast tracked". However, I would not have complained if that was noted in the PR description or a comment on the PR.

PR #982, opened 2018-07-10
Make etcd client TLS options configurable

### LINKED ISSUE #979: config: Support etcd client TLS config as configurable option
When connecting to an external etcd cluster, glusterd2 should be able to use `TLS`.  glusterd2 should take a set of configurable options that specifically pertain to glusterd2 to external etcd communication: `--etcd-ca-file`, `--etcd-cert-file` and `--etcd-key-file` command line flags.

prashanthpai:
This change adds three new options to glusterd2's config which enables
using TLS when connecting to an external etcd cluster. The options added
are:

        --etcd-client-cert-file
        --etcd-client-key-file
        --etcd-client-ca-file

The above options are modeled after similar options present in `etcdctl`
utility.

The scope of this change is limited to external/remote etcd cluster and
doesn't alter things for elasticetcd.

Closes #979 


prashanthpai:
retest this please

glusterd2/store/config.go
```diff
@@ -16,14 +15,28 @@ import (
...
 const (
-	noEmbedOpt       = "noembed"
-	etcdEndpointsOpt = "etcdendpoints"
-	etcdCURLsOpt     = "etcdcurls"
-	etcdPURLsOpt     = "etcdpurls"
-	etcdLogFileOpt   = "etcdlogfile"
-
+	// etcd client options
+	noEmbedOpt = "noembed"
+	// TODO: Remove noembed as config option, presence of "etcdendpoints" would be sufficient
```
kshlm: Not really. If "noEmbed" is set, the store will use `etcd/clientv3`. If not (by default), store will use `elasticetcd` as the etcdclient. `elasticetcd` would start its embedded etcd instance when connected to an external etcd cluster as no active leaders would exist and it would vote itself as the leader.

Issue #414, opened 2017-11-10
Use OpenAPI/swagger for API specification 


prashanthpai:

sabose:
Can we raise priority on this - to use Swagger for autogen of python-bindings?

aravindavk:
> Can we raise priority on this - to use Swagger for autogen of python-bindings?

Sure. I have in-progress API specification for Volume info, I will submit that PR. Other APIs specifications can be added later. Let us know if that is fine.

phlogistonjohn:
@ansiwen  is interested in this topic as well

See also: https://github.com/gluster/python-gluster-mgmt-client/issues/9

PR #985, opened 2018-07-10
Open API definition file and tool to generate documentation

aravindavk:
- Volume Info API is covered now
- JSON is selected instead of Yaml since it is easy to add example
- This can be made more modular by using `$ref` feature of Open API

Updates: #414

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ DEPENV ?=
 PLUGINS ?= yes
 FASTBUILD ?= yes
 
-.PHONY: all build binaries check check-go check-reqs install vendor-update vendor-install verify release check-protoc $(GD2_BIN) $(GD2_BUILD) $(CLI_BIN) $(CLI_BUILD) cli $(GD2_CONF) gd2conf test dist dist-vendor functest
+.PHONY: all build binaries check check-go check-reqs install vendor-update vendor-install verify release check-protoc $(GD2_BIN) $(GD2_BUILD) $(CLI_BIN) $(CLI_BUILD) cli $(GD2_CONF) gd2conf test dist dist-vendor functest apidoc
 
 all: build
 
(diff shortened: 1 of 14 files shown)


aravindavk:
Documentation can be generated using swagger tools. To see the sample documentation online, go to https://generator.swagger.io/ and paste the below link and click "Explore"

     https://raw.githubusercontent.com/aravindavk/glusterd2/5bee22a61be7ba84b7ee0d5c8f94c49ede8f8565/doc/openapi/api.json

aravindavk:
@devyanikota @sac please see this file can be used for python client generation.

prashanthpai:
+1 for the `/doc/openapi` changes :+1: That's our long term goal.

I'm not a big fan of the python tool though as it's intermediate tool and generates markdown. For developers, the current API reference (godoc) should be sufficient and it generates doc from code.

Isn't it enough if we can use the swagger web UI to point to the json/yaml file in repo ?

prashanthpai:
> I looked for available tools for generating documents, but I couldn't get swagger tools working with openapi 3.x(All existing tools are supporting openapi/swagger 2.x)

That's right. I had checked this when I filed https://github.com/gluster/glusterd2/issues/414 and doesn't seem like stable support for 3.x is coming anytime soon

Can we simply have our spec in openapi/swagger 2.x ?

aravindavk:
> Can we simply have our spec in openapi/swagger 2.x ?

Possible. but it is easy to organize in separate files with spec 3.0. 

What problems you are seeing with the python based generator?

prashanthpai:
> What problems you are seeing with the python based generator?

It's something we have to maintain and keep regenerating markdown document. OpenAPI spec 2.0 has generators which have been stable and tested. The UI is a lot better than our markdown page. It can also generate client libraries, I'm not sure how useful that would be though.

You also mention:

> Python tool is temporary solution, if we get support for one of the standard tools then we can remove this.

OpenAPI 2.0 based tools can be those standard tools.


> Possible. but it is easy to organize in separate files with spec 3.0.

IMO, the advantages of using spec 2.0 outweigh the disadvantage of having a single large json/yml file.

prashanthpai:
To be fair, I'm not proud of our current `endpoints.md` either. It's not a substitute for a proper REST API documentation with examples. That too was a temporary solution

aravindavk:
Closing this PR for now, since Openapi tools not yet have support for version 3.0 spec. New PR can be opened for openapi spec supporting 2.0

PR #992, opened 2018-07-11
return HTTP response as nil

Madhu-1:
currently,we are returning a string response in response body and setting the ```Content-Type``` as
```"application/json; charset=UTF-8"```

to avoid setting this header for string response, send nil to disable setting this header.


Madhu-1:
@prashanthpai @aravindavk @rishubhjain please review

prashanthpai:
@Madhu-1 The PR title is a side effect. It's not the fix per se. The PR title can be - `Do not return plaintext  in response body`

The goal is to return no body or json body.

prashanthpai:
Also can you check if string is passed in other parts of code ?

Madhu-1:
@prashanthpai, I have checked in all file. we are not returning string response

Issue #993, opened 2018-07-11
Improper error message for failed REST authentication


prashanthpai:
```sh
$ glustercli peer list
Failed to get Peers list

Error: Request failed with HTTP Status code 401
invalid character 'A' looking for beginning of value
```

I pulled the latest glusterd2 changes which presumably had the change which turns on REST authentication by default.

aravindavk:
@Madhu-1 please look into this. Is this related to the restauth changes?

Madhu-1:
@aravindavk this is not related to recent changes, this is due to the error handling when we get an unexpected status code in restclient.

PR #997, opened 2018-07-12
send HTTP error response using  restutils.SendHTTPError()

Madhu-1:
send HTTP Error response using ```restutils.SendHTTPError()```
to avoid unmarshal error in ```restclient```

Fixes: #993 

sample output:
```
./build/glustercli peer list 

Failed to get Peers list

Error: Request failed with HTTP Status code 401
Authorization header is required

```


Madhu-1:
@prashanthpai @aravindavk  please review

glusterd2/middleware/auth.go
```diff
@@ -80,7 +82,7 @@ func Auth(next http.Handler) http.Handler {
 
 		// Check if token is Valid
 		if err != nil || !token.Valid {
-			http.Error(w, err.Error(), http.StatusUnauthorized)
+			restutils.SendHTTPError(ctx, w, http.StatusUnauthorized, err.Error())
```
rishubhjain: just send err instead of err.Error() and in this case the error can be nil. I think the if condition should have && instead of ||, or send a different error if error is nil.

prashanthpai: That's a good catch! If `err` is nil, `err.Error()` will crash.

Madhu-1: if ```err is nil``` am not in favor of sending nil response in this case
am not sure when we can hit this issue, but for the safer side, I have separated out the checks.

glusterd2/middleware/auth.go
```diff
@@ -36,18 +38,18 @@ func Auth(next http.Handler) http.Handler {
...
-			http.Error(w, "Authorization header is required", http.StatusUnauthorized)
+			restutils.SendHTTPError(ctx, w, http.StatusUnauthorized, errors.New("Authorization header is required"))
 			return
 		}
 
 		// Verify the Authorization header format "Bearer <TOKEN>"
 		authHeaderParts := strings.Split(authHeader, " ")
 		if len(authHeaderParts) != 2 || strings.ToLower(authHeaderParts[0]) != "bearer" {
-			http.Error(w, "Authorization header format must be Bearer <TOKEN>", http.StatusUnauthorized)
+			restutils.SendHTTPError(ctx, w, http.StatusUnauthorized, errors.New("Authorization header format must be Bearer <TOKEN>"))
```
prashanthpai: `Authorization header must be of the format - Bearer <TOKEN>`

glusterd2/middleware/auth.go
```diff
@@ -79,8 +81,12 @@ func Auth(next http.Handler) http.Handler {
...
 
 		// Check if token is Valid
-		if err != nil || !token.Valid {
-			http.Error(w, err.Error(), http.StatusUnauthorized)
+		if err != nil {
+			restutils.SendHTTPError(ctx, w, http.StatusUnauthorized, err.Error())
+			return
+		}
+		if !token.Valid {
+			restutils.SendHTTPError(ctx, w, http.StatusUnauthorized, errors.New("Invalid Token in Authorization header"))
```
prashanthpai: `token` (all small caps)

`Invalid token specified in Authorization header`

Also, for easy understand-ability, use single quotes around `Authorization` to make it clear that it's a header field. Here and elsewhere

Madhu-1: done

glusterd2/middleware/auth.go
Madhu-1: done

glusterd2/middleware/auth.go
phlogistonjohn: I actually think it maybe impossible to hit based on some similar work I did in heketi. It seems to be a quirk of the jwt library's api that it has this "extra" property but I think that in normal use of the api you can't hit this as you'll always get non-nil error from the parse function (IIRC). It made getting full code coverage in this area in heketi impossible! :-o

prashanthpai:
retest this please

PR #1004, opened 2018-07-12
Glusterfs regression tests

aravindavk:
All tests from `glusterfs` project are copied to `e2e/glusterfs-tests` directory and populated `all_tests.sh` file using `find . -name "*.t" | awk {print # prove -v } >> all_tests.sh`

For incremental integration, enable each test in `all_tests.sh` by uncommenting and identify the issues while running `./all_tests.sh`

**Notes**:

- glusterfs project uses `prove` for the tests. Install `perl-Test-Harness` package
- Default values for glusterd2 and glustercli is hard coded now, that needs to be changed


aravindavk:
- First commit is copy "tests" directory from glusterfs.
- Second commit enables `basic/volume.t` test

aravindavk:
> @aravindavk I am not sure if having a copy of all glusterfs tests into GD2 repo would make much sense here. The primary concern what I have is the redundancy and maintenance. I do understand that making same test from glusterfs source to work with GD1 & GD2 is very difficult but is that impossible? Probably if you can make a write up on what's the overall thinking here that'd help us to get onboarded here.

Makes sense. I am still figuring out best way to port the tests. I posted here temporarily for review.

Are we expecting that these tests to be backward compatible with old versions of gluster or only expected to run with glusterd2?

aravindavk:
Closing this PR for now. Changes needs to be pushed to glusterfs repo as discussed in the comments

Issue #1019, opened 2018-07-16
`peer add' fails with `Authorization header is required' message


sac:
I tried to add a peer following the documentation at:

https://github.com/gluster/glusterd2/blob/master/doc/quick-start-user-guide.md#add-peer

It fails with message `Authorization header is required'. However, when I add a peer using the glustercli, the peer add works fine. We have add documentation on how deal with the `Authorization' part.

PR #1021, opened 2018-07-17
doc: REST API authentication

aravindavk:
Closes: #1019


pkg/tools/glustercli-auth-header.py
```diff
@@ -0,0 +1,21 @@
...
+    claims['exp'] = datetime.utcnow() + timedelta(seconds=10)
+
+    token = jwt.encode(claims, secret, algorithm='HS256')
+    return (b'Authorization: bearer ' + token).decode()
+
+
+if __name__ == "__main__":
+    user = "glustercli"
+    secret_file = sys.argv[1]
+    secret = open(secret_file).read()
```
rishubhjain: I think its better to use:

`with open(secret_file) as f:`

instead of open().read() since file getting closed immediately is not assured. After python 3.2 it will thow a warning.ResourceWarning

aravindavk: this is only development tool, errors are not handled

aravindavk: I will update the tool

rishubhjain:
@aravindavk The document looks good, maybe we can add another heading stating the error received by the user if the auth fails or if proper authentication steps are not followed. It can be a part of another document (capturing all errors and their solutions) as well.

pkg/tools/glustercli-auth-header.py
aravindavk: Done

pkg/tools/glustercli-auth-header.py
```diff
@@ -0,0 +1,33 @@
...
+import jwt
+
+
+def jwt_token(user, secret):
+    claims = dict()
+    claims['iss'] = user
+    claims['iat'] = datetime.utcnow()
+    claims['exp'] = datetime.utcnow() + timedelta(seconds=10)
+
+    token = jwt.encode(claims, secret, algorithm='HS256')
```
Madhu-1: This looks fine, To provide more security, we need to add URL tampering protection also
@aravindavk any thoughts on  this

in the server side, we need to enable (HTTP method) validation to avoid miss use of the token.

aravindavk: makes sense. Please open new issue for the same.

doc/rest-authentication.md
```diff
@@ -0,0 +1,106 @@
...
+    Authorization: bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnbHVzdGVyY2xpIiwiaWF0IjoxNTMxNzI5NzQyLCJleHAiOjE1MzE3Mjk3NTJ9._WIwO7PrHUHIT62SdzfkyNjqD1GEgX2cYqN8ACZCtaw
+
+**Note**: REST Authentication can be disabled by adding `restauth=false`
+in Glusterd2 config file(Default path is
+`/var/lib/glusterd2/glusterd2.toml` in case of rpm installation)
+
+## Default user
+
+When `glusterd2` starts for the first time, it creates the
+`$GLUSTERD2_STATEDIR/auth` file which will contain the secret. If
```
atinmu: Do we document anywhere what GLUSTERD2_STATEDIR path mean? Might be good to point out to a link.

doc/rest-authentication.md
```diff
@@ -0,0 +1,106 @@
...
+"gluster" user group is available in the system then this `auth` file
```
atinmu: Do we mention how to create "gluster" user group?

doc/rest-authentication.md
```diff
@@ -0,0 +1,106 @@
...
+
+Secret is taken by `glustercli` in following order of precedence
+(highest to lowest):
+
+    --secret
+    --secret-file
+    GLUSTERD2_AUTH_SECRET (environment variable)
+    --secret-file (default path)
+
+## Curl example
```
atinmu: Should be under ReST API section?

doc/rest-authentication.md
kshlm: > Do we document anywhere what GLUSTERD2_STATEDIR path mean? Might be good to point out to a link
I don't think we have a document about the paths yet. We should do it.

Apart from that, I have a concern about using the GD2 secret by external tools.

The secret is meant to be accessible only by GD2. GD2 should be using this secret to create and handout tokens. Having the clients access the secret is a big security risk. Later in the document, we ask that the secret be copied out externally and have external client generate their own token. This is a serious lapse of security. Because the secret can be passed along to unintended users and those users would have full access to the GD2 cluster.

Ideally, as I mentioned earlier GD2 should be the one generating and handing out tokens. Even for `glustercli` it would be preferrable that you use token. GD2 should generate a token for CLI and save it to a known place. And we should add a way to generate tokens to GD2 (ReST or CLI to be decided).

Sorry about bringing this up now. I should have been paying attention when the actual PR was being reviewed. I'll open up a new issue about `glustercli` not using the secret file directly. We can merge this PR without waiting for that, but only if it is modified to address my comments later on.

doc/rest-authentication.md
kshlm: I don't think we need to do this. This is simple unix knowledge. In any case, if installed from packages, the package should take care of creating the users/groups. I don't think any gluster packages do this yet.

doc/rest-authentication.md
kshlm: A link should be added from the ReST API document to this doc.

doc/rest-authentication.md
```diff
@@ -0,0 +1,106 @@
...
+    resp = requests.get("http://localhost:24007/v1/peers",
+                        headers={"Authorization": "bearer " + token}
+
+    peers = []
+    if resp.status_code == 200:
+        peers = json.loads(resp.content)
+
+
+## Using REST APIs from outside the Cluster nodes
+
```
kshlm: I'd prefer you remove this section completely for now. Or at least modify such that it doesn't involve copying out the secret file. The `glustercli` example would need to be completely removed. For the `curl` example, you should change it so that the python script is used to generate a token on one of peers, and the token is copied over.

aravindavk:
I think confusion here is user documentation vs developer documentation. I should convert this as developer documentation since users need not make any change to glustercli arguments if deployment is unchanged.(Like custom localstatedir)

I will remove the last section which talks about the temporary arrangement to use this from outside the cluster. Once user management is implemented then they can use those credentials for making REST calls.

Till the user management is implemented, we can't use these REST APIs from outside the cluster.

doc/rest-authentication.md
aravindavk: makes sense. I updated the same in other comment https://github.com/gluster/glusterd2/pull/1021#issuecomment-405943600

> For the curl example, you should change it so that the python script is used to generate a token on one of peers, and the token is copied over.

We are following shared secret approach. No change required for curl example unless we decide to do shared token approach again.

aravindavk:
Added comment about Shared token vs Shared secret approach here https://github.com/gluster/glusterd2/issues/1030

aravindavk:
> Apart from that, I have a concern about using the GD2 secret by external tools.

This is temporary till the user management is in place. Once we implement user management, only  `glustercli` will use the secret file. Future REST API documentation may looks like,

- Create user/app using `glustercli useradd <username>` which will prompt for secret and creates the user. For example: `glustercli useradd webui`
- Once the user is created, REST APIs can be consumed from any node by authenticating using the credentials specified in previous command. For example, `glustercli --user webui --secret-file=~/.gluster/webui.secret peer status`. Secret can be specified using `--secret`, `--secret-file`, env var etc.

doc/rest-authentication.md
phlogistonjohn: It may be worthwhile to point out this is a unix group interacting with file permissions and not some special gluster thing in that case. The word group and phrase "in that machine" alone may not hint at that strongly enough.

doc/rest-authentication.md
```diff
@@ -0,0 +1,106 @@
...
+"gluster" user group is available in the system then this `auth` file
+can be read by any user in that machine who are part of "gluster"
+group.
+
+**Note**: If "gluster" user group is not available during first start
+of `glusterd2` it limits the read permission to `root:root`
+
+## glustercli
+
+With default installation, `glustercli` will know the location of
```
phlogistonjohn: Perhaps, "The glustercli command knows the default location of the auth file." would read more clearly.

doc/rest-authentication.md
```diff
@@ -0,0 +1,106 @@
...
+`auth` file. `glustercli` will generate JWT token using the secret
+available in `auth` file and attach it with every REST API calls.
```
phlogistonjohn: "every REST API call."  or "all REST API calls."

doc/rest-authentication.md
```diff
@@ -0,0 +1,106 @@
...
+Thats all, use `gcurl` wherever `curl` is necessory. For example, to start a Volume
+
+    gcurl -XPOST http://localhost:24007/v1/volumes/gv1/start
+
+## Python example
+
+Install `jwt` library using `pip install jwt` or using `dnf install
+python-jwt`. Generate JWT token using,
+
+    import jwt
```
phlogistonjohn: It might make sense to enclose this block with ````python` esp. if its expected this will be viewed using github/github-style-markdown-renderer.

Madhu-1:
@aravindavk anything pending on this PR?

Madhu-1:
@aravindavk ping, can you address review comments and  resolve the merge conflict so that we can get  this PR in.

PR #1028, opened 2018-07-18
cli/restclient: Make client timeout configurable

### LINKED ISSUE #1027: Make client side timeout configurable
...right now, client side timeout is hard-coded to 30s

@rmadaka will be using this in his scale testing

prashanthpai:
The client timeout is now configurable at the restclient layer and also
in the CLI. The client timeout defaults to 30 seconds in both restclient
and CLI.

Timeout can be configured in:
  * restclient using SetTimeout() API.
  * CLI using --timeout command line flag.

This client timeout is the overall timeout which includes the time taken
from setting up TCP connection till client finishes reading the response
body from glusterd2.

Closes #1027 

### DIFF
--- a/glustercli/cmd/root.go
+++ b/glustercli/cmd/root.go
@@ -23,6 +23,7 @@ var (
 
 const (
 	defaultLogLevel = "INFO"
+	defaultTimeout  = 30 // in seconds
 )
 
 // RootCmd represents main command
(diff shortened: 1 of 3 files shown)


glustercli/cmd/root.go
```diff
@@ -84,14 +85,15 @@ var (
...
 	flagInsecure   bool
-	flagLogLevel   string
 	verbose        bool
+	flagCacert     string
+	flagLogLevel   string
 	flagUser       string
 	flagSecret     string
 	flagSecretFile string
+	flagEndpoints  []string
+	flagTimeout    int
```
Madhu-1: question:
can we declare ```flagTimeout``` as ```uint```?

prashanthpai: I initially thought about it but decided not do it though. Let me try it out and see if that can be a substitute for validating it.

prashanthpai: I've changed it to uint :+1:

PR #1032, opened 2018-07-18
Custom error message when peer remove fails because peer does not exists

### LINKED ISSUE #1022: Need custom error messages for failed commands
We need better messaging for failed commands. For example when I try to remove a non-existent peer GD2 throws error:

```
ERRO[2018-07-17 11:45:55.898972] peer remove failed error="Request failed with HTTP Status code 500\ncould not validate delete request" peerID=de46f1ce-0809-4f8d-aefa-cda1c84f9ee3
```
Which is a generic error and status code 500 refers to "Internal Server Error and is a very general HTTP status code that means something has gone wrong on the website's server."

Which is not the case here. The users are clueless as to why the command failed.

rishubhjain:
Closes #1022

### DIFF
--- a/glusterd2/commands/peers/deletepeer.go
+++ b/glusterd2/commands/peers/deletepeer.go
@@ -38,15 +38,11 @@ func deletePeerHandler(w http.ResponseWriter, r *http.Request) {
 	// Check whether the member exists
 	p, err := peer.GetPeerF(id)
 	if err != nil {
-		logger.WithError(err).Error("failed to get peer")
-		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, "could not validate delete request")
-		return
-	} else if p == nil {
-		logger.Debug("request denied, received request to remove unknown peer")
-		restutils.SendHTTPError(ctx, w, http.StatusNotFound, "peer not found in cluster")
+		logger.WithError(err).WithField("peerid", id).Error("Failed to get peer")
+		status, err := restutils.ErrToStatusCode(err)
+		restutils.SendHTTPError(ctx, w, status, err)
 		return
 	}
-
 	// You cannot remove yourself
 	if id == gdctx.MyUUID.String() {
 		logger.Debug("request denied, received request to delete self from cluster")


glusterd2/commands/peers/deletepeer.go
```diff
@@ -38,15 +39,15 @@ func deletePeerHandler(w http.ResponseWriter, r *http.Request) {
 	// Check whether the member exists
 	p, err := peer.GetPeerF(id)
 	if err != nil {
-		logger.WithError(err).Error("failed to get peer")
+		if err == errors.ErrPeerNotFound {
```
prashanthpai: What's wrong with returning the `err` back to client directly ? I'm sure that has the same or similar error string. I prefer not to handle `ErrPeerNotFound` in a special way

glusterd2/commands/peers/deletepeer.go
```diff
@@ -38,15 +39,15 @@ func deletePeerHandler(w http.ResponseWriter, r *http.Request) {
...
+			logger.WithError(err).WithField("peerid", id).Error("request denied, received request to remove unknown peer")
+			restutils.SendHTTPError(ctx, w, http.StatusNotFound, "peer not found in cluster")
```
Madhu-1: instead of "peer not found in cluster" return ```err```   in ```restutils.SendHTTPError```

Madhu-1: can we use ```ErrToStatusCode```, I think this will handle both the cases.

glusterd2/commands/peers/deletepeer.go
```diff
@@ -38,15 +39,11 @@ func deletePeerHandler(w http.ResponseWriter, r *http.Request) {
 	// Check whether the member exists
 	p, err := peer.GetPeerF(id)
 	if err != nil {
-		logger.WithError(err).Error("failed to get peer")
-		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, "could not validate delete request")
-		return
-	} else if p == nil {
-		logger.Debug("request denied, received request to remove unknown peer")
-		restutils.SendHTTPError(ctx, w, http.StatusNotFound, "peer not found in cluster")
+		log.WithError(err).WithField("peerid", id).Error("Failed to get peer")
```
Madhu-1: use ```logger ``` instead of ```log```

prashanthpai:
@rishubhjain You'll have to fix the PR description to automatically close the issue. Read more about it here: https://help.github.com/articles/closing-issues-using-keywords/

You can also refer to other PRs

PR #1036, opened 2018-07-19
Volume: Add auto provision flag to brick info

rafikc30:
Brick should be able to identify whether it is auto provisioned
or not. This is required for cases like

* Mounting of a brick in glusterd restart
* Umounting of a brick on volume stop/delete
* delete brick directoroy when volume is deleted
* delete vg/lv during volume delete
* To smoothly handle replace brick/reset brick on auto provisioned bricks

### DIFF
--- a/glusterd2/commands/volumes/volume-create-txn.go
+++ b/glusterd2/commands/volumes/volume-create-txn.go
@@ -60,6 +60,8 @@ func voltypeFromSubvols(req *api.VolCreateReq) volume.VolType {
 
 func populateSubvols(volinfo *volume.Volinfo, req *api.VolCreateReq) error {
 	var err error
+	provisionType := volinfo.GetProvisionType()
+
 	for idx, subvolreq := range req.Subvols {
 		if subvolreq.ReplicaCount == 0 && subvolreq.Type == "replicate" {
 			return errors.New("replica count not specified")
@@ -106,7 +108,7 @@ func populateSubvols(volinfo *volume.Volinfo, req *api.VolCreateReq) error {
 				return err
 			}
 		}
-		s.Bricks, err = volume.NewBrickEntriesFunc(subvolreq.Bricks, volinfo.Name, volinfo.VolfileID, volinfo.ID)
+		s.Bricks, err = volume.NewBrickEntriesFunc(subvolreq.Bricks, volinfo.Name, volinfo.VolfileID, volinfo.ID, provisionType)
 		if err != nil {
 			return err
 		}
@@ -144,14 +146,21 @@ func newVolinfo(req *api.VolCreateReq) (*volume.Volinfo, error) {
 		volinfo.Transport = "tcp"
 	}
 
-	if err := populateSubvols(volinfo, req); err != nil {
-		return nil, err
-	}
-
 	if req.Metadata != nil {
 		volinfo.Metadata = req.Metadata
 	} else {
 		volinfo.Metadata = make(map[string]string)
+	}
+
+	if req.Size > 0 {
+		//AutoProvisioned volume
+		volinfo.Metadata[brick.ProvisionKey] = string(brick.AutoProvisioned)
+	} else {
+		volinfo.Metadata[brick.ProvisionKey] = string(brick.ManuallyProvisioned)
+	}
+
+	if err := populateSubvols(volinfo, req); err != nil {
+		return nil, err
 	}
 
 	return volinfo, nil
--- a/glusterd2/brick/types.go
+++ b/glusterd2/brick/types.go
@@ -12,6 +12,9 @@ import (
 //go:generate stringer -type=Type
 type Type uint16
 
+//ProvisionType is the way a brick is provisioned
+type ProvisionType string
+
 const (
 	// Brick represents default type of brick
 	Brick Type = iota
@@ -38,6 +41,7 @@ type Brickinfo struct {
 	VolumeID       uuid.UUID
 	Type           Type
 	Decommissioned bool
+	PType          ProvisionType
 	MountInfo
 }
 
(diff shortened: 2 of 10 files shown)


rafikc30:
With the above approach,  we can save changes in req struct. But IMO, It cannot be in metadata field for permanent. Given that auto-provision is a brick specific feature, it is not a metadata information for a volume.

Assume a case where we have two bricks. Initially, both are self-provisioned. Then with a replace brick operation lets say we changed one brick to a normal brick. Then the volume metadata field won't make sense, as we have one brick as self-provisioned and one not. Going little further if we do replace brick on the brick as well, then we have to remove the metadata field.

phlogistonjohn:
I'm going to complain about the name 'SelfProvision' because I find "self" a bit ambiguous to the reader. People already talk about things as auto provisioned, as in the title of your pr, why not 'AutoProvisioned' ? 

I'm also frequently worried about upgradability and the nature of fields perhaps not always making sense as a boolean forever. In that vein, one could consider a new type like:
```golang
type ProvisionState string
const (
  ManuallyProvisioned ProvisionState = ""
  AutoProvisioned ProvisionState = "auto"
  MagicProvisioned ProvisionState = "magic-future-thing"
)
```
But I'm willing to be told this is a case of YAGNI.


> Assume a case where we have two bricks. Initially, both are self-provisioned. Then with a replace brick operation lets say we changed one brick to a normal brick. Then the volume metadata field won't make sense, as we have one brick as self-provisioned and one not. Going little further if we do replace brick on the brick as well, then we have to remove the metadata field.

I think this operation should simply not be allowed. Either the brick is replaced through device management functions or a special case of replace that understands the auto-provisioning mechanisms. There simply shouldn't be a way to mix auto-provisioned and manually-provisioned bricks on one volume IMO.

rafikc30:
@phlogistonjohn I agree that AutoProvisioned name suits better. And also instead of having a boolean field, we could have a more meaningful struct that can scale better in future.

IMO, Provisioning a brick is a brick related feature, not a volume field. We can do add/replace/reset brick through a different channel for auto-provisioned bricks, but it is also can be done with the same existing code. I still haven't convinced myself about the problem having both auto-provisioned and normal bricks in the same volume.

Having such flexibility really helps in features like snapshot, where snapshot bricks are auto-provisioned, but we have to support add/remove/replace/reset brick operations as normal. And this kind of operations becomes common when we have highly available snapshots. If bricks were missing at time taking snapshots, then we have to perform replace brick once we restored that snapshot.

So IMHO, strcture `type ProvisionState string`  should be a member of brick strcture.

Madhu-1:
@rafikc30 , need to resolve conflicts.

 question:
in volume response is there any field for differentiating the volume is a normal or smart volume?

rafikc30:
@Madhu-1 we have to add that field. I think we should add that field to brick response if we are having this struct in brickinfo.

kshlm:
Agree with @phlogistonjohn about having a `ProvisionState` type. That gives us more ways to describe the provisioning method ("auto" or "snapshot clone" vs just "self-provisioned") and leaves room for expansion later on. Although I'm not a fan of adding fields to the base volinfo and brickinfo structs that aren't  absolutely necessary for normal functioning of a manually created gluster volume. I'd prefer that we do make use of `Volinfo.Metadata`, and add a `Brickinfo.Metadata` as well for such needs.

rafikc30:
@aravindavk @kshlm @phlogistonjohn @Madhu-1 

I have refreshed the PR to address your comments.

Madhu-1:
@rafikc30  there is an import cycle issue in code please fix it
```
import cycle not allowed in test
```

aravindavk:
@rafikc30 do not add metadata through request struct. Set metadata in the Volume create handler. Volume create should not accept ineternal Metadata which starts with `_`(Currently it is accepted because validation is missing). Internal metadata keys are for plugins/components which can set Metadata key directly.

- Volume create request remains same as now
- Volume create handler sets internal Metadata based on `req.Size`

rafikc30:
retest this please

glusterd2/bricksplanner/planner.go
```diff
@@ -137,6 +141,9 @@ func getBricksLayout(req *api.VolCreateReq) ([]api.SubvolReq, error) {
 				TpMetadataSize: deviceutils.GetPoolMetadataSize(eachBrickTpSize),
 				FsType:         "xfs",
 				MntOpts:        "rw,inode64,noatime,nouuid",
+				Metadata: map[string]string{
```
aravindavk: This is still adding Metadata to request structure(All Metadata key starting with `_` need to be restricted via APIs). Add this directly to volinfo in the handler

pkg/utils/utils.go
```diff
@@ -19,6 +19,45 @@ import (
...
+
+const (
+	//ProvisionKey is used to set the type of provisioning in volume/brick metadata
+	ProvisionKey string = "_brick-provision-type"
+
+	//ManuallyProvisioned bricks will be provisioned by a user
+	ManuallyProvisioned ProvisionState = ""
+	//AutoProvisioned bricks will be provsioned by device manager from glusterd2
+	AutoProvisioned ProvisionState = "auto"
+	//SnapshotProvisioned brciks will be created by performing a snapshot of a gluster brick
```
prashanthpai: typo

glusterd2/commands/volumes/volume-create-txn.go
```diff
@@ -144,16 +147,23 @@ func newVolinfo(req *api.VolCreateReq) (*volume.Volinfo, error) {
...
-
 	if req.Metadata != nil {
 		volinfo.Metadata = req.Metadata
 	} else {
 		volinfo.Metadata = make(map[string]string)
 	}
 
+	if req.Size > 0 {
+		//AutoProvisioned volume
+		volinfo.Metadata[utils.ProvisionKey] = string(utils.AutoProvisioned)
```
prashanthpai: Any reason why you chose `Metadata` over adding a field to `Volinfo` ?

pkg/utils/utils.go
```diff
@@ -19,6 +19,45 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+//ProvisionState is the way a brick or all bricks are provisioned
```
prashanthpai: Utils package is not a good place for this

Why not volume package ?

pkg/utils/utils.go
```diff
@@ -19,6 +19,45 @@ import (
...
+	ManuallyProvisioned ProvisionState = ""
+	//AutoProvisioned bricks will be provsioned by device manager from glusterd2
+	AutoProvisioned ProvisionState = "auto"
+	//SnapshotProvisioned brciks will be created by performing a snapshot of a gluster brick
+	SnapshotProvisioned ProvisionState = "snapshot"
+)
+
+//IsManuallyProvisioned will return true if manually provisioned
+func (p ProvisionState) IsManuallyProvisioned() bool {
+	if p == ManuallyProvisioned {
```
prashanthpai: You can just `return p == ManuallyProvisioned`. Here and below

pkg/utils/utils.go
rafikc30: ProvisionState is also used in brickinfo, that is why I decided to define in utils

glusterd2/commands/volumes/volume-create-txn.go
rafikc30: That was based on previous review comments

pkg/utils/utils.go
prashanthpai: In that case, there can be two separate types - one in brick package and one in volume package.

prashanthpai:
@kshlm @aravindavk Metadata, when introduced was meant for users/clients (and plugins) to store arbitrary data (akin to labels in k8s) which aren't interpreted by core glusterd2 in any way. In this specific case, how a volume is provisioned is an actionable internal information used by features that are core i.e volume management and snapshot. IMO, this is better stored as a field in `Volinfo` and `Brickinfo`. It need not be exposed in API though.

atinmu:
@aravindavk @kshlm @prashanthpai can we get to a consensus on the approach?

PR #1048, opened 2018-07-20
arbiter: Support replica 2 arbiter 1 usage

### LINKED ISSUE #515: Arbiter volume creation syntax needs to change to 'replica 2 arbiter 1'
The  feedback from the community is that it would be better to use `replica 2 arbiter 1` instead of `replica 3 arbiter 1` in the volume create/ display syntax. See http://lists.gluster.org/pipermail/gluster-users/2017-December/033092.html

prashanthpai:
Change syntax for creation of arbiter volumes from:
`replica 3 arbiter `
to
`replica 2 arbiter 1`

Closes #515 

### DIFF
--- a/glustercli/cmd/volume-create.go
+++ b/glustercli/cmd/volume-create.go
@@ -166,26 +166,43 @@ func volumeCreateCmdRun(cmd *cobra.Command, args []string) {
 	if flagCreateReplicaCount > 0 {
 		// Replicate Volume Support
 
-		if numBricks%flagCreateReplicaCount != 0 {
-			failure("Invalid number of bricks specified", nil, 1)
+		// if arbiter count is specified
+		if cmd.Flags().Changed("arbiter") {
+			if flagCreateArbiterCount != 1 && flagCreateReplicaCount != 2 {
+				failure("Invalid arbiter/replica count specified. Supported arbiter configuration: replica 2, arbiter 1", nil, 1)
+			}
 		}
 
-		numSubvols := numBricks / flagCreateReplicaCount
+		if numBricks%(flagCreateReplicaCount+flagCreateArbiterCount) != 0 {
+			failure("Invalid number of bricks specified. Number of bricks must be a multiple of replica (+arbiter) count.", nil, 1)
+		}
 
+		// FIXME: Things would've been so much simpler if specifying the
+		// arbiter brick was simply a separate command flag such as
+		// --arbiter-brick or even a boolean as we always pick the last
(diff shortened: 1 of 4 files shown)


amarts:
I guess it is OK. Lets not keep the backward compatibility, if that is not the correct behavior. I did it in GD1 as it would be required for some time. But GD2 being a new project replacing GD1 later (at least not in next immediate release), we should be good here.

itisravi:
Okay, in that case it is fine.

glustercli/cmd/volume-create.go
```diff
@@ -166,26 +166,44 @@ func volumeCreateCmdRun(cmd *cobra.Command, args []string) {
...
+		if numBricks%(flagCreateReplicaCount+flagCreateArbiterCount) != 0 {
+			failure("Invalid number of bricks specified. Number of bricks must be a multiple of replica (+arbiter) count.", nil, 1)
 		}
 
-		numSubvols := numBricks / flagCreateReplicaCount
+		// FIXME: Things would've been so much simpler if specifying the
+		// arbiter brick was simply a separate command flag such as
+		// --arbiter-brick or even a boolean as we always pick the last
+		// brick as arbiter. Backward compatibility is a pain and we're
+		// doing the same bad choices all over again.
```
aravindavk: I think it is not only for backward compatibility, but it is for future extension. In future, if we support `replica 3` and `arbiter 2`. 

@itisravi please add your thoughts.

prashanthpai:
retest this please

rishubhjain:
retest this please

prashanthpai:
@aravindavk Can you review new changes in `glusterd2/bricksplanner/subvoltype_replicate.go` and `e2e/smartvol_ops_test.go` ?

glusterd2/bricksplanner/subvoltype_replicate.go
```diff
@@ -22,7 +22,7 @@ func (s *replicaSubvolPlanner) Init(req *api.VolCreateReq, subvolSize uint64) {
 }
 
 func (s *replicaSubvolPlanner) BricksCount() int {
-	return s.replicaCount
+	return s.replicaCount + s.arbiterCount
 }
 
 func (s *replicaSubvolPlanner) BrickSize(idx int) uint64 {
```
aravindavk: similar change required here as well

prashanthpai: Thank you. Done :+1:

Issue #891, opened 2018-06-15
Gluster user Documentation


aravindavk:
Porting upstream Gluster documentation to Glusterd2 and glustercli.

Upstream documentation: https://docs.gluster.org/en/latest

aravindavk:
We can host in `glusterd2/docs` initially, then we can send to `glusterdocs`. We can consider this as opportunity to restructure the documentation flow.

prashanthpai:
> We can host in glusterd2/docs initially, then we can send to glusterdocs. We can consider this as opportunity to restructure the documentation flow.

+1

prashanthpai:
> @prashanthpai Can we please have a section in gluster-doc about GD2 which can point to the quick start guide atleast to begin with?

I'll do that.

atinmu:
> Can we please have a section in gluster-doc about GD2 which can point to the quick start guide atleast to begin with?

https://github.com/gluster/glusterdocs/pull/387

Once we have coverage of all the glustercli commands, we need to point the same in https://github.com/gluster/glusterdocs/tree/master/CLI-Reference too.

PR #1051, opened 2018-07-20
doc: Managing the glusterd2 service

aravindavk:
Content mostly from glusterfs documentation, changed to suit glusterd2.

Additionally added "gluster" group creation step and `systemctl` commands

Updates: #891

### DIFF
--- a/doc/managing-the-glusterd2-service.md
+++ b/doc/managing-the-glusterd2-service.md
@@ -1,0 +1,54 @@
+# Managing the glusterd Service
+
+After installing GlusterFS and Glusterd2, you must start `glusterd2`
+service. The glusterd2 service serves as the Gluster elastic volume
+manager, overseeing glusterfs processes, and co-ordinating dynamic
+volume operations, such as adding and removing volumes across multiple
+storage servers non-disruptively.
+
+This section describes how to start the glusterd2 service in the
+following ways:
+
+**Note**: You must start glusterd2 on all GlusterFS servers.
+
+## Gluster user group
+
+On each server nodes, create "gluster" user group if not exists
+already using the following command.
+
+    # groupadd gluster
+
+Users from this group can run `glustercli` commands in GlusterFS
+servers.
+
+**Note**: Create "gluster" group before starting `glusterd2`
+
+## Starting and Stopping glusterd2 Manually
+
+This section describes how to start and stop glusterd2 manually
+
+To start glusterd2 manually, enter the following command:
+
+    # systemctl start glusterd2
+
+To stop glusterd2 manually, enter the following command:
+
+    # systemctl stop glusterd2
+
+## Starting glusterd Automatically
+
+This section describes how to configure the system to automatically
+start the glusterd2 service every time the system boots.
+
+    # systemctl enable glusterd2
+
+## Status
+
+Check the status of `glusterd2` using,
+
+    # systemctl status glusterd2
+
+To check the Cluster status, run the following command from one of the
+GlusterFS server.
+
+    # glustercli peer status


aravindavk:
Glusterfs documentation page(glusterd1): https://docs.gluster.org/en/latest/Administrator%20Guide/Start%20Stop%20Daemon/

doc/managing-the-glusterd2-service.md
```diff
@@ -0,0 +1,54 @@
+# Managing the glusterd Service
+
+After installing GlusterFS and Glusterd2, you must start `glusterd2`
+service. The glusterd2 service serves as the Gluster elastic volume
```
harigowtham: From the second line, we are going to repeat stuff.  As per the new doc, we are going to give them an insight about glusterd before reaching here. Its needed in the beginning for them to get a better understanding. If its in the beginning, this will be a repetition. will use the same snippet there.

doc/managing-the-glusterd2-service.md
```diff
@@ -0,0 +1,54 @@
...
+
+    # systemctl enable glusterd2
+
+## Status
+
+Check the status of `glusterd2` using,
+
+    # systemctl status glusterd2
+
+To check the Cluster status, run the following command from one of the
```
harigowtham: this section can go into a separate doc where we would have explained about peers and probing?

doc/managing-the-glusterd2-service.md
```diff
@@ -0,0 +1,54 @@
...
+already using the following command.
+
+    # groupadd gluster
+
+Users from this group can run `glustercli` commands in GlusterFS
+servers.
+
+**Note**: Create "gluster" group before starting `glusterd2`
+
+## Starting and Stopping glusterd2 Manually
```
harigowtham: Do we have to talk about starting from binaries? Need more thoughts here.
Thought of documenting how to install from rpm as well as the binaries. Do we need to have this as well or is it better to put that section as a separate link?

doc/managing-the-glusterd2-service.md
aravindavk: Setup and install is out of scope for this page. This page is for post installation steps

doc/managing-the-glusterd2-service.md
aravindavk: makes sense, I will remove this

doc/managing-the-glusterd2-service.md
aravindavk: I prefer not to mention about running binaries directly. It may lead to different non-standard configuration(No config file or different values in custom config file).

Also glusterd2 binary will not demonize by default, that may become issues.

doc/managing-the-glusterd2-service.md
rishubhjain: @harigowtham I think we should give a brief introduction on what is glusterd2 in the user guide and then provide a link to an extended explanation, else the doc will be too long and might be not very user-friendly to someone who already knows what glusterd2 is.

doc/managing-the-glusterd2-service.md
```diff
@@ -0,0 +1,54 @@
+# Managing the glusterd Service
+
+After installing GlusterFS and Glusterd2, you must start `glusterd2`
+service. The glusterd2 service serves as the Gluster elastic volume
+manager, overseeing glusterfs processes, and co-ordinating dynamic
+volume operations, such as adding and removing volumes across multiple
+storage servers non-disruptively.
+
+This section describes how to start the glusterd2 service in the
```
rishubhjain: I think this statement needs a re-phrase, and I think this section does more than describing "how to start the service."

doc/managing-the-glusterd2-service.md
```diff
@@ -0,0 +1,54 @@
...
+    # groupadd gluster
+
+Users from this group can run `glustercli` commands in GlusterFS
+servers.
+
+**Note**: Create "gluster" group before starting `glusterd2`
+
+## Starting and Stopping glusterd2 Manually
+
+This section describes how to start and stop glusterd2 manually
```
rishubhjain: Redundant, the title is self-explanatory.

doc/managing-the-glusterd2-service.md
harigowtham: yes. we have planned to give only a brief introduction. detailed introduction will be given as a link to the docs (https://docs.gluster.org/en/latest/) once the documetation for gd2 is available. The point i wanted to convey here is, adding the explanation about gd2 again and again will make this less user friendly because of repetition. so we can remove it here.

doc/managing-the-glusterd2-service.md
```diff
@@ -0,0 +1,54 @@
...
+storage servers non-disruptively.
+
+This section describes how to start the glusterd2 service in the
+following ways:
+
+**Note**: You must start glusterd2 on all GlusterFS servers.
+
+## Gluster user group
+
+On each server nodes, create "gluster" user group if not exists
```
phlogistonjohn: "On each server node" or "On all server nodes"

"if not not exists already" -> "if it does not already exist"

doc/managing-the-glusterd2-service.md
```diff
@@ -0,0 +1,54 @@
...
+already using the following command.
+
+    # groupadd gluster
+
+Users from this group can run `glustercli` commands in GlusterFS
```
phlogistonjohn: Nit: It seems more idomatic to me to say "on GlusterFS servers" rather than "in"

doc/managing-the-glusterd2-service.md
phlogistonjohn: It may make sense to adjust the title to read "Starting and Stopping glusterd2 Service Manually" and then there's less of a reason to mention binaries as systemd will be the most typical way to start and stop services on most of the systems where glusterd2 is installed (standalone).

Anyone running glusterd2 standalone with something that isn't systemd is kind of on their own IMO but they ought to be used to managing services with whatever they have anyway.

PR #1053, opened 2018-07-20
Snapshot auto mount bricks on glusterd restart 

rafikc30:
Snapshot bricks should automatically mount upon a glusterd restart. Especially after glusterd restart, mount won't be existing on the machine. This PR will mount the snapshot bricks during a glusterd restart.

### DIFF
--- a/glusterd2/brick/types.go
+++ b/glusterd2/brick/types.go
@@ -24,10 +24,10 @@ const (
 
 //MountInfo is used to store mount related information of a volume
 type MountInfo struct {
-	Mountdir   string
-	DevicePath string
-	FsType     string
-	MntOpts    string
+	BrickDirSuffix string
+	DevicePath     string
+	FsType         string
+	MntOpts        string
 }
 
 // Brickinfo is the static information about the brick
--- a/glusterd2/volume/fs_utils.go
+++ b/glusterd2/volume/fs_utils.go
@@ -11,8 +11,19 @@ import (
 	"os/exec"
 	"strings"
 	"syscall"
+	"time"
 
+	"github.com/gluster/glusterd2/glusterd2/brick"
+	"github.com/gluster/glusterd2/pkg/utils"
+
+	"github.com/pborman/uuid"
+	log "github.com/sirupsen/logrus"
 	config "github.com/spf13/viper"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	volumeIDXattrKey = "trusted.glusterfs.volume-id"
 )
 
 //For now duplicating SizeInfo till we have a common package for both brick and volume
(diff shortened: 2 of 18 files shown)


glusterd2/volume/fs_utils.go
```diff
@@ -144,3 +154,155 @@ func GetMounts() ([]*Mntent, error) {
...
+//MountDirectory will mount the bricks to the given path
+func MountDirectory(mountPath string, mountData brick.MountInfo) error {
+	err := utils.ExecuteCommandRun("mount", "-o", mountData.MntOpts, mountData.DevicePath, mountPath)
+	// Use syscall.Mount command to mount the bricks
+	return err
+}
+
+//StopBrick terminate the process and umount the brick directory
+func StopBrick(b brick.Brickinfo, logger log.FieldLogger) error {
+	var err error
```
rishubhjain: TRIVIAL:  IMO this declaration is not required, you can directly use err := b.Terminate.. and err:= UmountBrick.

Also you can directly use ` return UmountBrick(b)`

glusterd2/volume/fs_utils.go
```diff
@@ -144,3 +154,155 @@ func GetMounts() ([]*Mntent, error) {
...
+	for j := 0; j < 3; j++ {
+
+		err = UmountBrickDirectory(b.Path[:length])
+		if err == nil {
+			break
+		}
+		time.Sleep(3 * time.Second)
+	}
+	return err
+
```
rishubhjain: TRIVIAL: Unwanted line

glusterd2/volume/fs_utils.go
```diff
@@ -144,3 +154,155 @@ func GetMounts() ([]*Mntent, error) {
...
+}
+
+//UmountBrick will umount the brick directory
+func UmountBrick(b brick.Brickinfo) error {
+	//TODO Validate mount point before umount
+	var err error
+
+	length := len(b.Path) - len(b.MountInfo.BrickDir)
+	for j := 0; j < 3; j++ {
+
```
rishubhjain: TRIVIAL: Is this line required?

glusterd2/volume/fs_utils.go
```diff
@@ -144,3 +154,155 @@ func GetMounts() ([]*Mntent, error) {
...
+		return nil
+	}
+
+	if err := os.MkdirAll(mountRoot, os.ModeDir|os.ModePerm); err != nil {
+		log.WithError(err).Error("Failed to create snapshot directory ", brickinfo.String())
+		return err
+	}
+
+	if err := MountDirectory(mountRoot, brickinfo.MountInfo); err != nil {
+		log.WithFields(log.Fields{"error": err.Error(),
```
rishubhjain: Its better to log error using WithError. `log.WithError(err).WithFields...`

glusterd2/volume/fs_utils.go
```diff
@@ -144,3 +154,155 @@ func GetMounts() ([]*Mntent, error) {
...
+			"brickPath": brickinfo.String(),
+			"mountRoot": mountRoot}).Error("Failed to mount snapshot directory")
+
+		return err
+	}
+
+	if err := unix.Setxattr(brickinfo.Path, volumeIDXattrKey, vol.ID, 0); err != nil {
+		log.WithFields(log.Fields{"error": err.Error(),
```
rishubhjain: Its better to log error using WithError. log.WithError(err).WithFields...

glusterd2/commands/volumes/volume-create.go
```diff
@@ -47,7 +47,11 @@ func validateVolCreateReq(req *api.VolCreateReq) error {
 		return errors.New("invalid Volume Size, Minimum size required is " + strconv.Itoa(minVolumeSize))
 	}
 
-	if req.Size == 0 && len(req.Subvols) <= 0 {
+	if req.SelfProvision && req.Size == 0 {
+		return errors.New("volume size should provide to perform auto create")
+	}
+
+	if !req.SelfProvision && len(req.Subvols) <= 0 {
```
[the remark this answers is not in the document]
rafikc30: Since you requested for smartvol, given bricks should omit or we should have the capability of having both. But I will let @aravindavk answer the question as it is not specific to this PR.

glusterd2/volume/fs_utils.go
```diff
@@ -144,3 +155,148 @@ func GetMounts() ([]*Mntent, error) {
...
+			break
+		}
+	}
+	return false
+}
+
+//UmountBrickDirectory does an umount of the path
+func UmountBrickDirectory(path string) error {
+	err := syscall.Unmount(path, syscall.MNT_FORCE)
+	return err
```
Madhu-1: no need to create `err` variable you can return directly from here

glusterd2/volume/fs_utils.go
```diff
@@ -144,3 +155,148 @@ func GetMounts() ([]*Mntent, error) {
...
+			"xattr": volumeIDXattrKey}).Error("setxattr failed")
+		return err
+	}
+
+	return nil
+}
+
+//MountDirectory will mount the bricks to the given path
+func MountDirectory(mountPath string, mountData brick.MountInfo) error {
+	err := utils.ExecuteCommandRun("mount", "-o", mountData.MntOpts, mountData.DevicePath, mountPath)
```
Madhu-1: same here

glusterd2/brick/types.go
```diff
@@ -24,10 +24,10 @@ const (
 
 //MountInfo is used to store mount related information of a volume
 type MountInfo struct {
-	Mountdir   string
-	DevicePath string
-	FsType     string
-	MntOpts    string
+	BrickDirSuffix string
```
prashanthpai: What does `BrickDirSuffix` mean ? What's wrong with `Mountdir` ?

The `MountInfo` structure is analogous to the output of mount command.

plugins/device/deviceutils/store-utils.go
```diff
@@ -157,3 +158,19 @@ func UpdateDeviceFreeSize(peerid, vgname string) error {
 
 	return nil
 }
+
+//IsVgExist check whether the given vg exist in the device list for the local peer
```
prashanthpai: s/check/checks

glusterd2/brick/types.go
rafikc30: The mountdir used to store directory name after mount root. for example if the directory is mounted as /bricks/gluster and the brick path is /bricks/gluster/b1 then we store /b1 as mountdir . This name was causing confusion with mounted directory ie /bricks/gluster

aravindavk:
I see the review comments are addressed. I will merge this PR

PR #1056, opened 2018-07-23
add version command  in gluster CLI

Madhu-1:
version command prints glusterd2 version and glustercli version

```
 ./build/glustercli version
glusterd version: v4.1.0-106.gitd31eef4
glusterd api-version: 1
gluster-cli version: v4.1.0-106.gitd31eef4

```


Madhu-1:
@aravindavk please review

Madhu-1:
retest this please

glustercli/cmd/root.go
```diff
@@ -96,6 +97,19 @@ var (
...
 
+var versionCmd = &cobra.Command{
+	Use:   "version",
+	Short: "Print glusterd and gluster-cli version",
+	Run: func(cmd *cobra.Command, args []string) {
+		gd, err := client.Version()
+		if err == nil {
+			fmt.Println("glusterd version:", gd.GlusterdVersion)
+			fmt.Println("glusterd api-version:", gd.APIVersion)
+		}
```
phlogistonjohn: I'd like to see an else condition here that prints something like "unable to fetch server version" or something along those lines. If the command fetches the server version by default I don't think we ought to hide a failure condition if the client does not successfully get the server's version.

That said, it may also make sense to add a -S/--no-server flag for the subcommand so that a user who does not have a server to query (yet) can still get the version of the tool itself.

IOW, don't hide the error condition but provide the user a way to avoid hitting that error condition when they know better.

prashanthpai:
The `--version` should only print the CLI version. This does not involve any network call.

Getting the glusterd version should be a separate flag (perhaps `--server-version` or `glusterd-version` ?) and that should only print the glusterd version.

glustercli/cmd/root.go
Madhu-1: I can add else statement to the print error message only verbose flag is sent and discard printing gluster-cli version. IMO, it's not a good idea to print both error message and an actual response.

adding ```--no-server``` flag makes sense to discard error message while fetching server version fails.

@aravindavk  @prashanthpai, want to know your thoughts on this

kshlm:
> The --version should only print the CLI version. This does not involve any network call.
> 
> Getting the glusterd version should be a separate flag (perhaps --server-version or glusterd-version ?) and that should only print the glusterd version.

k8s and etcd version commands print both client and server versions. Following them isn't a bad idea. `kubectl version` fails if a k8s instance isn't configured, but `etcdctl version` prints out the cli and api versions, and the server version if a server is configured and available. I'd like it if `glustercli version` follows `etcdctl version`.

Madhu-1:
@kshlm, currently it follows the same mechanism, if the glustercli is not able to fetch the server version, it will print only glustercli version.

if fetching server version successful, it prints both versions.

* glusterd server is available
```
./build/glustercli version
glusterd version: v4.1.0-106.gitd31eef4
glusterd api-version: 1
gluster-cli version: v4.1.0-106.gitd31eef4
```

* glusterd server is unavailable
```
./build/glustercli version
gluster-cli version: v4.1.0-106.gitd31eef4
```

prashanthpai:
> I'd like it if glustercli version follows etcdctl version.

+1

I don't see `etcdctl` printing server version though:

```sh
[ppai@gd2-1 glusterd2]$ etcdctl --version
etcdctl version: 3.2.4
API version: 2
[ppai@gd2-1 glusterd2]$ etcdctl --version --endpoints=[127.0.0.1:2379]
etcdctl version: 3.2.4
API version: 2
```

The API version isn't fetched from etcd server.

kshlm:
> ./build/glustercli version
> gluster-cli version: v4.1.0-106.gitd31eef4

@Madhu-1 also print out the API version that the CLI supports as well. And in the message use 'glustercli' instead of 'gluster-cli'.

glustercli/cmd/root.go
```diff
@@ -96,6 +97,19 @@ var (
 	flagTimeout    uint
 )
 
+var versionCmd = &cobra.Command{
+	Use:   "version",
+	Short: "Print glusterd and gluster-cli version",
```
rafikc30: Do we need the argument validation here (number of argument)?

Madhu-1: @rafikc30 I will be updating this PR. Thinking to remove version as an argument.

PR #1059, opened 2018-07-24
Implement Delete all command for snapshot cli

rafikc30:
Delete all command is very important for snapshot. Because the number of snapshots can grow up to 256 for a volume. So deleting one by one is a tedious job.

This pull request implements commands to delete all snapshot or all snapshots of a particular volume.


rafikc30:
retest this please

rafikc30:
retest this please

glustercli/cmd/snapshot-delete.go
```diff
@@ -8,8 +8,10 @@ import (
 )
 
 const (
-	snapshotDeleteHelpShort = "Delete a Gluster Snapshot"
-	snapshotDeleteHelpLong  = "Delete a Gluster snapshot of the given name."
+	snapshotDeleteHelpShort    = "Delete snapshots"
+	snapshotDeleteHelpLong     = "It delete a snapshot if a snapshot name is given as an argument. It also takes command all to delete all snapshots in a cluster or all snapshots of a particular volume"
```
Madhu-1: `It delete a snapshot` to `It deletes a snapshot`

glustercli/cmd/snapshot-delete.go
```diff
@@ -8,8 +8,10 @@ import (
...
+	snapshotDeleteAllHelpShort = "Delete All snapshots"
+	snapshotDeleteAllHelpLong  = "Deletes all snapshots in the cluster. If volume name is not given. If volume name is given, then it deltes all snapshots of a volume"
```
Madhu-1: `deltes` to `deletes`

glustercli/cmd/snapshot-list.go
```diff
@@ -36,13 +37,22 @@ func snapshotListHandler(cmd *cobra.Command) error {
...
+			return nil
+		}
 		table.SetHeader([]string{"Name", "Origin Volume"})
 		for _, snap := range snaps {
 			for _, entry := range snap.SnapName {
 				table.Append([]string{entry, snap.ParentName})
 			}
 		}
 	} else {
+		if len(snaps[0].SnapName) == 0 {
```
Madhu-1: i think we need to add `index` validation here

rafikc30: Validation is not required here. Because if you have given volume name to list, then it is guaranteed to have one entry in the list. But that volume can have zero snapshots, hence the check.

glustercli/cmd/snapshot-delete.go
rafikc30: done

glustercli/cmd/snapshot-delete.go
rafikc30: done

PR #1068, opened 2018-07-25
Use opencensus grpc plugin to trace operations across glusterd2 nodes. 

### LINKED ISSUE #1049: GD2 Tracing: Add infrastructure to trace GD2 operations across gluster nodes.
Use opencensus plugins to trace GD2 calls across gluster nodes. Also, as part of this, add changes to trace the volume create calls across gluster nodes.

sseshasa:
Integrate the opencensus grpc plugin and add changes to trace volume create handler operations across the glusterd2 nodes.

Closes #1049


Madhu-1:
@sseshasa , I have a question here, I can see a lot of things are changes in ```Gopkg.lock``` file is this intended change for this PR.
example:
```
github.com/gorilla/mux  got updated from "v1.6.1" to "v1.6.2"
```

sseshasa:
@Madhu-1 This is one of the items I had mentioned in the stand-up. In order to compile after importing the opencensus grpc plugin, I had to make one change to the Gopkg.toml file as seen in the first commit. Apparently opencensus grpc plugin uses one of the later versions of grpc than what GD2 uses currently. As a result to keep all interdependencies sane a lot of packages including etcd and grpc are automatically updated in the Gopkg,lock file.

I am not sure if such huge jumps in versions across many packages are acceptable. This is something that needs to be discussed.

sseshasa:
@kshlm I will update etcd and grpc package versions in the Gopkg.toml file. In addition, there are other packages that have undergone minor version changes in the Gopkg.lock that are also mentioned in the toml file. Here are those packages,

github.com/gorilla/mux: v1.6.0 -> v1.6.2
github.com/spf13/cobra: v0.0.1 -> v0.0.3
github.com/spf13/pflag: v1.0.0 -> v1.0.1
github.com/spf13/viper: v1.0.0 -> v1.0.2
github.com/sirupsen/logrus: v1.0.3 -> v1.0.5
github.com/thejerf/suture: v2.0.0 -> v2.0.3

All the above have minor version changes. I think these need not be mentioned since the above have the "~" operator to cover the minor ranges. But just to confirm, should these be updated in the toml file as well?

kshlm:
> All the above have minor version changes. I think these need not be mentioned since the above have the "~" operator to cover the minor ranges. But just to confirm, should these be updated in the toml file as well?

Yup. These can remain the same.

PR #1110, opened 2018-08-02
added missing return 

Madhu-1:
avoid sending ```200 OK``` if get volume fails from ETCD


Madhu-1:

PR #1114, opened 2018-08-02
glusterd2: Allow setting config with environment variables

### LINKED ISSUE #1112: Accept config options as environment variables
The easiest way to pass config values when running GD2 in containers is to use environment variables. It allows us to keep the container image simple, and avoid having custom entrypoint scripts.

Having config environment variables will allow us to easily set up the external etcd endpoints when starting a container.

This is an easy enough fix. `spf13/viper` the package we use for managing configuration, supports env variables. We just need to enable it.

kshlm:
Closes #1112

### DIFF
--- a/glusterd2/config.go
+++ b/glusterd2/config.go
@@ -6,6 +6,7 @@ import (
 	"expvar"
 	"net"
 	"path"
+	"strings"
 
 	"github.com/gluster/glusterd2/glusterd2/gdctx"
 	"github.com/gluster/glusterd2/glusterd2/store"
@@ -32,13 +33,13 @@ var (
 	// defaultPathPrefix is set by LDFLAGS
 	defaultPathPrefix = ""
 
-	defaultlocalstatedir = defaultPathPrefix + "/var/lib/glusterd2"
-	defaultlogdir        = defaultPathPrefix + "/var/log/glusterd2"
-	defaultrundir        = defaultPathPrefix + "/var/run/glusterd2"
+	defaultlocalstatedir = path.Join(defaultPathPrefix, "/var/lib/glusterd2")
+	defaultlogdir        = path.Join(defaultPathPrefix, "/var/log/glusterd2")
+	defaultrundir        = path.Join(defaultPathPrefix, "/var/run/glusterd2")
 )
 
-// parseFlags sets up the flags and parses them, this needs to be called before any other operation
-func parseFlags() {
+// initFlags sets up the flags and parses them, this needs to be called before any other operation
+func initFlags() {
 	flag.String("localstatedir", defaultlocalstatedir, "Directory to store local state information.")
 	flag.String("rundir", defaultrundir, "Directory to store runtime data.")
 	flag.String("config", "", "Configuration file for GlusterD.")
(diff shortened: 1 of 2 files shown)


rishubhjain:
retest this please

glusterd2/config.go
```diff
@@ -117,7 +118,25 @@ func dumpConfigToLog() {
...
+// - defaults set using config.SetDefault
+// - flag defaults
+func initConfig() error {
+	// Use config given by flags
+	config.BindPFlags(flag.CommandLine)
+
+	// Allow config values from environment environment variables.
+	// All options settable from the command line are available to be set this way.
+	// The environment variable should be in uppercase, prefixed with "GD2" and have "-" replaced by "_" to be used.
+	config.AutomaticEnv()
```
prashanthpai: May be this doesn't matter - but I assume `AutomaticEnv` should come after setting the following two - `SetEnvPrefix` and `SetEnvKeyReplacer`, at least conceptually

glusterd2/config.go
```diff
@@ -135,11 +155,12 @@ func initConfig(confFile string) error {
 				Error("failed to read config file")
 			return err
 		}
+	} else {
+		config.AddConfigPath("/etc/glusterd2")
+		config.AddConfigPath("/usr/local/etc/glusterd2")
+		config.SetConfigName("glusterd2")
```
prashanthpai: Conceptually, shouldn't `SetConfigName` occur before `AddConfigPath` ?

Instead of hard-coding, you can derive the config path from `defaultPathPrefix` variable set during build time. We do not want those to pick up files from either of these.

aravindavk: default config file was removed previously. are we introducing that again?

prashanthpai: Ah, that's a good point @aravindavk 

I'd prefer that we don't. The `--config` shall be explicit argument in systemd service file anyway.

glusterd2/config.go
kshlm: This doesn't matter till a `config.Get` is done. But for the sake of conceptual consistency, I'll change the order.

glusterd2/config.go
kshlm: The order of `SetConfigName`, `AddConfigPath` or `SetConfigType` doesn't matter until `ReadInConfig` or `MergeInConfig` is called (which I've missed for the default case).

About having default search paths. I don't think there's any harm in having it. If the systemd-unit specifies a config file, we use it. And as we don't merge config files, there's no harm.

I'll correct the hardcoding of the paths.

glusterd2/config.go
```diff
@@ -126,20 +145,28 @@ func initConfig(confFile string) error {
...
+	}
+	if err := config.ReadInConfig(); err != nil {
+		// Ignore error if config file is not found, error out otherwise
+		if _, ok := err.(config.ConfigFileNotFoundError); ok {
 			log.WithError(err).
-				WithField("file", confFile).
-				Error("failed to read config file")
+				WithField("file", config.ConfigFileUsed()).
+				Warn("failed to load config from file")
```
aravindavk: both error message is same.

kshlm: One is a warning the other is an error. If a config file is found and we cannot parse it, it is an error. If we don't find a config file at all, we don't care.

PR #1118, opened 2018-08-02
Edit device

rishubhjain:
Getting device name from user in URL instead of request body during device edit.


Madhu-1:
@rishubhjain I have some questions related to device EDIT,
if the current device state and the requested device state doesn't match then we need to update device state

IMO if device state is already disabled we don't need to go and disable it again?

I think we are missing this check  in device EDIT API can you confirm on this?

rishubhjain:
@prashanthpai @aravindavk If the current state of the device is same as the requested state, then I am returning without updating etcd instead of throwing an error.

aravindavk:
What is the usecase for editing device name(what is device name?)?

plugins/device/rest.go
```diff
@@ -148,6 +148,12 @@ func deviceEditHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	deviceName := r.URL.Query().Get("device")
```
aravindavk: any issues while using requestBody? why query param?

Madhu-1:
why can't we return an error if device state is already set, currently if the volume is already started we are returning an error saying already started, why can't we follow the same approach here?
@rishubhjain @prashanthpai @aravindavk   need your thought on this.

prashanthpai:
> What is the usecase for editing device name(what is device name?)?

+1

plugins/device/rest.go
```diff
@@ -148,6 +148,12 @@ func deviceEditHandler(w http.ResponseWriter, r *http.Request) {
...
+	if deviceName == "" {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "Device name not provided in URL")
+		return
+	}
+
 	txn, err := transaction.NewTxnWithLocks(ctx, peerID)
```
Madhu-1: lock peer + device before editing it, so it won't be considered for brick/volume creation during device edit.

Madhu-1:
add TODO for

   * CLI
   * rest API
   * e2e test cases
   * endpoint update

rishubhjain:
@aravindavk @prashanthpai I have changed it to [device](https://github.com/gluster/glusterd2/blob/531301b57e2c049a2228494c6f25b14a4a76464d/plugins/device/api/req.go#L13).

plugins/device/rest.go
```diff
@@ -151,6 +151,12 @@ func deviceEditHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	device := r.URL.Query().Get("device")
```
aravindavk: Query parameters are mostly used for optional parameters. Please add device to URL itself or request body.

rishubhjain:
Fixes: https://github.com/gluster/glusterd2/issues/1283

atinmu:
@rishubhjain Ping! What's pending on this PR?

plugins/device/init.go
```diff
@@ -37,7 +37,7 @@ func (p *Plugin) RestRoutes() route.Routes {
 		route.Route{
 			Name:        "DeviceEdit",
 			Method:      "POST",
-			Pattern:     "/devices/{peerid}",
+			Pattern:     "/devices/{peerid}/edit",
```
aravindavk: `/devices/edit/{peerid}/{device:.*}`

plugins/device/rest.go
aravindavk: Use as part of url itself

plugins/device/rest.go
```diff
@@ -151,6 +151,12 @@ func deviceEditHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	device := mux.Vars(r)["device"]
```
aravindavk: Move this before Unmarshal

plugins/device/rest.go
aravindavk: lock peer+device similar to delete device

e2e/smartvol_ops_test.go
```diff
@@ -316,6 +317,41 @@ func testSmartVolumeDistributeDisperse(t *testing.T) {
...
+	var peerID string
+	for _, peer := range peerList {
+		deviceList, err = client.DeviceList(peer.ID.String())
+		if len(deviceList) > 0 {
+			peerID = peer.ID.String()
+			break
+		}
+	}
+
+	if len(deviceList) > 0 {
```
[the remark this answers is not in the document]
Madhu-1: as this is part of smart-volume test cases, we already have devices attached to the nodes, why can't we fail the test if there are not devices in attached to peers?

e2e/smartvol_ops_test.go
```diff
@@ -316,6 +317,41 @@ func testSmartVolumeDistributeDisperse(t *testing.T) {
...
+		device := deviceList[0]
+		if device.State == "enabled" {
+			err = client.DeviceEdit(peerID, device.Name, "disabled")
+			r.Nil(err)
+		} else if device.State == "disabled" {
+			err = client.DeviceEdit(peerID, device.Name, "enabled")
```
Madhu-1: I would like to see below-listed tests in this PR
* Disable device and check devices state is disabled
* when all devices are disabled, the volume creation should fail
* Enable all the devices, and check devices is enabled
* when all the devices are enabled, the volume creation should pass.

Madhu-1: @aravindavk ^^

pkg/restclient/device.go
```diff
@@ -15,3 +17,22 @@ func (c *Client) DeviceAdd(peerid string, device string) (deviceapi.AddDeviceRes
...
+// DeviceList lists the devices
+func (c *Client) DeviceList(peerid string) ([]deviceapi.Info, error) {
+	var deviceList deviceapi.ListDeviceResp
+	url := fmt.Sprintf("/v1/devices/%s", peerid)
+	err := c.get(url, nil, http.StatusOK, &deviceList)
+	return deviceList, err
+}
+
+// DeviceEdit edits devices
+func (c *Client) DeviceEdit(peerid string, device string, state string) error {
```
Madhu-1: `(peerid string, device string, state string)`  to `(peerid, device, state string)`

pkg/restclient/device.go
```diff
@@ -15,3 +17,22 @@ func (c *Client) DeviceAdd(peerid string, device string) (deviceapi.AddDeviceRes
...
+
+// DeviceList lists the devices
+func (c *Client) DeviceList(peerid string) ([]deviceapi.Info, error) {
+	var deviceList deviceapi.ListDeviceResp
+	url := fmt.Sprintf("/v1/devices/%s", peerid)
+	err := c.get(url, nil, http.StatusOK, &deviceList)
+	return deviceList, err
+}
+
+// DeviceEdit edits devices
```
Madhu-1: /devices to /device

plugins/device/rest.go
```diff
@@ -137,6 +138,16 @@ func deviceEditHandler(w http.ResponseWriter, r *http.Request) {
...
+	if device == "" {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "device not provided in URL")
+		return
+	}
+
+	if !strings.HasPrefix(device, "/") {
```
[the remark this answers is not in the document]
rishubhjain: there is a possibility that user provides extra slashes `/v1/devices/{peerid}//dev/vdb/`, to be on the safer side we can check for a slash before adding another slash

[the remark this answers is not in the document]
Madhu-1: please test this scenario, I think `//` will be converted into `/` by HTTP framework,

aravindavk:
Please squash all commits into one.

e2e/smartvol_ops_test.go
```diff
@@ -316,6 +317,61 @@ func testSmartVolumeDistributeDisperse(t *testing.T) {
 	checkZeroLvs(r)
 }
 
+func editDevice(t *testing.T) {
+	r := require.New(t)
+	peerList, err := client.Peers()
+	r.Nil(err)
```
Madhu-1: if peerList len is zero fail the test case.

e2e/smartvol_ops_test.go
```diff
@@ -316,6 +317,61 @@ func testSmartVolumeDistributeDisperse(t *testing.T) {
...
+	// create Distribute Replicate(2x3) Volume
+	createReq := api.VolCreateReq{
+		Name:               smartvolname,
+		Size:               40 * gutils.MiB,
+		DistributeCount:    2,
+		ReplicaCount:       3,
+		SubvolZonesOverlap: true,
+	}
+	_, err = client.VolumeCreate(createReq)
+	r.NotNil(err)
```
Madhu-1: enable the devices and create the volume, the volume creation should be successful.

e2e/smartvol_ops_test.go
rishubhjain: I think this should be done in the test case of peerList and not in the edit-device.

Madhu-1: as this is dependent on the peer list, no harm in having one more higher level check

Madhu-1:
>09:02:32     --- FAIL: TestSmartVolume/Edit_device (0.07s)
09:02:32     	require.go:765: 
09:02:32     			Error Trace:	smartvol_ops_test.go:374
09:02:32     			Error:      	Expected nil, but got: &errors.errorString{s:"volume not found"}
09:02:32     			Test:       	TestSmartVolume/Edit_device
@rishubhjain CI failure

Madhu-1:
@rishubhjain PR is having conflicts, please resolve.

rishubhjain:
retest this please

Issue #1119, opened 2018-08-02
Remove device API


rishubhjain:
W.r.t device management APIs discussed in https://github.com/gluster/glusterd2/issues/728, glusterd2 is yet to support removing device functionality. Users should be able to remove device after adding device.
This issue will track progress on remove device functionality.

atinmu:
Remove device functionality isn't a blocker for GCS/1.0 based on the revised MVP.

Issue #998, opened 2018-07-12
Concurrency support for dynamic Volume create API


aravindavk:
Problem:

Volume create API requests holds lock on Volume name till the transaction complete. If bricks auto provisioning is enabled then parallel Volume create requests can choose same device and one request will fail if the space is not available during transaction. 

Possible solution:

- Before choosing the devices for creating the brick, hold lock on all devices.
- Once the devices are selected, device available size will be updated by subtracting the to be allocated space. Along with this update, set a dirty flag with Volume name as identifier. Release lock on all devices after this(All the steps under the lock are in-memory or reading etcd data)
- On success Transaction, remove dirty flag(with lock).
- If Rollback is triggered, add available space again and clear dirty flag(with lock).

aravindavk:
Issue #851 is split into multiple issues. This is one among them

atinmu:
Not considered to be a blocker for GCS/1.0 based on the revised MVP.

PR #1120, opened 2018-08-02
API to remove device

rishubhjain:
Issue: https://github.com/gluster/glusterd2/issues/1119


plugins/device/init.go
```diff
@@ -48,6 +48,13 @@ func (p *Plugin) RestRoutes() route.Routes {
 			Version:      1,
 			ResponseType: utils.GetTypeString((*deviceapi.ListDeviceResp)(nil)),
 			HandlerFunc:  listAllDevicesHandler},
+		route.Route{
+			Name:        "DeviceDelete",
+			Method:      "DELETE",
+			Pattern:     "/devices/{peerid}",
```
Madhu-1: can we add a comment saying the device ID is a query param?

plugins/device/rest.go
```diff
@@ -167,6 +167,39 @@ func deviceEditHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	txn, err := transaction.NewTxnWithLocks(ctx, peerID)
+	if err != nil {
+		status, err := restutils.ErrToStatusCode(err)
+		restutils.SendHTTPError(ctx, w, status, err)
+		return
+	}
+	defer txn.Done()
+
+	err = deviceutils.DeleteDevice(peerID, deviceName)
```
Madhu-1: question: don't we need to lock device before deleting it?

plugins/device/rest.go
```diff
@@ -167,6 +167,39 @@ func deviceEditHandler(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		logger.WithError(err).WithField("peerid", peerID).Error("Failed to remove device")
+		status, err := restutils.ErrToStatusCode(err)
+		restutils.SendHTTPError(ctx, w, status, err)
+		return
+	}
+
+	restutils.SendHTTPResponse(ctx, w, http.StatusOK, nil)
```
Madhu-1: the response should be 204 for a delete operation

plugins/device/rest.go
Madhu-1: #998, as @aravindavk said we need to lock device for smart volume creation, the same way we should consider locking device before deleting it.

Madhu-1:
@rishubhjain I would suggest disabling the device first so it won't be considered for any volume/brick creation, later we can delete the device. just a question, how are we planning to move bricks from the device to another device before deleting it?

@aravindavk  @phlogistonjohn @raghavendra-talur @obnoxxx need input from your side on this

Madhu-1:
add TODO for
* CLI
* rest API
* e2e test cases
* endpoint update

atinmu:
@rishubhjain Ping! What's pending on this PR?

plugins/device/init.go
```diff
@@ -48,6 +48,15 @@ func (p *Plugin) RestRoutes() route.Routes {
 			Version:      1,
 			ResponseType: utils.GetTypeString((*deviceapi.ListDeviceResp)(nil)),
 			HandlerFunc:  listAllDevicesHandler},
+		// device name is passed as query param in url.
```
aravindavk: remove comment

pkg/restclient/device.go
```diff
@@ -1,17 +1,34 @@
...
+
+// DeviceDelete removes devices
+func (c *Client) DeviceDelete(peerid, device string) error {
+	device = strings.TrimLeft(device, "/")
+	url := fmt.Sprintf("/v1/devices/%s/%s", peerid, device)
+	return c.del(url, nil, http.StatusNoContent, nil)
+}
+
+// DeviceList lists the devices
+func (c *Client) DeviceList(peerid string) ([]deviceapi.Info, error) {
```
aravindavk: This is repeated in PR #1118

plugins/device/rest.go
```diff
@@ -170,6 +170,42 @@ func deviceEditHandler(w http.ResponseWriter, r *http.Request) {
 	restutils.SendHTTPResponse(ctx, w, http.StatusOK, nil)
 }
 
+func deviceDeleteHandler(w http.ResponseWriter, r *http.Request) {
```
aravindavk: sorry, I missed it earlier. We need to remove the device using `vgremove` and `pvremove` as a transaction step. Deleting device from etcd should be last step.

pkg/restclient/device.go
rishubhjain: Yes I required this for the e2e test cases, will rebase once that PR is merged.

rishubhjain:
retest this please

e2e/device_test.go
```diff
@@ -0,0 +1,155 @@
...
+	r.Nil(err)
+	dev, err = client.DeviceList(tc.gds[2].PeerID(), "/dev/gluster_loop3")
+	r.Nil(err)
+	r.Equal(dev[0].Device, "/dev/gluster_loop3")
+
+	t.Run("Edit device", editDevice)
+	t.Run("Delete device", testDeviceDelete)
+
+	// // Device Cleanup
+	r.Nil(loopDevicesCleanup(t))
```
Madhu-1: can we move this after device add? even if the device remove or device list fails, these block fo code won't get hit. it's good to do cleanup in `defer()`.
@aravindavk let me know your thoughts.

plugins/device/rest.go
```diff
@@ -194,3 +194,82 @@ func deviceEditHandler(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	deviceName := mux.Vars(r)["device"]
+	if deviceName == "" {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "Device name not provided in URL")
+		return
+	}
+
+	// Adding prefix(/) to device name
+	deviceName = "/" + deviceName
```
Madhu-1: am not able to find the check for whether the device is present or not and the device belongs to the requested Peer ID.

plugins/device/rest.go
```diff
@@ -194,3 +194,82 @@ func deviceEditHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	err = txn.Ctx.Set("device", &deviceName)
+	if err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
+		return
+	}
+
+	err = txn.Do()
+	if err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, "transaction to prepare device failed")
```
Madhu-1: even the device is not present in the peer is considered as internal server error, need to handle not found error also

plugins/device/transaction.go
```diff
@@ -59,3 +61,54 @@ func txnPrepareDevice(c transaction.TxnCtx) error {
...
+	}
+
+	// Remove VG
+	if err := lvmutils.RemoveVG(vgName); err != nil {
+		c.Logger().WithError(err).WithField("device", deviceName).Error("Failed to remove volume group")
+		return err
+	}
+
+	//Remove PV
+	if err := lvmutils.RemovePV(deviceName); err != nil {
```
Madhu-1: make use of `err` throughout the function. don't create a new variable for each block.

plugins/device/transaction.go
```diff
@@ -59,3 +61,54 @@ func txnPrepareDevice(c transaction.TxnCtx) error {
...
+
+	vgName := ""
+	for _, device := range devices {
+		if device.Device == deviceName {
+			vgName = device.VgName()
+		}
+	}
+
+	if vgName == "" {
+		return errors.New("No device found with given device name")
```
Madhu-1: an error message should start with the small case. and move this error message to error pkg. it will be helpful for the consumer.

plugins/device/transaction.go
```diff
@@ -59,3 +61,54 @@ func txnPrepareDevice(c transaction.TxnCtx) error {
...
+		return err
+	}
+
+	devices, err := deviceutils.GetDevices(peerID)
+	if err != nil {
+		return err
+	}
+
+	if len(devices) == 0 {
+		return errors.New("No devices added in the given peer")
```
Madhu-1: same here as well

plugins/device/rest.go
```diff
@@ -194,3 +194,82 @@ func deviceEditHandler(w http.ResponseWriter, r *http.Request) {
...
+			restutils.SendHTTPError(ctx, w, http.StatusNotFound, errors.ErrPeerNotFound)
+		} else {
+			restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, "failed to get peer details from store")
+		}
+		return
+	}
+
+	txn.Nodes = []uuid.UUID{peerInfo.ID}
+	txn.Steps = []*transaction.Step{
+		{
```
Madhu-1: @aravindavk  don't we need to check any volume present on the device before removing it?

plugins/device/rest.go
```diff
@@ -194,3 +194,82 @@ func deviceEditHandler(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, "transaction to prepare device failed")
+		return
+	}
+
+	restutils.SendHTTPResponse(ctx, w, http.StatusNoContent, nil)
+}
+
+func listAllDevicesHandler(w http.ResponseWriter, r *http.Request) {
+
```
Madhu-1: remove this extra line

plugins/device/rest.go
rishubhjain: Vgremove will fail if the device is being used. We don't wipe the device

e2e/device_test.go
rishubhjain: Should we move all the clean up task to defer? Like deleting and stopping volume etc.

plugins/device/rest.go
Madhu-1: no harm in checking any volume present on the device, so that user will get a meaning full message something like bricks present on the device, cannot delete the device,

plugins/device/transaction.go
```diff
@@ -59,3 +62,56 @@ func txnPrepareDevice(c transaction.TxnCtx) error {
...
+			vgName = device.VgName()
+		}
+	}
+
+	if vgName == "" {
+		return ErrDeviceNameNotFound
+	}
+
+	// Remove VG
+	if err = lvmutils.RemoveVG(vgName); err != nil {
```
Madhu-1: what happens when the VG is already deleted in the last execution and something failed in PV deletion below? 
before deleting the resource don't we need to check if the resource is available or not?

Madhu-1:
@rishubhjain please update the commit message with description

plugins/device/transaction.go
rishubhjain: @Madhu-1 Can you elaborate more on `whether resource is available or not`. Do you mean whether the device exist or in use or not?

Madhu-1: @rishubhjain during the device removal if the VG removal was successful and it got failed in PV remove stage, and if the user again tries to delete the device, it will fail with  VG  not found error right, how are we solving this problem?
dont we need to handle VG not found scenario and PV not found scenario?

aravindavk:
rebase required

plugins/device/rest.go
rishubhjain: Done

plugins/device/transaction.go
rishubhjain: @aravindavk Can you help me with this? the basic way would be to check pvs and vgs before removing to avoid issues. But it's not a clean solution, can you suggest a cleaner solution to handle this scenerio?

aravindavk: try to delete VG and continue if it is not exists error. But if Vg is not found make sure no Vg exists before removing PV.(If user manually deleted Vg and Created Vg with different name then gd2 may end up deleting, If PV fails when Vg exists then that is also Ok)

aravindavk: Also check PV not exists. If Pv not exists then remove only from store and return success

rishubhjain: This is only possible if user manually deletes PV right?

Issue #1085, opened 2018-07-29
peer state enablement API


atinmu:
Peer can be enabled/disabled for maintenance, if Peer state is disabled then that peer should not be used for Volume creation or for adding new entities. This seems to be a requirement coming from https://github.com/gluster/anthill . 

@aravindavk @jstrunk - please validate and add the missing details?

JohnStrunk:
Thoughts on operator/gd2 synchronization are at: https://github.com/gluster/anthill/pull/36

atinmu:
Not a blocker to GCS/1.0 based on the revised MVP.

PR #1121, opened 2018-08-02
[WIP] API to Enable/Disable peer

rishubhjain:
Issue: https://github.com/gluster/glusterd2/issues/1085

### DIFF
--- a/glusterd2/commands/peers/addpeer.go
+++ b/glusterd2/commands/peers/addpeer.go
@@ -111,6 +111,10 @@ func addPeerHandler(w http.ResponseWriter, r *http.Request) {
 		newpeer.Metadata[key] = value
 	}
 
+	if req.Disabled {
+		newpeer.Disabled = req.Disabled
+	}
+
 	//check if remotePeerAddress already present
 	found := utils.StringInSlice(remotePeerAddress, newpeer.PeerAddresses)
 	//if not found prepend the remotePeerAddress to peer details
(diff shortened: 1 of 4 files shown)


prashanthpai:
@rishubhjain 

The original use-case is to disable the peer only for the purpose of volume creation. A general `Disabled` field in peer object is too broad and might confuse users. Is there any other way to maintain this disabled state in devices/smartvol/DVP APIs ? Or can you come up with a better name than `disabled` ?

cc @aravindavk

aravindavk:
> Is there any other way to maintain this disabled state in devices/smartvol/DVP APIs ? Or can you come up with a better name than disabled ?

Devices already has this field. Peer state is required for peer maintenance activities.

prashanthpai:
If the `peer maintenance activities` are performed by operator (external entity), the onus is on the operator to store this either in it's own state/storage or store it as arbitrary value using glusterd2's peer metadata API as there's no actionable item/task related to this state.

prashanthpai:
cc @JohnStrunk

atinmu:
@JohnStrunk @prashanthpai did we manage to reach to a conclusion on this?

JohnStrunk:
This discussion happened while I was on vacation. Sorry @atinmu and @prashanthpai.

The operator is the admin's conduit to controlling the cluster. It must be possible to:
- Mark a node as "in maintenance" (aka "quiesced" in the anthill doc). While quiesced, that node shouldn't be used for IVP, nor should (automatic) migrations of bricks from that node be initiated as a consequence of the node being down. (I realize the auto migration isn't implemented.)
- Mark a node that should be decommissioned. Once marked, it should no longer be eligible for IVP, and a (to-be-written) plugin should migrate all the bricks to elsewhere in the cluster.

Whether we have a specific API call that sets a node as "disabled" or we just have a set of well-known node-level tags, I think the above are possible. The anthill doc provides a possible method of using labels to do the necessary handshaking to implement the above operations. The closest to the "disabled"  state proposed here is `shouldQuiesce` in that doc.

PR #1123, opened 2018-08-03
Bump dependencies version

prashanthpai:
* Bump up golang version required to 1.9.x
  Centos 7 is 1.9.4, latest stable release is 1.9.7
* etcd: Bump up etcd version required to 3.3.9
  We specifically need fixes related to auto compaction of etcd keyspace.
  These fixes are in an etcd version >= 3.3.3. Centos7 is still on 3.2.22 though.
* Bump up version of dep as we were using a really old version


scripts/install-reqs.sh
```diff
@@ -39,5 +39,27 @@ install_gometalinter() {
   rm -f "$GOBINDIR/$LINTER_TARBALL"
 }
 
+install_etcd() {
```
kshlm: We'd removed this! When?

Also, @sseshasa will be bumping up etcd version in Gopkg.toml in #1068.

kshlm:
retest this please

scripts/install-reqs.sh
prashanthpai: I"m not sure if it was present as elastic etcd was the default.

prashanthpai:
retest this please

Issue #1079, opened 2018-07-27
Batch storing volume info in etcd as single etcd txn


prashanthpai:
A glance at the following screenshot by @sseshasa reveals that `storeVolinfo` takes a lot of time:

![screenshot_20180725_141008](https://user-images.githubusercontent.com/36185324/43245169-f7c88aa6-90ca-11e8-976a-b76c82d9cb26.png)

The originator node txn step `storeVolinfo` took **68% of time taken** to complete entire volume operation.

```go
// storeVolInfo uses to store the volinfo based on key and to generate client volfile
func storeVolInfo(c transaction.TxnCtx, key string) error {
        var volinfo volume.Volinfo
        if err := c.Get(key, &volinfo); err != nil {
                c.Logger().WithError(err).WithField(
                        "key", "volinfo").Debug("Failed to get key from store")
                return err
        }

        if err := volume.AddOrUpdateVolumeFunc(&volinfo); err != nil {
                c.Logger().WithError(err).WithField(
                        "volume", volinfo.Name).Debug("failed to store volume info")
                return err
        }

        if err := volgen.Generate(); err != nil {
                c.Logger().WithError(err).WithField(
                        "volume", volinfo.Name).Debug("failed to generate volfiles")
                return err
        }

        return nil
}
```

We can batch all PUTs (including those from volgen) originating from `storeVolinfo` to etcd as a single etcd txn. May require some code changes there but totally worth it considering the disproportionate amount of time it takes.

aravindavk:
+1

May be Volgen is taking more time than  Volinfo, since each volfile is an entry in etcd. This time will grow when we have more volumes. We also need to enhance Volgen to generate incrementally instead of generating all volfiles.

aravindavk:
Volfiles are not stored in etcd, only one call made to save Volinfo to store. Closing this issue. Please reopen if this can be optimized further.

PR #1124, opened 2018-08-03
etcd txn: Add expvar counters to wrappers

prashanthpai:
...so that we can quickly know how many etcd operations are invoked
in a particular glusterd2 volume operation.

These expvar metrics can be seen using `/statedump` endpoint.

Updates https://github.com/gluster/glusterd2/issues/1079

### DIFF
--- a/glusterd2/store/store.go
+++ b/glusterd2/store/store.go
@@ -4,6 +4,7 @@ package store
 import (
 	"context"
 	"errors"
+	"expvar"
 	"fmt"
 	"os"
 	"sync"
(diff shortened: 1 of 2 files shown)


glusterd2/store/store.go
```diff
@@ -191,6 +191,7 @@ func Get(ctx context.Context, key string, opts ...clientv3.OpOption) (*clientv3.
 		defer cancel()
 	}
 
+	defer storeCounters.Add("get", 1)
```
rishubhjain: storeCounters not defined. Check failing because of the same.

prashanthpai:
retest this please

PR #1125, opened 2018-08-03
extras/nightly-container: Scripts to build nightly container

kshlm:
Builds containers using buildah and ansible, from nightly RPMs of
GlusterD2 and GlusterFS.

### DIFF
--- a/extras/nightly-container/build.sh
+++ b/extras/nightly-container/build.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+set -e
+
+## This scripts is used to build nightly containers out of nightly builds of GlusterD2 and GlusterFS.
+## Buildah and Ansible are used to build the container, and need to be installed on the host used.
+## Buildah >= 1.0.0
+## Ansible >= 2.4
+
+IMG_NAME="gluster/glusterd2-nightly"
+IMG_VERSION=$(date +%Y%m%d)
+
+echo "# Building $IMG_NAME:$IMG_VERSION"
+
+
+echo "## Setting up build container using buildah"
+# Base image on CentOS:7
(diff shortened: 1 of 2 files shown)


kshlm:
> but I have a basic question do we need to remove repo files added to yum.repo? or yum clean will remove it?

`yum clean all` cleans all the cached information from /var/cache/yum. The repo files being present does no harm, and could possibly help install debug packages in the container when required.

prashanthpai:
retest this please

extras/nightly-container/build.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
```
sac: Can we put a `set -e' at the top, so that the script fails in case of any errors
than continuing?

We do not do any error checks, so set -e should help in exiting as soon as some
error condition is hit.
For example for some reason if provision.yml is not present or errors out...

extras/nightly-container/provision.yml
```diff
@@ -0,0 +1,86 @@
+---
+- hosts: builder
+  gather_facts: no
+
```
sac: Can we put remote_user or become_method? But default if not run as root yum will
fail.

Or is it that by default it is run as root?

extras/nightly-container/build.sh
kshlm: Will do.

extras/nightly-container/provision.yml
kshlm: It's null by default, so I most likely picks the user the buildah container is being run as. I'll force it to be root.

prashanthpai:
Merging this as @sac's comments have been addressed.

Issue #1140, opened 2018-08-07
Failed to get volume status if one of the node glusterd2 is stopped


rmadaka:
-> Having 3 node glusterd2 cluster(n1, n2, n3) with external etcd running.
-> Create some volumes.
-> Then stop the glusterd2  on one of the node(n2) from 3 node cluster.
-> Then check for volume status, volume status is failing with below error.

glustercli volume status
Volume : Disp3
Error getting Volume status

Response headers:
X-Gluster-Cluster-Id: e56e67b7-51a3-4b22-8d5b-73679baafe91
X-Gluster-Peer-Id: 91b6edad-6246-4b77-9a0d-6bf92a6a1514
X-Request-Id: 47614cb2-5750-41d9-97c4-1b11c1a0aabe

Response body:
Transaction step bricks-status.Check failed on peer 5d9f6dbc-3231-4e78-93ba-148ed8fd6249 with error: rpc error: code = 14 desc = grpc: the connection is unavailable

Logs:

time="2018-08-07 15:36:49.271988" level=info msg="127.0.0.1 - - [07/Aug/2018:15:36:49 +0530] \"GET /v1/volumes HTTP/1.1\" 200 10699" reqid=743a5d1d-1ea5-4d24-ad4d-af7d1ca24930
time="2018-08-07 15:36:49.279308" level=error msg="failed RPC call" error="rpc error: code = 14 desc = grpc: the connection is unavailable" remotepeer="5d9f6dbc-3231-4e78-93ba-148ed8fd6249(dhcp35-33.lab.eng.blr.redhat.com)" reqid=f508d34b-5cdd-4e17-894a-8c02730c2e9a rpc=TxnSvc.RunStep source="[rpc-client.go:73:transaction.runStepOn]" txnid=d935ec41-512f-4cb2-8790-91dc84514e82
time="2018-08-07 15:36:49.279397" level=error msg="Step failed on node." error="rpc error: code = 14 desc = grpc: the connection is unavailable" node=5d9f6dbc-3231-4e78-93ba-148ed8fd6249 reqid=f508d34b-5cdd-4e17-894a-8c02730c2e9a source="[step.go:119:transaction.runStepFuncOnNodes]" step=bricks-status.Check txnid=d935ec41-512f-4cb2-8790-91dc84514e82
[...]

PR #1152, opened 2018-08-09
discard error message if node is unreachable and DontCheckAlive flag is set

Madhu-1:
if nodes are unreachable and `DontCheckAlive` flag is set, we need to discard
the error message and return nil

Fixes: #1140 

### DIFF
--- a/glusterd2/transaction/transaction.go
+++ b/glusterd2/transaction/transaction.go
@@ -13,6 +13,8 @@ import (
 	"github.com/coreos/etcd/clientv3/concurrency"
 	"github.com/pborman/uuid"
 	log "github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 )
 
 const (
@@ -133,6 +135,9 @@ func (t *Txn) Do() error {
 		}
 
 		if err := s.do(t.OrigCtx, t.Ctx); err != nil {
+			if t.DontCheckAlive && isNodeUnreachable(err) {
+				continue
+			}
 			expTxn.Add("initiated_txn_failure", 1)
 			if !t.DisableRollback {
 				t.Ctx.Logger().WithError(err).Error("Transaction failed, rolling back changes")
@@ -144,6 +149,18 @@ func (t *Txn) Do() error {
 
 	expTxn.Add("initiated_txn_success", 1)
 	return nil
+}
+
+func isNodeUnreachable(err error) bool {
+	unreachable := true
+	if s, ok := err.(*stepResp); ok {
+		for _, e := range s.Resps {
+			if grpc.Code(e.Error) != codes.Unavailable {
+				unreachable = false
+			}
+		}
+	}
+	return unreachable
 }
 
 // undo undoes a transaction and will be automatically called by Perform if any step fails.


Madhu-1:
logs
* before stopping node
```
[root@dhcp43-106 glusterd2]# ./build/glustercli volume status
Volume : testvol
+--------------------------------------+-----------------------------------+-------------+--------+-------+-------+
|               BRICK ID               |               HOST                |    PATH     | ONLINE | PORT  |  PID  |
+--------------------------------------+-----------------------------------+-------------+--------+-------+-------+
| c68051ee-b2c9-4178-8062-cdc75331deb9 | dhcp43-106.lab.eng.blr.redhat.com | /export1/v1 | true   | 49152 |  8800 |
| 20c93da6-1386-49f6-bb72-adf4d4144cac | dhcp43-209.lab.eng.blr.redhat.com | /export/v1  | true   | 49152 | 10580 |
+--------------------------------------+-----------------------------------+-------------+--------+-------+-------+
```
* After stopping node
```
[root@dhcp43-106 glusterd2]# ./build/glustercli volume status
Volume : testvol
+--------------------------------------+-----------------------------------+-------------+--------+-------+------+
|               BRICK ID               |               HOST                |    PATH     | ONLINE | PORT  | PID  |
+--------------------------------------+-----------------------------------+-------------+--------+-------+------+
| c68051ee-b2c9-4178-8062-cdc75331deb9 | dhcp43-106.lab.eng.blr.redhat.com | /export1/v1 | true   | 49152 | 8800 |
| 20c93da6-1386-49f6-bb72-adf4d4144cac | dhcp43-209.lab.eng.blr.redhat.com | /export/v1  | false  |     0 |    0 |
+--------------------------------------+-----------------------------------+-------------+--------+-------+------+
```

Madhu-1:
@prashanthpai  @aravindavk @rishubhjain  please review

atinmu:
Reporting a brick as offline which is hosted by a node where glusterd service is down seems to be wrong to me. This actually breaks the functionality where user might think that brick isn't running at all which might not be true. In GD1's world we don't display such brick information. We have received good number of feedbacks where the expectation is to highlight such bricks into "no information available" category.

glusterd2/transaction/transaction.go
```diff
@@ -133,6 +135,9 @@ func (t *Txn) Do() error {
 		}
 
 		if err := s.do(t.OrigCtx, t.Ctx); err != nil {
+			if t.DontCheckAlive && isNodeUnreachable(err) {
```
prashanthpai: Don't you have to continue here and not return ?

glusterd2/transaction/transaction.go
```diff
@@ -146,6 +151,18 @@ func (t *Txn) Do() error {
 	return nil
 }
 
+func isNodeUnreachable(err error) bool {
+	unreachable := true
+	if s, ok := err.(*stepResp); ok {
+		for _, e := range s.Resps {
```
prashanthpai: Just am observation:

Each step response will contain responses from multiple nodes. It is possible that, one among such nodes is down (unreachable error) and there's a genuine error on one of the other nodes. 

This "checkalive and ignore error" infra is only used by volume status command. In case of volume status, all or any error from one or more nodes shouldn't stop the transaction, not just node unreachable error. So we should continue running txn steps regardless of type of error.

For now, this solution is okay, for the purposes of volume status.

prashanthpai:
> Reporting a brick as offline which is hosted by a node where glusterd service is down seems to be wrong to me. This actually breaks the functionality where user might think that brick isn't running at all which might not be true. In GD1's world we don't display such brick information. We have received good number of feedbacks where the expectation is to highlight such bricks into "no information available" category.

Very valid point. However, this change is a bug fix for a different issue.  I'll file a separate issue for that.

glusterd2/transaction/transaction.go
Madhu-1: yes, we need to continue. 
added continue as moved function after `t.DisableRollback` check

glusterd2/transaction/transaction.go
Madhu-1: if anyone of the error messages is not equal to unreachable error, we are returning the error message

glusterd2/transaction/transaction.go
```diff
@@ -138,6 +140,9 @@ func (t *Txn) Do() error {
 				t.Ctx.Logger().WithError(err).Error("Transaction failed, rolling back changes")
 				t.undo(i)
 			}
+			if t.DontCheckAlive && isNodeUnreachable(err) {
```
prashanthpai: The earlier place was better. Checking it here will trigger undo. Isn't it ?

Madhu-1: yes, the earlier place was better, if an error is other than unreachable undo will get triggered.

PR #1153, opened 2018-08-09
Fix snapshot clone issue if snap mount dir is symlinked

aravindavk:
In my machine, `/var/run` is symlinked to `/run`. Snapshot bricks are
mounted in `/var/run` but `df` or `mount` output shows `/run`.

Snapshot clone fails with following error

    Transaction step snap-clone.Validate failed on peer \
        4e646eb6-0f51-4385-9892-a3a4ec7b1a67 with error: \
        mount point not found

This PR fixes the issue by comparing realpath with the mount command
output.

### DIFF
--- a/glusterd2/volume/volume-utils.go
+++ b/glusterd2/volume/volume-utils.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


Madhu-1:
retest this please

Issue #1128, opened 2018-08-03
return HTTP status code as part of response in rest client


Madhu-1:
whoever is using the restclient should come to know for what reason it failed (whether its bad request, internal server error or not found)

if I want to delete a volume, (if it fails)we may get different errors, the current restclient returns only the error message based on that am not able to take any further action, if we return status code from restclient it will be useful

prashanthpai:
This is already implemented AFAIK.

prashanthpai:
https://github.com/gluster/glusterd2/blob/master/pkg/restclient/common.go#L33-L38

Madhu-1:
AFAIK, we are returning the response error message to the caller, by this it will be hard to understand why it failed.

in case of volume delete
* does it failed because volume not found
* some internal server error
* txn locking failure etc

as per CSI if the volume is not found during the volume deletion driver should return back with success.

if we return HTTP status it will be easy further processing, as a caller, I don't need to decode error message and do the string comparison

prashanthpai:
The entire 'http.Response' is available to rest client consumer which will contain status code too. Currently it's only set just for error cases. That's straight-forward to fix.

prashanthpai:
This has been fixed already in PR #1154

PR #1154, opened 2018-08-09
restClient: return http.Response along with error from all Rest Client APIs

oshankkumar:
…t APIs

User of Rest Client may need Response Status Code or Response Headers in case
of failure. Returning http.Response from all rest API will help.

Fixes: #1128


oshankkumar:
@Madhu-1, Rest Client User may need resp header as well as resp status code, so instead of sending them separately it's good to send entire http.Response.

Issue #1045, opened 2018-07-19
Peer add is failing with incorrect error message


rmadaka:
-> Terminate the external etcd which is in running state
-> Then try to perform the peer add
-> Peer add is failing with error  peer is part of another cluster which is not expected
# glustercli peer add 10.70.42.78
Peer add failed

Error: Request failed with HTTP Status code 500
peer is part of another cluster

aravindavk:
Error is "peer is part of another cluster". Looks like setup and cleanup issue. Please provide more details about external etcd usage details and current configuration and node details.

prashanthpai:
> -> Terminate the external etcd which is in running state

In that case, how is the error not expected ? Just terminating external etcd isn't equivalent to peer remove/detach.

PR #1155, opened 2018-08-10
return an error when failed to connect to the store

Madhu-1:
during peer add, if the node is not able to connect to the store(ETCD).
return error message saying failed to connect to the store.

Fixes: #1045 

### DIFF
--- a/glusterd2/commands/peers/peer-rpc-svc.go
+++ b/glusterd2/commands/peers/peer-rpc-svc.go
@@ -53,13 +53,22 @@ func (p *PeerService) Join(ctx context.Context, req *JoinReq) (*JoinRsp, error) 
 
 	// TODO: Ensure no other operations are happening
 
-	peers, _ := peer.GetPeersF()
+	peers, err := peer.GetPeersF()
+	if err != nil {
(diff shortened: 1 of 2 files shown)


glusterd2/commands/peers/peer-rpc-svc.go
```diff
@@ -53,13 +53,22 @@ func (p *PeerService) Join(ctx context.Context, req *JoinReq) (*JoinRsp, error)
 
 	// TODO: Ensure no other operations are happening
 
-	peers, _ := peer.GetPeersF()
+	peers, err := peer.GetPeersF()
+	if err != nil {
+		logger.WithError(err).Error("failed to connect to store")
```
rishubhjain: @Madhu-1  Are you sure GetPeersF will return an error only if it fails to connect to store? unmarshalling error is also possible

Madhu-1: @rishubhjain we are not returning an error if [JSON unmarshal fails](https://github.com/gluster/glusterd2/blob/master/glusterd2/peer/store-utils.go#L133)

Issue #1170, opened 2018-08-14
Snapshot activate API fails if optional argument is skipped in the request


Shrivaibavi:
Steps:
1. I tried to activate the snapshot but it did not get activated.
[root@dhcp35-5 ~]# curl -i -X POST http://10.70.35.5:24007/v1/snapshot/snap3/activate
```
HTTP/1.1 422 Unprocessable Entity
Content-Type: application/json; charset=UTF-8
X-Gluster-Cluster-Id: f010c420-2797-46e5-b249-098eb1ec50f1
X-Gluster-Peer-Id: 8fa3e7e6-a5ac-47b2-9ba3-ca29997dbeb0
X-Request-Id: 0205014a-dc8a-4b36-87f2-a7e867f6c434
Date: Tue, 14 Aug 2018 10:38:01 GMT
Content-Length: 40

{"errors":[{"code":1,"message":"EOF"}]}
```
2. Then added optional argument, i could activate the snapshot successfully.

[root@dhcp35-5 ~]# curl -i -X POST http://10.70.35.5:24007/v1/snapshot/snap3/activate -d '{"force": false}'
```
HTTP/1.1 200 OK
Content-Type: application/json; charset=UTF-8
X-Gluster-Cluster-Id: f010c420-2797-46e5-b249-098eb1ec50f1
X-Gluster-Peer-Id: 8fa3e7e6-a5ac-47b2-9ba3-ca29997dbeb0
X-Request-Id: b2158462-b4a9-41b2-b497-3c82a6e90377
Date: Tue, 14 Aug 2018 10:41:37 GMT
Transfer-Encoding: chunked

[...]

PR #1171, opened 2018-08-14
snapshot activate fail if optional argument is skipped

rishubhjain:
Fixes : https://github.com/gluster/glusterd2/issues/1170


glusterd2/commands/snapshot/snapshot-create.go
```diff
@@ -696,7 +696,7 @@ func snapshotCreateHandler(w http.ResponseWriter, r *http.Request) {
 	httpStatus, err := unmarshalSnapCreateRequest(req, r)
 	if err != nil {
 		logger.WithError(err).Error("Failed to unmarshal snaphot create request")
-		restutils.SendHTTPError(ctx, w, httpStatus, err)
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, err)
```
Madhu-1: I think you can return only `err` from `unmarshalSnapCreateRequest`

doc/endpoints.md
```diff
@@ -1,95 +0,0 @@
-
```
prashanthpai: Why is this deleted ?

rishubhjain:
@aravindavk @prashanthpai Something went wrong, i have updated the PR.

PR #1172, opened 2018-08-14
glusterd2: Support setting Peer and Cluster IDs from the environment

### LINKED ISSUE #1113: Allow passing cluster-id (and possibly peer-id) as a config option
When using an external etcd instance, if we can pass a cluster-id as config option, we can avoid having to perform peer probes. This will be useful for GCS, as it will simplify deployment of GD2 on k8s.

With this in place, the deployment tool (operator or something else), will just need to start the GD2 pods with the right config options, to form a GD2 cluster. Without this, the deployment tool would need to perform an additional probe operation.

If we also support passing a peer-id, node/pod replacement, upgrades all become simpler still.

kshlm:
This is an extension to b8f9625, which makes it much easier to provision
and deploy GD2 clusters in a K8S.

With this patch GD2 reads GD2_PEER_ID and GD2_CLUSTER_ID from the
environment if available, and saved into LOCALSTATEDIR/uuid.toml.
If both environment variables and the uuid file have IDs, preference is
given to the environment variables, over the uuid file.

Closes #1113

### DIFF
--- a/glusterd2/gdctx/id.go
+++ b/glusterd2/gdctx/id.go
@@ -1,20 +1,33 @@
 package gdctx
 
 import (
+	"errors"
 	"expvar"
 	"io/ioutil"
 	"os"
 	"path"
+	"strings"
+	"sync"
 
 	"github.com/pborman/uuid"
 	toml "github.com/pelletier/go-toml"
-	log "github.com/sirupsen/logrus"
 	config "github.com/spf13/viper"
 )
 
+const (
+	uuidFileName    = "uuid.toml"
+	peerIDKey       = "peer-id"
+	clusterIDKey    = "cluster-id"
+	envPrefix       = "GD2"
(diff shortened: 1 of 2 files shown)


glusterd2/gdctx/id.go
```diff
@@ -24,74 +37,112 @@ var (
...
-const uuidFileName = "uuid.toml"
+func uuidFilePath() string {
+	return path.Join(config.GetString("localstatedir"), uuidFileName)
+}
 
-// UUIDConfig is a type that is read from and written to uuidFileName file.
-type UUIDConfig struct {
-	PeerID    string `toml:"peer-id"`
-	ClusterID string `toml:"cluster-id"`
+// uuidConfig is a type that is gives the configured values for peer and cluster ids
```
prashanthpai: nit: that gives

prashanthpai:
retest this please

Issue #1086, opened 2018-07-29
SelfHeal-Split-Brain Operations


vpandey-RH:
Original Issue : [#427](https://github.com/gluster/glusterd2/issues/427)
Implement Self-Heal Split-Brain Operations API and CLI .

volume heal <VOLNAME> split-brain {bigger-file <FILE> | latest-mtime <FILE> |source-brick <HOSTNAME:BRICKNAME> [<FILE>]}

Issue #427, opened 2017-11-15
Self Heal commands for Gd2 managed Volumes


aravindavk:
Collected the Heal commands details from AFR team.

## Volume Start
Start self heal daemon in nodes with atleast one brick which belong to replicate/EC/Arbiter volume.

## Volume Stop
If all Replicate/EC/Arbiter volumes are stopped then stop Self heal daemon.

## Enable/Disable
Remember the enable or disable state so that it will be used for other APIs. Note: Self heal daemons are not stopped with gd1, can we stop self heal daemons if it is disabled?

CLI:

    volume heal <VOLNAME> enable | disable

APIs:

    POST /v1/volumes/:name/heal/enable
    POST /v1/volumes/:name/heal/disable

## Trigger Heal

CLI:

    volume heal <VOLNAME> full

API:

    POST /v1/volumes/:name/heal

Glusterd will notify running self heal daemons. Have to get more details about the notification mechanism

## Heal Info
With Gd1, CLI is running another command as child process and displays stdout of the child process.

With Gd2, Glusterd2 will run the same CLI command and converts stdout as JSON response

CLI:

    volume heal <VOLNAME> statistics [heal-count [replica <HOSTNAME:BRICKNAME>]]
    volume heal <VOLNAME> info [summary | split-brain]

APIs:

    GET /v1/volumes/:name/heal

## Options

CLI:

    volume heal <VOLNAME> granular-entry-heal {enable | disable}

API:

    POST /v1/volumes/:name/heal/granular-entry/enable
    POST /v1/volumes/:name/heal/granular-entry/disable

## Split brain

CLI:

    volume heal <VOLNAME> split-brain {bigger-file <FILE> | latest-mtime <FILE> |source-brick <HOSTNAME:BRICKNAME> [<FILE>]}

atinmu:
## Checklist
- [ ] ReST APIs availibility in wiki
- [ ] CLI commands
- [ ] Sufficient logging
- [ ] Availibility of metrics data if applicable
- [ ] Xlator options table upto date?
- [ ] Test cases & test coverage of the functionality

itisravi:
The glfsheal binary which is used for getting heal info is also used for split-brain resolution. If resolving split-brain fails, it prints various error messages to stdout. In GD2, we would need to make sure that these are also captured and eventually propagated to the gluster CLI or the clients which interact with GD2 using the rest APIs.

aravindavk:
[...]

PR #1173, opened 2018-08-15
Split brain operations API and CLI

vpandey-RH:
Fixes [#1086](https://github.com/gluster/glusterd2/issues/1086)
Updates [#427](https://github.com/gluster/glusterd2/issues/427)

Implements Split Brain Operation API and CLI.


vpandey-RH:
@prashanthpai  @aravindavk  Please review.

pkg/restclient/glustershd.go
```diff
@@ -37,3 +37,39 @@ func (c *Client) SelfHeal(volname string, healType string) error {
 
 	return c.post(url, nil, http.StatusOK, nil)
 }
+
+// SelfHealSplitBrain sends request to start split-brain operations on a volume
+func (c *Client) SelfHealSplitBrain(volname string, operation string, args map[string]string) error {
+	var url string
+	switch operation {
+	case "bigger-file":
+		if _, ok := args["file"]; ok {
```
Madhu-1: as ` bigger-file` and `latest-mtime` having the same code
I think we can avoid code duplication in the case by using `fallthorugh`

plugins/glustershd/rest.go
```diff
@@ -201,3 +201,90 @@ func selfHealHandler(w http.ResponseWriter, r *http.Request) {
...
+	for index := range bricks {
+		if bricks[index].Hostname == hostname && bricks[index].Path == brickname {
+			return true
+		}
+	}
+	return false
+}
+
+func splitBrainOperationHandler(w http.ResponseWriter, r *http.Request) {
+	// Collect inputs from URL
```
Madhu-1: question: can we have a request body for this API instead of query parameters

plugins/glustershd/rest.go
```diff
@@ -201,3 +201,90 @@ func selfHealHandler(w http.ResponseWriter, r *http.Request) {
...
+			filename = val[0]
+			options = append(options, operation, brickPath, filename, "glusterd-sock", glusterdSockpath)
+		} else {
+			options = append(options, operation, brickPath, "glusterd-sock", glusterdSockpath)
+		}
+
+	}
+	_, err = runGlfshealBin(volname, options)
+	if err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, err)
```
Madhu-1: please a log here in case of error

e2e/glustershd_test.go
```diff
@@ -68,4 +72,129 @@ func TestSelfHeal(t *testing.T) {
...
+
+	if _, err := os.Lstat("/dev/fuse"); os.IsNotExist(err) {
+		t.Skip("skipping mount /dev/fuse unavailable")
+	}
+
+	mntPath := testTempDir(t, "mnt")
+	defer os.RemoveAll(mntPath)
+
+	host, _, _ := net.SplitHostPort(tc.gds[0].ClientAddress)
+	mntCmd := exec.Command("mount", "-t", "glusterfs", host+":"+volname, mntPath)
```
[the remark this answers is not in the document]
vpandey-RH: @Madhu-1  We are using exec.Command in quite a number of places. Not entirely sure if I should change this one. @prashanthpai What's your take ?

e2e/glustershd_test.go
```diff
@@ -68,4 +72,129 @@ func TestSelfHeal(t *testing.T) {
...
+	err = mntCmd.Run()
+	r.Nil(err, fmt.Sprintf("mount failed: %s", err))
+
+	f, err := os.Create(mntPath + "/file1.txt")
+	r.Nil(err, fmt.Sprintf("file creation failed: %s", err))
+	f.Close()
```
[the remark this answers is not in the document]
vpandey-RH: If there is an error in os.Create shouldn't the r.Nil check raise the error with an appropriate message there itself and terminate the test ?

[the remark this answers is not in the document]
Madhu-1: if its terminate we are good to go

glustercli/cmd/glustershd.go
```diff
@@ -29,6 +38,13 @@ func init() {
 	selfHealCmd.AddCommand(selfHealIndexCmd)
 	selfHealCmd.AddCommand(selfHealFullCmd)
 
+	selfHealSplitBrainCmd.Flags().BoolVar(&flagSplitBrainBiggerFile, "bigger-file", false, "Use bigger-file to resolve split-brain")
+	selfHealSplitBrainCmd.Flags().BoolVar(&flagSplitBrainLatestMtime, "latest-mtime", false, "Use latest-mtime to resolve split-brain")
+	selfHealSplitBrainCmd.Flags().BoolVar(&flagSplitBrainSourceBrick, "source-brick", false, "Use a brick as source to resolve split-brain")
+	selfHealSplitBrainCmd.Flags().StringVar(&flagFileName, "file", "", "Specify filename that is in split-brain")
+	selfHealSplitBrainCmd.Flags().StringVar(&flagBrickHostAndPath, "brick", "", "Specify hostname and brick path in format <hostname:brickpath>")
```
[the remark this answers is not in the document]
vpandey-RH: @Madhu-1  I guess glfsheal binary doesn't support peerid  as well in arguments ?

plugins/glustershd/rest.go
vpandey-RH: @Madhu-1 I guess having a request body will make it a lot more easier to use. Will make the changes.

vpandey-RH:
@Madhu-1  Addressed the review comments.

plugins/glustershd/rest.go
```diff
@@ -201,3 +201,93 @@ func selfHealHandler(w http.ResponseWriter, r *http.Request) {
...
+	// Check if volume is started
+	if volinfo.State != volume.VolStarted {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, gderrors.ErrVolNotStarted)
+		return
+	}
+
+	var options []string
+	glusterdSockpath := path.Join(config.GetString("rundir"), "glusterd2.socket")
+	switch operation {
+	case "bigger-file", "latest-mtime":
```
Madhu-1: can we use fallthrough here also

e2e/volume_ops_test.go
```diff
@@ -69,6 +69,10 @@ func TestVolume(t *testing.T) {
 	t.Run("DisperseMount", tc.wrap(testDisperseMount))
 	t.Run("DisperseDelete", testDisperseDelete)
 	t.Run("testShdOnVolumeStartAndStop", tc.wrap(testShdOnVolumeStartAndStop))
+
+	// Self Heal test
+	t.Run("SelfHeal", tc.wrap(testSelfHeal))
```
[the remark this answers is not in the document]
vpandey-RH: @Madhu-1 I am not sure if there is any advantage of having a self-heal test file seperately. I mean all the self heal testcases are in a seperate file than the rest of the volume operations which I think should be clear enough for others. So it wouldn't make any difference if we execute it under Volume operations or in a seperate category.

pkg/restclient/glustershd.go
```diff
@@ -37,3 +37,27 @@ func (c *Client) SelfHeal(volname string, healType string) error {
 
 	return c.post(url, nil, http.StatusOK, nil)
 }
+
+// SelfHealSplitBrain sends request to start split-brain operations on a volume
+func (c *Client) SelfHealSplitBrain(volname string, operation string, req shdapi.SplitBrainReq) error {
+	var url string
+	switch operation {
+	case "bigger-file", "latest-mtime":
```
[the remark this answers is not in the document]
vpandey-RH: @Madhu-1 Just a small question. Is there any advantage of using  fallthrough instead of multiple cases ?

vpandey-RH:
@Madhu-1  As far as using peerid is concerned I think IP is easier to use and prefered over peerid. Maybe we can have opinion from others as well, on this.

pkg/restclient/glustershd.go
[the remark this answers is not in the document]
Madhu-1: AFAIK functionality is same, but in glusterd2 we have fallthrough instead of multiple values in case

e2e/volume_ops_test.go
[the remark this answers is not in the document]
Madhu-1: testSelfHeal test cases are written in a different file, I feel code readability increases, and a number of lines in the volume_ops_test file in getting increased.

plugins/glustershd/rest.go
```diff
@@ -201,3 +201,95 @@ func selfHealHandler(w http.ResponseWriter, r *http.Request) {
...
+				restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "filename should be an absolute path in volume, should start with / notation")
+				return
+			}
+			options = append(options, operation, brickPath, req.FileName, "glusterd-sock", glusterdSockpath)
+		} else {
+			options = append(options, operation, brickPath, "glusterd-sock", glusterdSockpath)
+		}
+
+	}
+	_, err = runGlfshealBin(volname, options)
```
harigowtham: Don't we have to error out  when splitbrain resolution is already started? 
This is missed in the testcase as well. Can you please add that as well?

e2e/glustershd_test.go
```diff
@@ -188,3 +188,148 @@ func testGranularEntryHeal(t *testing.T, tc *testCluster) {
...
+
+	var prevKilledBrick string
+	getBricksStatus, err := client.BricksStatus(volname)
+	r.Nil(err, fmt.Sprintf("brick status operation failed: %s", err))
+
+	for brick := range getBricksStatus {
```
harigowtham: This iterates through both the bricks and kills them. and doesn't prevKilledBrick get assigned to the last brick in the loop? As we have only two bricks we can have one brick killed here in this loop and the another will be killed in the next loop. If you have better implementation, its welcome.

e2e/glustershd_test.go
```diff
@@ -188,3 +188,148 @@ func testGranularEntryHeal(t *testing.T, tc *testCluster) {
...
+	}
+
+	f1, err := os.OpenFile(mntPath+"/file1.txt", os.O_RDWR, 0777)
+	r.Nil(err, fmt.Sprintf("failed to open file: %s", err))
+	_, err = f1.WriteString("hello")
+	r.Nil(err, fmt.Sprintf("failed to write to file: %s", err))
+	f1.Sync()
+	defer f1.Close()
+
+	err = syscall.Unmount(mntPath, syscall.MNT_FORCE|syscall.MNT_DETACH)
```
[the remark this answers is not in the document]
vpandey-RH: @harigowtham  I added mount and unmount just to include the whole lifecycle of a volume in this testcase.

glustercli/cmd/glustershd.go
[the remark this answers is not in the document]
vpandey-RH: The CLI structure in glusterd is somewhat similar. I have added flagSplitBrainSourceBrick just to make sure which type of split resolution use wants and then an additional flag flagBrickHostAndPath to supply the source-brick.

plugins/glustershd/rest.go
```diff
@@ -201,3 +201,95 @@ func selfHealHandler(w http.ResponseWriter, r *http.Request) {
 	}
 	restutils.SendHTTPResponse(ctx, w, http.StatusOK, nil)
 }
+
+func checkBrickExistence(volinfo *volume.Volinfo, hostname string, brickname string) bool {
```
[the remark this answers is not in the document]
vpandey-RH: Sure, will add under utils.

plugins/glustershd/rest.go
vpandey-RH: Didn't get what you are trying to state here ?

glustercli/cmd/glustershd.go
```diff
@@ -113,3 +129,59 @@ var selfHealFullCmd = &cobra.Command{
...
+			if flagFileName != "" {
+				req.FileName = flagFileName
+			}
+			operation = "source-brick"
+		}
+		err := client.SelfHealSplitBrain(volname, operation, req)
+		if err != nil {
+			failure(fmt.Sprintf("Failed to resolve split-brain for volume %s\n", volname), err, 1)
+		}
+		fmt.Printf("Healed %s \n", flagFileName)
```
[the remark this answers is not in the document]
vpandey-RH: Sure.

e2e/glustershd_test.go
vpandey-RH: It won't iterate through both bricks as `break` will force to end the loop as soon as it finds one brick PeerID  matching.

harigowtham: sorry, my bad missed the break at the end.

plugins/glustershd/rest.go
harigowtham: Do we support issuing splitbrain bigger-file/lastest-mtime or any splitbrain command on a second file when the first file is already being resolved? 
If we don't support it we need to check if its already running and error out asking the user to wait.

vpandey-RH: @harigowtham Can't really do anything in here. The error has be raised in the glfsheal binary. Not entirely sure if this is the case in glfsheal right now.

plugins/glustershd/rest.go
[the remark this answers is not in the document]
vpandey-RH: Done

glustercli/cmd/glustershd.go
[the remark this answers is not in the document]
vpandey-RH: Done

e2e/glustershd_test.go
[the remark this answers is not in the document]
vpandey-RH: Done

glustercli/cmd/glustershd.go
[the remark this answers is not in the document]
harigowtham: Can we receive the brick host and path in the source brick itself, because IMO this one flag is enough to get the necessary arguments and also say what type of splitbrain resolution it is.

plugins/glustershd/rest.go
harigowtham: okay

vpandey-RH:
@harigowtham  Addressed the issue of multiple flags for source-brick operation PTAL.
@Madhu-1 Added support for peer-id as well in case of source-brick operation. PTAL

atinmu:
@harigowtham @Madhu-1 Please do a final round of review and take this to closure. This PR is being hanging around for almost 3 months now :-)

plugins/glustershd/rest.go
```diff
@@ -201,3 +203,95 @@ func selfHealHandler(w http.ResponseWriter, r *http.Request) {
...
+		}
+		options = append(options, operation, req.FileName, "glusterd-sock", glusterdSockpath)
+
+	case "source-brick":
+		if req.HostName == "" || req.BrickName == "" {
+			restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "please provide hostname and brickname")
+			return
+		}
+		if !volume.CheckBrickExistence(volinfo, req.HostName, req.BrickName) {
+			restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "brick doesn't exist")
```
harigowtham: Can we update this message as well mentioning about the host "brick doesn't exist in the provided host"? Something that doesn't confuse the user whether the brick is not in the host. or the brick or the host provided itself is wrong would be better. This might need some update on the CheckBrickExistence function. Being a minor change will enhance the user experience.

plugins/glustershd/rest.go
```diff
@@ -201,3 +203,95 @@ func selfHealHandler(w http.ResponseWriter, r *http.Request) {
...
+	volinfo, err := volume.GetVolume(volname)
+	if err != nil {
+		status, err := restutils.ErrToStatusCode(err)
+		restutils.SendHTTPError(ctx, w, status, err)
+		return
+	}
+
+	// Validate volume type
+	if !isVolReplicate(volinfo.Type) {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "invalid operation for this volume type")
```
harigowtham: Can you change the string to something like this "invalid operation: the volume is not a replicate volume"

glusterd2/volume/store-utils.go
```diff
@@ -218,3 +218,14 @@ func Exists(name string) bool {
 
 	return resp.Count == 1
 }
+
+// CheckBrickExistence checks if a brick is part of a particular volume
```
harigowtham: can you update the comment to something like "CheckBrickExistence checks if a brick is part of a host in a volume"
Want this to be mentioned as it checks against the hostname as well.

vpandey-RH:
@harigowtham  Nice suggestions. Made the changes.

glusterd2/volume/store-utils.go
```diff
@@ -218,3 +219,24 @@ func Exists(name string) bool {
...
+func CheckBrickExistence(volinfo *Volinfo, hostname string, brickname string) error {
+	bricks := volinfo.GetBricks()
+	hostFound := false
+	for index := range bricks {
+		if bricks[index].Hostname == hostname {
+			hostFound = true
+		}
+	}
+	if hostFound {
+		for index := range bricks {
```
Madhu-1: this can be
```
for _,b:=range bricks{
if b.Path == brickname && b.Hostname == hostname {
return nil
}```

File: glusterd2/volume/store-utils.go [hunk 19]
```diff
@@ -218,3 +219,24 @@ func Exists(name string) bool {
 
 	return resp.Count == 1
 }
+
+// CheckBrickExistence checks if a brick is part of a host in the volume
+func CheckBrickExistence(volinfo *Volinfo, hostname string, brickname string) error {
+	bricks := volinfo.GetBricks()
+	hostFound := false
+	for index := range bricks {
```
Madhu-1: This can be
```
for _,b:=range bricks{
if b.Hostname ==hostname{
hostFound=true
}
}
```

pkg/restclient/glustershd.go
```diff
@@ -37,3 +37,29 @@ func (c *Client) SelfHeal(volname string, healType string) error {
...
 }
+
+// SelfHealSplitBrain sends request to start split-brain operations on a volume
+func (c *Client) SelfHealSplitBrain(volname string, operation string, req shdapi.SplitBrainReq) error {
+	var url string
+	switch operation {
+	case "latest-mtime":
+		fallthrough
+	case "bigger-file":
+		if req.FileName != "" {
```
Madhu-1: add negative check first

```
if req.FileName == "" {
                    return errors.New("Please specify filename in split-brain")
}
url = fmt.Sprintf("/v1/volumes/%s/split-brain/%s", volname, operation)
```

same change is required in all case

plugins/glustershd/rest.go
```diff
@@ -11,11 +11,13 @@ import (
 	"strings"
 
 	"github.com/gluster/glusterd2/glusterd2/gdctx"
+	"github.com/gluster/glusterd2/glusterd2/peer"
 	restutils "github.com/gluster/glusterd2/glusterd2/servers/rest/utils"
 	"github.com/gluster/glusterd2/glusterd2/transaction"
 	"github.com/gluster/glusterd2/glusterd2/volume"
 	gderrors "github.com/gluster/glusterd2/pkg/errors"
 	glustershdapi "github.com/gluster/glusterd2/plugins/glustershd/api"
+	"github.com/pborman/uuid"
```
Madhu-1: add new line for import grouping

plugins/glustershd/rest.go
```diff
@@ -265,3 +267,95 @@ func selfHealHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	var req glustershdapi.SplitBrainReq
+	if err := restutils.UnmarshalRequest(r, &req); err != nil {
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, gderrors.ErrJSONParsingFailed)
+		return
+	}
+
+	// check if hostname is in form of uuid, if yes, convert uuid to respective ip addr
+	if uuid.Parse(req.HostName) != nil {
+		if peers, err := peer.GetPeer(req.HostName); err == nil {
```
Madhu-1: this can be 
```
if peers, err := peer.GetPeer(req.HostName); err != nil {
	restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, "couldn't find ip for the peerid")
        return
}
req.HostName = strings.Split(peers.PeerAddresses[0], ":")[0]

```

plugins/glustershd/rest.go
```diff
@@ -265,3 +267,95 @@ func selfHealHandler(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	var options []string
+	glusterdSockpath := path.Join(config.GetString("rundir"), "glusterd2.socket")
+	switch operation {
+	case "latest-mtime":
+		fallthrough
+	case "bigger-file":
+		if req.FileName != "" {
```
Madhu-1: do negative validation so we can increase code readability

plugins/glustershd/rest.go
```diff
@@ -265,3 +267,95 @@ func selfHealHandler(w http.ResponseWriter, r *http.Request) {
...
+			if !strings.HasPrefix(req.FileName, "/") {
+				restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "filename should be an absolute path in volume, should start with / notation")
+				return
+			}
+		} else {
+			restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "please provide filename")
```
Madhu-1: if its required move error message to error pkg, it will help the REST consumers to import error package compare error messages

Madhu-1: applicable for errors in all places

vpandey-RH:
@Madhu-1  Made the changes

pkg/errors/error.go
```diff
@@ -6,56 +6,63 @@ import (
...
+	ErrFetchingVolfileContent          = errors.New("unable to fetch volfile content")
+	ErrPidFileNotFound                 = errors.New("pid file not found")
+	ErrInvalidSnapName                 = errors.New("invalid snapshot name")
+	ErrInvalidClusterOption            = errors.New("Invalid cluster option key")
+	ErrInvalidVolFileTmplNamespace     = errors.New("invalid template namespace")
+	ErrInvalidVolFileTmplName          = errors.New("invalid template name")
+	ErrInvalidSplitBrainOp             = errors.New("Invalid split-brain operation specified")
+	ErrInvalidHostName                 = errors.New("hostname doesn't exist")
+	ErrInvalidBrickName                = errors.New("brick doesn't exist on this host")
+	ErrFilenameNotFound                = errors.New("Please specify filename for split-brain operation")
```
Madhu-1: error messages should start with a small case.

Madhu-1:
@aravindavk PTAL

Issue #1163, opened 2018-08-13
warning prompt for Yes/No should be there before deleting a volume 


ksandha:
A prompt for yes/no should be generating before deleting a volume.
```
[root@dhcp35-88 ~]# glustercli volume  delete newvol2
Volume newvol2 deleted successfully

```
```
[root@dhcp35-88 ~]# rpm -qa | grep glusterd2
glusterd2-4.1.0-143.git3c9ca84.el7.x86_64
```

PR #1181, opened 2018-08-16
cli: warning prompt yes/no before volume deletion

oshankkumar:
closes: #1163

### DIFF
--- a/glustercli/cmd/root.go
+++ b/glustercli/cmd/root.go
@@ -56,6 +56,7 @@ func addSubCommands(rootCmd *cobra.Command) {
 
 // GlustercliOption will have all global flags set during run time
 type GlustercliOption struct {
+	ScriptMode bool
 	XMLOutput  bool
 	JSONOutput bool
 	Insecure   bool
@@ -72,6 +73,7 @@ type GlustercliOption struct {
 //AddPersistentFlag will initialize the Global Flags of root command.
 func (gOpt *GlustercliOption) AddPersistentFlag(flagSet *pflag.FlagSet) {
 	// Global flags, applicable for all sub commands
+	flagSet.BoolVarP(&gOpt.ScriptMode, "script-mode", "y", false, "running in script mode")
 	flagSet.BoolVarP(&gOpt.XMLOutput, "xml", "", false, "XML Output")
 	flagSet.BoolVarP(&gOpt.JSONOutput, "json", "", false, "JSON Output")
 	flagSet.StringSliceVar(&gOpt.Endpoints, "endpoints", []string{"http://127.0.0.1:24007"}, "glusterd2 endpoints")
--- a/glustercli/cmd/utils.go
+++ b/glustercli/cmd/utils.go
@@ -2,8 +2,10 @@ package cmd
 
 import (
 	"errors"
+	"fmt"
 	"regexp"
 	"strconv"
+	"strings"
 )
 
 var (
@@ -53,4 +55,21 @@ func sizeToMb(value string) (uint64, error) {
 		size = sizeValue
 	}
 	return size, nil
+}
+
+func readString(prompt string, args ...interface{}) string {
+	var s string
+	fmt.Printf(prompt, args...)
+	fmt.Scanln(&s)
+	return s
+}
+
+// PromptConfirm prompts for confirmation
+func PromptConfirm(prompt string, args ...interface{}) bool {
+	switch strings.ToLower(readString(prompt, args...)) {
+	case "yes", "y":
+		return true
+	default:
+		return false
+	}
 }
--- a/glustercli/cmd/volume.go
+++ b/glustercli/cmd/volume.go
@@ -218,6 +218,11 @@ var volumeDeleteCmd = &cobra.Command{
 	Args:  cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		volname := cmd.Flags().Args()[0]
+		if !GlobalFlag.ScriptMode {
+			if ok := PromptConfirm("Are you sure you want to delete volume %s [yes/no]? ", volname); !ok {
+				return
+			}
+		}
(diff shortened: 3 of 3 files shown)


oshankkumar:
retest this please

Madhu-1:
@oshankkumar I think we should not provide this flag, if we provide warning prompt here, we should provide for all delete operations.
@aravindavk @prashanthpai @atinmu whats your thought on this?

aravindavk:
> I think we should not provide this flag, if we provide warning prompt here, we should provide for all delete operations as well.

Yes, required for all destructive operations. But it can be incremental, other commands can be added in different PRs.

Also in `gluster`, these prompts can be disabled using `--mode=script`(or something similar).

glustercli/cmd/volume.go
```diff
@@ -218,6 +219,9 @@ var volumeDeleteCmd = &cobra.Command{
 	Args:  cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		volname := cmd.Flags().Args()[0]
+		if ok := prompt.Confirm("Are you sure you want to delete volume %s [yes/no]", volname); !ok {
```
Madhu-1: question: why can't we implement this functionality in glusterd2 instead of importing external package for this?

Gopkg.lock
```diff
@@ -354,6 +360,12 @@
   packages = ["xdr2"]
   revision = "1a41d1a06c93bf8a0e0385be3847a277bb793187"
 
+[[projects]]
+  branch = "master"
```
Madhu-1: we should not manually edit `Gopkg.lock` file, need to add this dependency if required in `Gopkg.toml` file and regenerate this file.

glustercli/cmd/volume.go
oshankkumar: this doesn't require much functionality, we can easily implement this.

Gopkg.lock
oshankkumar: when I regenerate this file, it default revision which it took was not working. so I modified the revision with the current one.

Madhu-1: but we should not edit lock file, anyhow you are implementing the function for displaying prompt we don't require this one.

glustercli/cmd/utils.go
```diff
@@ -54,3 +55,22 @@ func sizeToMb(value string) (uint64, error) {
...
+	var s string
+	fmt.Printf(prompt+"? ", args...)
+	fmt.Scanln(&s)
+	return s
+}
+
+// PromptConfirm prompts for confirmation
+func PromptConfirm(prompt string, args ...interface{}) bool {
+	for {
+		switch readString(prompt, args...) {
```
Madhu-1: convert the `args` string to lower case of before switch statement

glustercli/cmd/root.go
```diff
@@ -72,6 +73,7 @@ type GlustercliOption struct {
 //AddPersistentFlag will initialize the Global Flags of root command.
 func (gOpt *GlustercliOption) AddPersistentFlag(flagSet *pflag.FlagSet) {
 	// Global flags, applicable for all sub commands
+	flagSet.BoolVarP(&gOpt.ScriptMode, "script-mode", "", false, "running in script mode")
```
Madhu-1: as we are not using `shorthand` change `BoolVarP` to `BoolVar`

Gopkg.lock
```diff
@@ -580,6 +580,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "6c5bb30b3fc2f361b83a245f51cd824ba49f4e46a125ca38702c8b561cd5dd83"
+  inputs-digest = "62a12e75f5497a585e2cf71c56e6129dbdec6d69886bb10e4346eb67008de949"
```
Madhu-1: why is this change required?

glustercli/cmd/utils.go
aravindavk: use `bufio.ReadString('\n')`

aravindavk: reader := bufio.NewReader(os.Stdin)
    resp, _ := reader.ReadString('\n')                                                                                                                                                        
    switch strings.TrimSpace(strings.ToLower(resp)) {                                                                                                                                         
    case "y", "yes":                                                                                                                                                                          
        return true                                                                                                                                                                           
    default:                                                                                                                                                                                  
        return false                                                                                                                                                                          
    }

glustercli/cmd/utils.go
```diff
@@ -54,3 +55,22 @@ func sizeToMb(value string) (uint64, error) {
...
+func readString(prompt string, args ...interface{}) string {
+	var s string
+	fmt.Printf(prompt+"? ", args...)
+	fmt.Scanln(&s)
+	return s
+}
+
+// PromptConfirm prompts for confirmation
+func PromptConfirm(prompt string, args ...interface{}) bool {
+	for {
```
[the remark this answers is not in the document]
oshankkumar: if  user has given some other input instead of yes/no then it will prompt again

Gopkg.lock
oshankkumar: this is autogenerated

aravindavk: You can revert this change, I think it is changed because of external lib used earlier

glustercli/cmd/utils.go
aravindavk: ignore my comment about bufio, `Scanln` also looks good

Gopkg.lock
oshankkumar: sure

oshankkumar:
retest this please

prashanthpai:
I'll wait for @Madhu-1's approval before merging

oshankkumar:
@Madhu-1 as suggested by @aravindavk , -y flag will become similar to dnf and other commands.

Issue #1180, opened 2018-08-16
volume metadata is not copied to snapshot response


Madhu-1:
### Observed behavior
snapshot response is having metadata and snaplist fields as `null`
### Expected/desired behavior
copy volume metadata to snapshot during snapshot create
if snaplist field is not required we should remove those fields

### Details on how to reproduce (minimal and precise)
1)create a snapshot from volume and check the response
```
 ],
        "metadata": null,
        "snap-list": null
    },
    "parentname": "vol1",
    "description": ""
```

### Information about the environment:

- Glusterd2 version used (e.g. v4.1.0 or master): master
- Operating system used: centos
- Glusterd2 compiled from sources, as a package (rpm/deb), or container: source
- Using External ETCD: (yes/no, if yes ETCD version): 
- If container, which container image: 
- Using kubernetes, openshift, or direct install: direct
- If kubernetes/openshift, is gluster running inside kubernetes/openshift or outside: 

### Other useful information

- glusterd2 config files from all nodes (default /etc/glusterd2/glusterd2.toml)
- glusterd2 log files from all nodes (default /var/log/glusterd2/glusterd2.log)
- ETCD configuration
- Contents of `uuid.toml` from all nodes (default /var/lib/glusterd2/uuid.toml)
- Output of `statedump` from any one of the node

### Useful commands

- To get glusterd2 version
    ```
    glusterd2 --version
    ```
- To get ETCD version
    ```
    etcd --version
    ```
- To get output of statedump
    ```
    curl http://glusterd2-IP:glusterd2-Port/statedump
    ```

PR #1182, opened 2018-08-17
snapshot: fix displaying null value in JSON response

Madhu-1:
we need to initialize map and slice to avoid displaying
as a `null` value in JSON response

Fixes: #1180 

### DIFF
--- a/glusterd2/commands/snapshot/snapshot-create.go
+++ b/glusterd2/commands/snapshot/snapshot-create.go
@@ -569,6 +569,9 @@ func createSnapinfo(c transaction.TxnCtx) error {
 
 	snapVolinfo.State = volume.VolCreated
 	snapVolinfo.GraphMap = volinfo.GraphMap
+	if snapVolinfo.GraphMap == nil {
+		snapVolinfo.GraphMap = make(map[string]string)
+	}
 	snapVolinfo.ID = uuid.NewRandom()
 	snapVolinfo.Name = req.SnapName
 	snapVolinfo.VolfileID = "snaps/" + req.SnapName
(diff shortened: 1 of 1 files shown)


glusterd2/commands/snapshot/snapshot-create.go
```diff
@@ -568,6 +568,7 @@ func createSnapinfo(c transaction.TxnCtx) error {
 	}
 
 	snapVolinfo.State = volume.VolCreated
+	snapVolinfo.GraphMap = make(map[string]string)
```
prashanthpai: This will be over-written below. `volinfo.GraphMap` may also be nil

glusterd2/commands/snapshot/snapshot-create.go
```diff
@@ -610,6 +611,9 @@ func duplicateVolinfo(vol, v *volume.Volinfo) {
 	v.DistCount = vol.DistCount
 	v.Type = vol.Type
 
+	v.Metadata = make(map[string]string)
+	v.Metadata = vol.Metadata
```
prashanthpai: Same here.

Madhu-1: `vol.Metadata` is already handled during volume creation time, this wont become nil

glusterd2/commands/snapshot/snapshot-create.go
Madhu-1: same here

prashanthpai: As per your comment, the values will be over-written (=) only by non-nil values, in which case there's no need for this change at all. Isn't it ?

PR #1194, opened 2018-08-23
Use glusterfs binary to mount

prashanthpai:
..as `mount.glusterfs` utility might be installed in non-standard
location in test environments.

Besides, mounts in real code (volume status) already uses the binary
to mount and not the mount utility script.

### DIFF
--- a/e2e/snapshot_ops_test.go
+++ b/e2e/snapshot_ops_test.go
@@ -316,6 +316,9 @@ func testRestoredVolumeMount(t *testing.T, tc *testCluster) {
 	err := mountVolume(host, snapTestName, mntPath)
 	r.Nil(err, fmt.Sprintf("mount failed: %s", err))
 
+	err = testMount(mntPath)
+	r.Nil(err)
+
 	umntCmd := exec.Command("umount", mntPath)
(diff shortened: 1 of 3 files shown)


prashanthpai:
retest this please

prashanthpai:
Unlike invoking `mount.glusterfs` as a command, `glusterfs` daemonizes itself. Will update this PR later with that fix.

amarts:
> Unlike invoking mount.glusterfs as a command, glusterfs daemonizes itself. Will update this PR later with that fix. 

Didn't understand what you are trying to say. Are you saying daemonizing doesn't work ? then recommend using `-N` option.

prashanthpai:
> Didn't understand what you are trying to say. Are you saying daemonizing doesn't work ? then recommend using -N option.

In go, `exec.Command.Run` starts the specified command and waits for it to complete. As glusterfs mount will be a daemon, we just had to use [exec.Command.Start](https://golang.org/pkg/os/exec/#Cmd.Start) instead.

> Was just thinking. While at it, can we at least run one more command on mount? Like 'touch' or at least stat ?

 :+1:

PR #1195, opened 2018-08-23
e2e tests: fix the mount command from any path

amarts:
So that gd2-tests can pass

Signed-off-by: Amar Tumballi <amarts@redhat.com>


amarts:
@prashanthpai

amarts:
Noticed that @prashanthpai also submitted the patch :-) @ #1194

PR #1197, opened 2018-08-24
Tracing: Trace volume list operation

### LINKED ISSUE #1196: Tracing: Trace "volume list" command.
Add instrumentation to trace the "volume list" command. Also add attribute to the high level span on
the number of volumes being listed to help debugging latency issues (if any) during the listing
operation.

### Observed behavior
NA

### Expected/desired behavior
span must also show the number of volumes being listed as part of the request on the tracing UI.

NA

- Operating system used: fedora 27
- Using External ETCD: (yes/no, if yes ETCD version): no
- If container, which container image: no
- Using kubernetes, openshift, or direct install: direct

### Other useful information

- ETCD configuration

### Useful commands

- To get glusterd2 version
    ```
    glusterd2 --version
    ```
- To get ETCD version
    ```
    etcd --version
    ```
- To get output of statedump
    ```
    ```

sseshasa:
Add instrumentation to trace volume list operation. Add the count of the
number of volumes listed by the command as an attribute in the top level
span of the trace.

Closes #1196

### DIFF
--- a/glusterd2/volume/store-utils.go
+++ b/glusterd2/volume/store-utils.go
@@ -11,6 +11,7 @@ import (
 	"github.com/coreos/etcd/clientv3"
 	"github.com/pborman/uuid"
 	log "github.com/sirupsen/logrus"
+	"go.opencensus.io/trace"
 )
 
 const (
@@ -116,8 +117,14 @@ func getFilterType(filterParams map[string]string) metadataFilter {
 
 //GetVolumes retrives the json objects from the store and converts them into
 //respective volinfo objects
-func GetVolumes(filterParams ...map[string]string) ([]*Volinfo, error) {
-	resp, e := store.Get(context.TODO(), volumePrefix, clientv3.WithPrefix())
+func GetVolumes(ctx context.Context, filterParams ...map[string]string) ([]*Volinfo, error) {
+	if ctx != context.TODO() {
+		var span *trace.Span
+		ctx, span = trace.StartSpan(ctx, "volume.GetVolumes")
+		defer span.End()
+	}
+
+	resp, e := store.Get(ctx, volumePrefix, clientv3.WithPrefix())
 	if e != nil {
 		return nil, e
 	}
@@ -169,7 +176,7 @@ func GetVolumes(filterParams ...map[string]string) ([]*Volinfo, error) {
 // belong to different volumes.
 func GetAllBricksInCluster() ([]brick.Brickinfo, error) {
 
-	volumes, err := GetVolumes()
+	volumes, err := GetVolumes(context.TODO())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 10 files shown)


glusterd2/commands/volumes/volume-create.go
```diff
@@ -100,7 +101,12 @@ func registerVolCreateStepFuncs() {
 func volumeCreateHandler(w http.ResponseWriter, r *http.Request) {
 
 	ctx := r.Context()
-	ctx, span := trace.StartSpan(ctx, "/volumeCreateHandler")
+	fnName := "?"
+	if pc, _, _, ok := runtime.Caller(0); ok {
```
Madhu-1: now we can move this function to utils and use it

sseshasa:
@Madhu-1 I addressed your comment. Please verify.

glusterd2/volume/store-utils.go
```diff
@@ -7,10 +7,13 @@ import (
 	"github.com/gluster/glusterd2/glusterd2/brick"
 	"github.com/gluster/glusterd2/glusterd2/store"
 	gderror "github.com/gluster/glusterd2/pkg/errors"
+	"github.com/gluster/glusterd2/pkg/utils"
 
 	"github.com/coreos/etcd/clientv3"
 	"github.com/pborman/uuid"
 	log "github.com/sirupsen/logrus"
+
```
Madhu-1: no need of extra line, as these packages are third-party libraries

sseshasa: Sure, will remove it.

glusterd2/commands/volumes/volume-create.go
```diff
@@ -100,7 +101,8 @@ func registerVolCreateStepFuncs() {
 func volumeCreateHandler(w http.ResponseWriter, r *http.Request) {
 
 	ctx := r.Context()
-	ctx, span := trace.StartSpan(ctx, "/volumeCreateHandler")
+	fnName := utils.GetFuncName()
+	ctx, span := trace.StartSpan(ctx, "/"+fnName)
```
kshlm: You really don't need to get the function name programatically here. You'd need the helper utility if `StartSpan` could be called dynamically at runtime, and you wouldn't know the callers before hand. But here you know the caller,  so you could just hard code the function name string whenever you call it. Just a simple `ctx, span := trace.StartSpan(ctx, "/volumeCreateHandler")` is good here. 
The same also applies to the other `StartSpan` calls.

sseshasa: I will revert it here and hard code it in the other places.

prashanthpai:
@sseshasa Volume listing operation doesn't really need much of tracing. It's best if tracing is first added to those operations that create a transaction.

PR #1199, opened 2018-08-24
restclient: applying tlsconfig to existing client Transport

oshankkumar:
applying tlsconfig to existing http.Client instead of
creating a new one.

### DIFF
--- a/pkg/restclient/common.go
+++ b/pkg/restclient/common.go
@@ -22,24 +22,18 @@ const (
 // ClientFunc receives a Client and overrides its members
 type ClientFunc func(*Client) error
 
-// WithTLSConfig overrides http.Client member with a client created
-// using specified TLS configuration.
+// WithTLSConfig applies tls config to underlying http.Client Transport
 func WithTLSConfig(tlsOpts *TLSOptions) ClientFunc {
 	return func(client *Client) error {
 		tlsConfig, err := NewTLSConfig(tlsOpts)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to create tlsconfig, err: %s", err.Error())
 		}
-		tr := &http.Transport{
-			DisableCompression: true,
-			DisableKeepAlives:  true,
-			TLSClientConfig:    tlsConfig,
+		if transport, ok := client.httpClient.Transport.(*http.Transport); ok {
+			transport.TLSClientConfig = tlsConfig
+			return nil
 		}
-		httpClient := &http.Client{
-			Transport: tr,
-		}
-		client.httpClient = httpClient
-		return nil
+		return fmt.Errorf("failed to apply tlsconfig on Transport : %T", client.httpClient.Transport)
 	}
 }
 
@@ -102,7 +96,7 @@ type Client struct {
 // to send request.
 func NewClientWithOpts(opts ...ClientFunc) (*Client, error) {
 	client := &Client{
-		httpClient: http.DefaultClient,
+		httpClient: defaultHTTPClient(),
 	}
 	for _, fn := range opts {
 		if err := fn(client); err != nil {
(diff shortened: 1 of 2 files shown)


oshankkumar:
WithTLSConfig creates a new http.Client everytime, instead of creating a new http.Client, we should only apply the tlsconfig to underlying client Transport.

oshankkumar:
@kshlm @Madhu-1 @prashanthpai @aravindavk Please review it

pkg/restclient/common.go
```diff
@@ -22,24 +22,18 @@ const (
...
 
-// WithTLSConfig overrides http.Client member with a client created
-// using specified TLS configuration.
+// WithTLSConfig applies tls config to underlying http.Client Transport
 func WithTLSConfig(tlsOpts *TLSOptions) ClientFunc {
 	return func(client *Client) error {
 		tlsConfig, err := NewTLSConfig(tlsOpts)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to create tlsconfig, err: %s", err)
```
Madhu-1: `err` is of type `error` for `%s` you need to convert use `err.Error()`

pkg/restclient/common.go
```diff
@@ -112,6 +106,16 @@ func NewClientWithOpts(opts ...ClientFunc) (*Client, error) {
 	return client, nil
 }
 
+func defaultHTTPClient() *http.Client {
+	tr := &http.Transport{
```
Madhu-1: why can't we use `http.DefaultTransport` and update required values in that?

pkg/restclient/common_test.go
```diff
@@ -36,3 +45,51 @@ func TestNewClientWithOpts_Error(t *testing.T) {
...
+		CaCertFile: certOut.Name(),
+	}
+
+	client, err := NewClientWithOpts(WithTLSConfig(tlsOpts))
+	r.Nil(err)
+
+	transport, ok := client.httpClient.Transport.(*http.Transport)
+	assert.True(ok)
+	assert.NotNil(transport.TLSClientConfig)
+
```
Madhu-1: remove extra line

pkg/restclient/common_test.go
```diff
@@ -36,3 +45,50 @@ func TestNewClientWithOpts_Error(t *testing.T) {
...
+	tlsOpts := &TLSOptions{
+		CaCertFile: certOut.Name(),
+	}
+
+	client, err := NewClientWithOpts(WithTLSConfig(tlsOpts))
+	r.Nil(err)
+
+	transport, ok := client.httpClient.Transport.(*http.Transport)
+	assert.True(ok)
```
Madhu-1: why we need to assert here, I think we can have `r.True(ok)`
instead of `assert.NotNil` you can use r.NotNil()

pkg/restclient/common.go
```diff
@@ -102,7 +96,9 @@ type Client struct {
 // to send request.
 func NewClientWithOpts(opts ...ClientFunc) (*Client, error) {
 	client := &Client{
-		httpClient: http.DefaultClient,
+		httpClient: &http.Client{
+			Transport: http.DefaultTransport,
```
Madhu-1: I think as a default configure you also need to set these values,
```
DisableCompression: true,
DisableKeepAlives:  true
```

oshankkumar: @aravindavk @prashanthpai  why we are disabling keep alives?  it will prevent user from re-using of TCP connections
between different HTTP requests. But we can reuse the TCP connection.

pkg/restclient/common.go
```diff
@@ -22,24 +22,18 @@ const (
...
 func WithTLSConfig(tlsOpts *TLSOptions) ClientFunc {
 	return func(client *Client) error {
 		tlsConfig, err := NewTLSConfig(tlsOpts)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to create tlsconfig, err: %s", err.Error())
 		}
-		tr := &http.Transport{
-			DisableCompression: true,
-			DisableKeepAlives:  true,
```
oshankkumar: @aravindavk @prashanthpai  why we are disabling keep alives?  it will prevent user from re-using of TCP connections
between different HTTP requests. But we can reuse the TCP connection.

prashanthpai: There's no good reason to do it in restclient level. I'm assuming it was there because CLI was the only consumer of restclient and that doesn't need to have keep alive connections. The CLI command usually does one HTTP request and exits.

PR #1201, opened 2018-08-24
Scripts: Add gnu-tar via brew during the install-reqs.

wildneuro:
During the install on MacOS (Darwin arch) found that `tar` utilized `--overwrite` feature, which is bundled in gnu tar.
This PR solves this behavior for Darwin platforms, as well as introduces script to make it easy to add other OS specific tools in the future.


prashanthpai:
add to whitelist

PR #1203, opened 2018-08-27
Tracing: Trace volume delete operation

### LINKED ISSUE #1202: Tracing: Trace "volume delete" command
Add instrumentation to trace volume delete operation. Add the requestID  and volume name as an attribute in the top level span of the trace.

### Observed behavior
NA

### Expected/desired behavior
span must also show the requestID and volume name as part of the request on the tracing UI.

- Operating system used: fedora 27
- Using External ETCD: (yes/no, if yes ETCD version):no
- If container, which container image: no

### Other useful information

- ETCD configuration

### Useful commands

- To get glusterd2 version
    ```
    glusterd2 --version
    ```
- To get ETCD version
    ```
    etcd --version
    ```
- To get output of statedump
    ```
    ```

sseshasa:
Add instrumentation to trace volume delete operation. Add the requestID
and volume name as an attribute in the top level span of the trace.

Closes #1202

### DIFF
--- a/glusterd2/commands/volumes/volume-delete.go
+++ b/glusterd2/commands/volumes/volume-delete.go
@@ -13,6 +13,8 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/pborman/uuid"
+
+	"go.opencensus.io/trace"
 )
 
 func deleteVolume(c transaction.TxnCtx) error {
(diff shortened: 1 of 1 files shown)


glusterd2/commands/volumes/volume-delete.go
```diff
@@ -45,6 +49,14 @@ func volumeDeleteHandler(w http.ResponseWriter, r *http.Request) {
 	logger := gdctx.GetReqLogger(ctx)
 	volname := mux.Vars(r)["volname"]
 
+	fnName := "?"
+	if pc, _, _, ok := runtime.Caller(0); ok {
```
Madhu-1: can we move this function to utils and use this function elsewhere

sseshasa: Do you mean factor out this chunk to utils?

	fnName := "?"
	if pc, _, _, ok := runtime.Caller(0); ok {
		fn := runtime.FuncForPC(pc)
		fnName = filepath.Base(fn.Name())
	}

Madhu-1: yes, because I see we are using this function in most of the places, 
to avoid code duplication I prefer moving this function to a utils package

PR #1211, opened 2018-09-03
gd2: fixed tlsmatcher to be used to multiplex tls connection

oshankkumar:
We should read tls version from handshake layer not from the record layer.
It is the Handshake layer version number that is important.

### DIFF
--- a/glusterd2/servers/rest/rest.go
+++ b/glusterd2/servers/rest/rest.go
@@ -69,7 +69,7 @@ func NewMuxed(m cmux.CMux) *GDRest {
 	keyfile := config.GetString("key-file")
 
 	if certfile != "" && keyfile != "" {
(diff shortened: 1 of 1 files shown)


oshankkumar:
@kshlm @prashanthpai @aravindavk please review this.

oshankkumar:
in the initial message tls version number is set as v1.0 in record layer, even if we are using tls v1.2, after tls negotiation version, tls version number is set as v1.2 in record layer. I am attaching Wireshark screenshot
wireshark-packet

oshankkumar:
@kshlm  Since in the initial packet, tls version in record layer is set as 1.0, and we are just reading initial packet record layer to verify tls version, I was facing an issue in communication between client and gd2 server when ssl is enabled.

oshankkumar:
added tlsv1.0 and tlsv1.1 matcher

Issue #1209, opened 2018-09-03
gd2 smoke: Crash in running e2e tests


prashanthpai:
This was seen in [glusterd2 smoke tests](https://build.gluster.org/job/gd2-smoke/) that run against glusterfs patches.

Seems like unmount is stuck for whatever reason. This isn't reproducible on my local centos7 machines or on glusterd2 centos CI.

**Log and stacktrace**:

```
09:23:44 SIGQUIT: quit
09:23:44 PC=0x4835d8 m=0 sigcode=0
09:23:44 
09:23:44 goroutine 1 [syscall, 9 minutes]:
09:23:44 syscall.Syscall6(0xf7, 0x1, 0x55f3, 0xc420819ae0, 0x1000004, 0x0, 0x0, 0x7f1458703850, 0x0, 0x0)
09:23:44 	/usr/lib/golang/src/syscall/asm_linux_amd64.s:44 +0x5 fp=0xc420819a88 sp=0xc420819a80 pc=0x4835b5
09:23:44 os.(*Process).blockUntilWaitable(0xc420018e10, 0xc420018e10, 0x0, 0x0)
09:23:44 	/usr/lib/golang/src/os/wait_waitid.go:31 +0xa5 fp=0xc420819b90 sp=0xc420819a88 pc=0x4a5e95
09:23:44 os.(*Process).wait(0xc420018e10, 0xc42082a020, 0xc4208340d8, 0x1)
09:23:44 	/usr/lib/golang/src/os/exec_unix.go:22 +0x42 fp=0xc420819c18 sp=0xc420819b90 pc=0x49f0c2
09:23:44 os.(*Process).Wait(0xc420018e10, 0x0, 0x0, 0xe5e178)
09:23:44 	/usr/lib/golang/src/os/exec.go:115 +0x2b fp=0xc420819c48 sp=0xc420819c18 pc=0x49e60b
09:23:44 os/exec.(*Cmd).Wait(0xc420834000, 0x0, 0x0)
09:23:44 	/usr/lib/golang/src/os/exec/exec.go:446 +0x62 fp=0xc420819cd0 sp=0xc420819c48 pc=0x73a412
09:23:44 os/exec.(*Cmd).Run(0xc420834000, 0x6, 0xc420819d98)
09:23:44 	/usr/lib/golang/src/os/exec/exec.go:289 +0x5c fp=0xc420819cf8 sp=0xc420819cd0 pc=0x73981c
[...]

PR #1215, opened 2018-09-04
Try force unmounting in e2e tests

prashanthpai:
...and add log so that we can see what's being unmounted.

Updates #1209 

### DIFF
--- a/e2e/utils_test.go
+++ b/e2e/utils_test.go
@@ -178,7 +178,8 @@ func cleanupAllBrickMounts(t *testing.T) {
 				continue
 			}
 
-			err = exec.Command("umount", parts[2]).Run()
+			testlog(t, fmt.Sprintf("cleanupAllBrickMounts(): umounting %s", parts[2]))
+			err = exec.Command("umount", "--force", "--lazy", parts[2]).Run()
 			if err != nil {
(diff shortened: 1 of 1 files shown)


prashanthpai:
I now know why the logs don't print on centos CI or on my local machine.

`txnCleanBricks` already would've unmounted the bricks and cleaned it up as part of volume delete operation. So there's nothing left for `loopDevicesCleanup`->`cleanupAllBrickMounts` to cleanup. Hence `cleanupAllBrickMounts` doesn't find any mounted volumes under `/tmp/gd2_func_test/` path.

If need be, we can later use force unmount in other places of e2e tests and also in code:

```diff
$ git diff
diff --git a/plugins/device/deviceutils/utils.go b/plugins/device/deviceutils/utils.go
index 7cc835d..44fe0b4 100644
--- a/plugins/device/deviceutils/utils.go
+++ b/plugins/device/deviceutils/utils.go
@@ -127,7 +127,7 @@ func BrickMount(dev, mountdir string) error {
 
 // BrickUnmount unmounts the Brick
 func BrickUnmount(mountdir string) error {
-       return utils.ExecuteCommandRun("umount", mountdir)
+       return utils.ExecuteCommandRun("umount", "--force", "--lazy", mountdir)
 }

```

For now, this limited change is okay. It's not fixing anything per se.

@aravindavk Can you please review and merge this one ?

Issue #1206, opened 2018-08-30
Volume expansion works even after providing wrong replica count


Akarsha-rai:
### Steps

1) Created replica 3 volume and started it
2) Tried to expand volume with wrong replica count, volume expanded successfully
```
[root@dhcp35-229 ~]# glustercli volume add-brick vol1 --replica=2 10.70.35.229:/bricks/brick1/vol11 10.70.35.163:/bricks/brick1/vol11 10.70.35.19:/bricks/brick1/vol11
vol1 Volume expanded successfully
```
Tried with rest-api's, same behaviour
```
[root@dhcp35-229 ~]# cat expandvol.json 
{
	"replicacount": 2,
        "bricks": [
	    {"peerid": "a2855277-c88c-4313-b8a8-2102f2d952b4", "path": "/bricks/brick5/vol11"},
            {"peerid": "6b200d2d-e606-4b41-ad80-211b13701bf3", "path": "/bricks/brick5/vol11"},
            {"peerid": "7f631335-8c92-4672-aa35-4a8bc521d66f", "path": "/bricks/brick5/vol11"}
        ]
}
```
### Observed behavior
Volume expands successfully even after providing wrong replica count.

### Expected/desired behavior
Volume expand should fail with appropriate error message.

### Details on how to reproduce (minimal and precise)
Steps mentioned in the summary

### Information about the environment:
```
Glusterd2 version used (e.g. v4.1.0 or master):
[root@dhcp35-229 ~]# rpm -qa | grep glusterd2
glusterd2-4.1.0-164.gite8ebe18.el7.x86_64

Operating system used:
[root@dhcp35-229 ~]# cat /etc/centos-release
CentOS Linux release 7.5.1804 (Core)

Glusterd2 compiled from sources, as a package (rpm/deb), or container:
package

Using External ETCD: (yes/no, if yes ETCD version):
yes, etcdmain: etcd Version: 3.3.8

If container, which container image:

Using kubernetes, openshift, or direct install:
direct install

If kubernetes/openshift, is gluster running inside kubernetes/openshift or outside:
```

### Other useful information

- glusterd2 config files from all nodes (default /etc/glusterd2/glusterd2.toml)
```
[root@dhcp35-229 ~]# cat /etc/glusterd2/glusterd2.toml 

localstatedir = "/var/lib/glusterd2"
logdir = "/var/log/glusterd2"
logfile = "glusterd2.log"
loglevel = "INFO"
[...]

PR #1218, opened 2018-09-04
Handle volume expansion for distributed replicate volumes

vpandey-RH:
Fixes [#1206](https://github.com/gluster/glusterd2/issues/1206)


vpandey-RH:
@aravindavk Please review.

vpandey-RH:
@prashanthpai @kshlm @Madhu-1 @rishubhjain  Would like to get your opinion on this.

So, currently we are handling volume expansion with three approaches.
1- If the newReplicaCount is greater than current replica count then Add the bricks to each of the existing subvolumes, thus making no of bricks in each subvolume equal to new replica count. but the number of bricks to be added should be equal to (newReplicaCount - currentReplicaCount) * no. of subvolumes

2- If newReplicaCount is equal to current replica count then just add a new subvolume conatining the bricks that need to be added (no. of bricks should be equal to replicaCount)

3- If the newReplicaCount < current Replica Count then just throw an error that the number of bricks are invalid. 

I wanted to add a fourth approach i.e a mix of the 1st and 2nd approach where newReplicaCount > currentReplicaCount and  if the number of bricks to be added are sufficient then add bricks to existing subvolumes and if we are left with more bricks i.e. if we are left with bricks in multiples of newReplicaCount we can form new subvolumes out of those bricks.

For e.g. - Let's say we have a volume with 2 subvols and replica - 2 and we send a brick add request with new replica count as 3 but we specify 5 bricks in the request. So, in this case what we have currently is we will be throwing  an error stating "Invalid number of bricks" according to 1st approach. 
But, I think we can add few lines of code that will actually add first 2 bricks to two subvolumes and then add other 3 bricks as a new subvolume. Thus, changing the replica count to 3 successfully and along with that we won't have a limitation on how many bricks user can specify in volume expand request.

aravindavk:
> I wanted to add a fourth approach i.e a mix of the 1st and 2nd approach where newReplicaCount > currentReplicaCount and if the number of bricks to be added are sufficient then add bricks to existing subvolumes and if we are left with more bricks i.e. if we are left with bricks in multiples of newReplicaCount we can form new subvolumes out of those bricks.

As discussed over chat, increasing the replica count will not increase the Volume size(calling API as Volume expand may mislead). Adding new sub volume will increase the volume size.

Mixing these two usecases may add confusion to users. My preference is to approach 1 if number of bricks provided are more than the required bricks to increase the replica count.

@prashanthpai @Madhu-1 @kshlm thoughts?

vpandey-RH:
We can actually replace approach 1 with approach 4 and leave rest of the code same.

prashanthpai:
@aravindavk @vpandey-RH 

How's the current behaviour in glusterd1 ?

vpandey-RH:
@prashanthpai  Gd1 implements the first approach.

atinmu:
@vpandey-RH What's the benefit of adding the 2nd approach into the equation? Is there an use case what we're trying to solve here?

Please note from GCS perspective, we may not eventually do any add-brick operations. Instead we just need to extend the space of the bricks by doing lv resize. At this moment, my take is not to complicate things and stick to the same functionality offerings.

prashanthpai:
@vpandey-RH We're open to future enhancements. Don't let this deter you. However, from prioritisation of near term goals, achieving glusterd1 behaviour parity is good enough for now. Feel free to open a separate issue for the enhancements you'd want to propose.

vpandey-RH:
@prashanthpai Sure.

prashanthpai:
@vpandey-RH If you can add e2e test case, as suggested by @aravindavk , we can get this forward.

vpandey-RH:
@prashanthpai I will add the test.

vpandey-RH:
@aravindavk @prashanthpai  Please review.

e2e/volume_ops_test.go
```diff
@@ -261,6 +261,53 @@ func testVolumeExpand(t *testing.T, tc *testCluster) {
...
+		ReplicaCount: 3,
+		Bricks: []api.BrickReq{
+			{PeerID: tc.gds[0].PeerID(), Path: brickPaths[8]},
+			{PeerID: tc.gds[1].PeerID(), Path: brickPaths[9]},
+		},
+		Flags: flags,
+	}
+
+	//expand with new brick dir which is not created
+	volinfo, err = client.VolumeExpand("TestVolumeExpand", expandReq)
```
Madhu-1: @vpandey-RH can you add some negative test scenarios also.

vpandey-RH: Sure

vpandey-RH: @Madhu-1 I have added a couple of negative cases.

glusterd2/commands/volumes/volume-expand-txn.go
```diff
@@ -40,12 +40,14 @@ func expandValidatePrepare(c transaction.TxnCtx) error {
...
-		if req.ReplicaCount < volinfo.Subvols[0].ReplicaCount {
-			return errors.New("invalid number of bricks")
-		} else if req.ReplicaCount == volinfo.Subvols[0].ReplicaCount {
-			return errors.New("replica count is same")
+	if volinfo.Type == volume.Replicate || volinfo.Type == volume.DistReplicate {
+		if req.ReplicaCount != 0 {
+			// TODO: Only considered first sub volume's ReplicaCount
+			if req.ReplicaCount < volinfo.Subvols[0].ReplicaCount {
+				return errors.New("invalid number of bricks")
+			} else if req.ReplicaCount == volinfo.Subvols[0].ReplicaCount {
```
harigowtham: if the replica count is the same as it was, then we are supposed to add a new subvol aren't we?

aravindavk: usability issue if we add new sub volume. If user specified replica count with some expectation, if we add sub volume instead of notifying user then it will be wrong.

PR #1227, opened 2018-09-06
Glustercli : Displaying volumes not found message if no volume exists

rishubhjain:
`glustercli volume status` does not display any message if no volume exist.

```
[root@localhost build]# ./glustercli volume status
[root@localhost build]
```
**Expected:**

```
[root@localhost build]# ./glustercli volume status
No volumes found
[root@localhost build]#
```


prashanthpai:
I don't see this as a bug.

atinmu:
> I don't see this as a bug. The log cleanups are welcome.

Not really at least if you compare this with GD1's behaviour.

root@995414497df8:/home/glusterfs# gluster v status
No volumes present

Issue #1018, opened 2018-07-16
Extend metadata based filtering to other operations


prashanthpai:
Currently, only volume and peer listing can be filtered based on metadata labels (key-value pairs). Similar to labels in k8s, we should support most of our volume operations to take labels/filters as input.

Example: Stop all volumes having specific keys and values:

```
glustercli volume stop all --key=datatype --value=thumbnail
``` 

```
glustercli volume stop all --key=workload --value=db
```

prashanthpai:
This might be a requirement from Gluster k8s operator.

cc @aravindavk

Issue #462, opened 2017-11-30
volume: provide an option to set a 'tag' 


amarts:
... (or many tags) on a volume while creating (or with volume update apis).

This tag can be used to display reduced set of volumes, like

 `gluster volume info --tag=something`

If we get each different layer integrating the API to set their own tag while creating the volume, it would be very useful for debug, analytics (like which tool is actually creating more volumes etc).

To start with we can mark everything going in with cli as `tag:'cli'`.

@JohnStrunk (add more if something is missed)

atinmu:
So this can be an option in the volume create request API and then 'gluster volume info --tag=something' can filter this out. Is this what the requirement is? If yes, this should be quite easy to implement. Need a change in volume create & volume info (ReST APIs and CLI commands)

atinmu:
## Checklist
- [ ] ReST APIs availibility in wiki
- [ ] CLI commands
- [ ] Sufficient logging
- [ ] Availibility of metrics data if applicable
- [ ] Xlator options table upto date?
- [ ] Test cases & test coverage of the functionality

JohnStrunk:
To clarify my original intent with this feature: I'm looking for the ability to store additional metadata with gluster entities that are exposed externally (e.g., volumes, nodes, clusters, snapshots). If devices/bricks are exposed, they should be taggable also.
My use case is for external tools to tag entities that they are managing, both for loosely registering ownership as well as storing configuration information. For example, a kube provisioner might tag a volume with information such as:
- owner = kube-gluster-subvol <== this volume is being used for subdir-based PVs
- overcommit = 2.0 <== the provisioner will hand out PVs with quota totaling 2x the capacity of the volume
- max-subvols = 5000 <== the provisioner will put a max of 5000 subdir-based PVs on this volume.

The above are just examples of how I could see this being used. glusterd should not interpret these tags in any way other than to allow filtering as @amarts showed above.

prashanthpai:
It's in the plan to allow custom metadata (key/value pairs) for peer and volume objects. Core glusterd shall not interpret or use these in any way. However, at most glusterd2 will impose a size (in bytes) restriction.

prashanthpai:
@vpandey-RH Checklist:

- [ ] Add volume edit REST API
- [ ] Add CLI support
- [ ] Add support to delete metadata keys

prashanthpai:
[...]

PR #1228, opened 2018-09-06
Implement Labels 

rafikc30:
Labels are used to tag different object, and perform
collective operation based on label configuration values

More information on #issue:1094


prashanthpai:
As discussed offline, I expected some discussion on the issue first. There's clearly a overlap between metadata and labels. See these issues for some previous discussions:

https://github.com/gluster/glusterd2/issues/462
https://github.com/gluster/glusterd2/issues/1018
https://github.com/gluster/glusterd2/issues/1018

prashanthpai:
@rafikc30 What is generally referred as tags or labels in other projects, we call them metadata (key-value pairs). They are not meant to be interpreted by core. There is no inherent meaning (or actions) attached to these arbitrary strings in the core glusterd2.

If you wish to implement snapshot scheduler as an external entity/daemon or a glusterd2 plugin, you can use the metadata infrastructure. 

> But it cannot take intelligent decisions like when should we delete

The core is not supposed to - regardless of whether you call them labels, tags or metadata.

On the issue you mention:

> The first consumer of labels will be Snapshot. We can use labels to list snapshot, delete the snapshot and perform other high-level actions.

You can use metadata for all of these usecases. These higher level actions are performed by user or some external entity. For example:

* list all snapshots that have has a label/tag `backup`
* delete all snapshots that have a label/tag `thumbnails`
* delete all snapshots that have `timestamp` label/tag set to a value older than certain date

> For example, In a customer setup, a snapshot can schedule based on different labels like Monthly, Hourly, Yearly, Daily etc. Each one of this can have different configuration values. In a typical set up, the maximum number of a snapshot for the Hourly label can set to 24, while the same value will be 12 for the Monthly snapshot. So performing auto-delete based on this max values will be based on label values.

The core of glusterd2 will not be doing these automatic operations. You can do it as an external daemon or a plugin.

As snapshot is considered core in glusterd2, if you want these automatic operations to be done as core, you might as well make these operations concrete with a well-defined API to schedule+prune snapshots and not use metadata which is arbitrary.

What other specific and real-world usecase does your labels proposal solve beyond snapshot for it to be a generic infrastructure ?

As pointed out earlier, this discussion is best done over the issue.

atinmu:
@aravindavk @kshlm Would like to get this picked up for GCS/0.4 milestone

atinmu:
@kshlm can you please add this to your review backlog ?

kshlm:
@rafikc30 was there any further discussion about metadata vs labels between you and @prashanthpai? The conversation seems to have stopped in between.

PR #1230, opened 2018-09-10
Better support for environment variables in container image

kshlm:
This PR enables a few settings in the glusterd2.service systemd unit to enable
better support for environment variables in containers.

Also, prepares for v5.0.0-rc0 tagging.

### DIFF
--- a/extras/nightly-container/provision.yml
+++ b/extras/nightly-container/provision.yml
@@ -77,6 +77,33 @@
       args:
         warn: no
 
+    - name: Create override directory for glusterd2.service
+      file:
+        path: /etc/systemd/system/glusterd2.service.d
+        state: directory
+
+    # This makes systemd pass environment variables set for GD2 by kubernetes
+    - name: Enable PassEnvironment override for glusterd2.service
+      ini_file:
+        path: /etc/systemd/system/glusterd2.service.d/override.conf
+        section: Service
+        option: PassEnvironment
+        # PassEnvironment requires that each variable be mentioned individually
+        # More variables will need to be added here as required
+        value: GD2_ETCDENDPOINTS GD2_CLUSTER_ID GD2_RESTAUTH GD2_CLIENTADDRESS GD2_PEERADDRESS
+
+    - name: Create /etc/sysconfig/glusterd2/
+      file:
+        path: /etc/sysconfig/glusterd2
+        state: directory
+
+    - name: Disable embedded etcd for GD2
+      lineinfile:
+        path: /etc/sysconfig/glusterd2/noembed
+        create: yes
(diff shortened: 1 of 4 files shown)


extras/nightly-container/provision.yml
```diff
@@ -77,11 +77,37 @@
...
+        # PassEnvironment requires that each variable be mentioned individually
+        # More variables will need to be added here as required
+        value: GD2_ETCDENDPOINTS GD2_CLUSTER_ID GD2_RESTAUTH
+
+    - name: Create /etc/sysconfig/glusterd2/
+      file:
+        path: /etc/sysconfig/glusterd2
+        state: directory
+
+    - name: Disable embedded etcd for GD2
```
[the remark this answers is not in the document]
kshlm: I still want to have embedded etcd be the default for normal GD2 usecases. So I prefer it this way.

extras/nightly-container/provision.yml
```diff
@@ -77,11 +77,37 @@
...
+      lineinfile:
+        path: /etc/sysconfig/glusterd2/noembed
```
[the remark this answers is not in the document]
kshlm: This is being configured as an environment variable here too. But since this is gonna be the default for all launched containers, we're setting this with the image itself. The other options that we pass as pod environment variables change for each pod.

version/version.go
```diff
@@ -14,13 +14,13 @@ var (
...
 const (
-	MaxOpVersion = 40100
+	MaxOpVersion = 50000
 	APIVersion   = 1
 )
 
 // GlusterdVersion and GitSHA
 var (
-	GlusterdVersion = "4.1.0"
+	GlusterdVersion = "5.0.0-rc0"
```
prashanthpai: When I had introduced this variable, the intention was to not initialise it with any hard-coded value. We populate it during build time:

**scripts/build.sh**
```sh
```

kshlm: I understand that this is always reset by the build script. But I didn't want to leave it at 4.1.0.

prashanthpai: Can be set to an empty string.

extras/nightly-container/provision.yml
```diff
@@ -77,11 +77,37 @@
       args:
         warn: no
 
+    - name: Create override directory for glusterd2.service
+      file:
+        path: /etc/systemd/system/glusterd2.service.d
+        state: directory
+
+    # This makes systemd pass environment variables set for GD2 by kubernetes
```
Madhu-1: @kshlm, I think this is not only meant for kubernetes.  can we update the comments saying any CO can set this environment variables,

PR #1241, opened 2018-09-14
snapshot/delete: Implement snapshot delete all

rafikc30:
Glustercli will have an added feature of deleting all
snapshots in the cluster or all snapshots of a particular
volume


Madhu-1:
@rafikc30 can you add example CLI commands for this PR.

glustercli/cmd/snapshot-delete.go
```diff
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 
+	"github.com/gluster/glusterd2/pkg/api"
```
Madhu-1: add a new line here

glustercli/cmd/snapshot-delete.go
```diff
@@ -19,24 +20,78 @@ var (
...
+}
+
+func snapshotDeleteAllCmdRun(cmd *cobra.Command, args []string) {
+	var snaps api.SnapListResp
+	var err error
+	volname := ""
+	if len(args) > 0 {
+		volname = args[0]
+	}
+
```
rishubhjain: @rafikc30 I think there is a need to throw an error if the volume is not provided.

rafikc30: Volume is an optional argument. If omitted, then perform the deletion of all snapshots in the cluster. I have updated the example cli commands in commit message

glustercli/cmd/snapshot-delete.go
```diff
@@ -19,24 +21,78 @@ var (
...
+		return err
 	}
 	fmt.Printf("%s Snapshot deleted successfully\n", snapname)
+	return nil
+}
+
+func snapshotDeleteCmdRun(cmd *cobra.Command, args []string) {
+
+	if len(args) != 1 {
+		//snapshotDeleteAllCmdRun will take care of else case.
```
Madhu-1: @rafikc30 just a question, `snapshotDeleteAllCmdRun` will take care if `len(args) != 1` right?
if yes, should this comment need to be updated?

rafikc30: yes, you are right

glustercli/cmd/snapshot-delete.go
```diff
@@ -19,24 +21,78 @@ var (
...
-		failure("Snapshot delete failed", err, 1)
+		return err
 	}
 	fmt.Printf("%s Snapshot deleted successfully\n", snapname)
+	return nil
+}
+
+func snapshotDeleteCmdRun(cmd *cobra.Command, args []string) {
+
+	if len(args) != 1 {
```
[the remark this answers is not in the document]
rafikc30: I thought the command `all` and 'voname' will also count in the handler snapshotDeleteCmdRun. I was wrong, I will remove the code and will make cobra.ExactArgs(1)

Issue #1156, opened 2018-08-10
unmount volumes irrespective of test pass or fail


Madhu-1:
even if the test cases fails we need to unmount the volumes.

```
[root@glusterd2]# make functest
dep package is available
gometalinter package is available

=== RUN   TestBitrot
=== RUN   TestBitrot/Replica-volume
=== RUN   TestBitrot/Dist-volume
--- PASS: TestBitrot (8.03s)
    --- PASS: TestBitrot/Replica-volume (0.84s)
    --- PASS: TestBitrot/Dist-volume (0.60s)
=== RUN   TestGeorepCreateDelete
--- PASS: TestGeorepCreateDelete (6.73s)
=== RUN   TestSelfHeal
--- PASS: TestSelfHeal (5.12s)
=== RUN   TestAddRemovePeer
--- FAIL: TestAddRemovePeer (21.11s)
	require.go:765: 
			Error Trace:	peer_ops_test.go:42
			Error:      	Expected nil, but got: &errors.errorString{s:"failed to connect to store"}
			Test:       	TestAddRemovePeer
=== RUN   TestRESTAPIAuth
--- PASS: TestRESTAPIAuth (4.05s)
=== RUN   TestRestart
--- PASS: TestRestart (6.68s)
=== RUN   TestSmartVolume
=== RUN   TestSmartVolume/Smartvol_Distributed_Volume
=== RUN   TestSmartVolume/Smartvol_Replicate_2_Volume
=== RUN   TestSmartVolume/Smartvol_Replicate_3_Volume
=== RUN   TestSmartVolume/Smartvol_Arbiter_Volume
=== RUN   TestSmartVolume/Smartvol_Disperse_Volume
=== RUN   TestSmartVolume/Smartvol_Distributed-Replicate_Volume
=== RUN   TestSmartVolume/Smartvol_Distributed-Disperse_Volume
=== RUN   TestSmartVolume/Smartvol_Without_Name
--- PASS: TestSmartVolume (34.97s)
	smartvol_ops_test.go:243: Using temp dir: /tmp/gd2_func_test/TestSmartVolume227634819
    --- PASS: TestSmartVolume/Smartvol_Distributed_Volume (2.15s)
    --- PASS: TestSmartVolume/Smartvol_Replicate_2_Volume (0.96s)
    --- PASS: TestSmartVolume/Smartvol_Replicate_3_Volume (1.69s)
[...]

Issue #1235, opened 2018-09-14
cleanup resources if tests in e2e fails


Madhu-1:
### Observed behavior
if test cases in e2e fail, we are not cleaning up the resource, which leaves the mounted volume as it is

### Expected/desired behavior
even if the test cases fails we need to clean up the used resources

### Logs
```
=== RUN   TestSmartVolume/Smartvol_Without_Name
--- FAIL: TestSmartVolume (50.11s)
	smartvol_ops_test.go:263: Using temp dir: /tmp/gd2_func_test/TestSmartVolume022562123
    --- FAIL: TestSmartVolume/Smartvol_Distributed_Volume (3.13s)
    	require.go:765: 
    			Error Trace:	smartvol_ops_test.go:63
    			Error:      	Expected nil, but got: &errors.errorString{s:"Transaction step vol-delete.CleanBricks failed on peer 083e19d9-ea0c-4c9e-b339-bb57e76d9839 with error: exit status 5;   Failed to enumerate udev device list.\n  WARNING: Device for PV mV2M0a-iqMo-6THZ-ahXh-O3E0-Ul8H-uHb3af not found or rejected by a filter.\n  Aborting vg_write: No metadata areas to write to!\n\n"}
    			Test:       	TestSmartVolume/Smartvol_Distributed_Volume
    --- FAIL: TestSmartVolume/Smartvol_Replicate_2_Volume (2.28s)
    	require.go:157: 
    			Error Trace:	smartvol_ops_test.go:36
    			            				smartvol_ops_test.go:87
    			Error:      	Not equal: 
    			            	expected: 0
    			            	actual  : 1
    			Test:       	TestSmartVolume/Smartvol_Replicate_2_Volume
    --- FAIL: TestSmartVolume/Smartvol_Replicate_3_Volume (2.76s)
    	require.go:157: 
    			Error Trace:	smartvol_ops_test.go:36
    			            				smartvol_ops_test.go:111
    			Error:      	Not equal: 
    			            	expected: 0
    			            	actual  : 1
    			Test:       	TestSmartVolume/Smartvol_Replicate_3_Volume
    --- FAIL: TestSmartVolume/Smartvol_Arbiter_Volume (2.66s)
    	require.go:157: 
    			Error Trace:	smartvol_ops_test.go:36
    			            				smartvol_ops_test.go:140
    			Error:      	Not equal: 
    			            	expected: 0
    			            	actual  : 1
[...]

PR #1243, opened 2018-09-17
fix issue in e2e 

Madhu-1:
remove code duplication in volume_ops_test
call checkFuseAvailable from utils package to avoid code duplication.

after volume mount even if some of the operations fail on the volume. we should umount the volume
calling umount in defer immediately after the volume mount will ensure to call the umount if some operation
fails.
Partial: #1235 

### DIFF
--- a/e2e/snapshot_ops_test.go
+++ b/e2e/snapshot_ops_test.go
@@ -314,6 +314,8 @@ func testRestoredVolumeMount(t *testing.T, tc *testCluster) {
 	err := mountVolume(host, snapTestName, mntPath)
 	r.Nil(err, fmt.Sprintf("mount failed: %s", err))
 
+	defer syscall.Unmount(mntPath, syscall.MNT_FORCE)
+
 	err = testMount(mntPath)
 	r.Nil(err)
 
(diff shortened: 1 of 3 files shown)


e2e/volume_ops_test.go
```diff
@@ -383,9 +383,7 @@ func testVolumeInfo(t *testing.T) {
 }
 
 func testVolumeStatus(t *testing.T) {
-	if _, err := os.Lstat("/dev/fuse"); os.IsNotExist(err) {
-		t.Skip("skipping mount /dev/fuse unavailable")
-	}
+	checkFuseAvailable(t)
```
harigowtham: I cant find this function. Am i missing anything?

rishubhjain: https://github.com/gluster/glusterd2/blob/eb548f7cf915b7528cd0605e0f744b3d796782d3/e2e/utils_test.go#L361

e2e/snapshot_ops_test.go
```diff
@@ -314,6 +314,8 @@ func testRestoredVolumeMount(t *testing.T, tc *testCluster) {
 	err := mountVolume(host, snapTestName, mntPath)
 	r.Nil(err, fmt.Sprintf("mount failed: %s", err))
 
+	defer syscall.Unmount(mntPath, syscall.MNT_FORCE|syscall.MNT_DETACH)
```
prashanthpai: Please don't use `syscall.MNT_DETACH`. Here and elsewhere in this PR. These are all glusterfs mounts and not XFS mounts.

Remove it from `e2e/glustershd_test.go` too

Madhu-1:
Fixes: #1156

Madhu-1:
@prashanthpai addressed review comments please review

Issue #1244, opened 2018-09-17
The volume name given to the smart volume to be taken as volume name for further operations


Shrivaibavi:
### Observed Behaviour
When created a smart volume with a name and tried to start the volume with the name it says volume not found

``` 
[root@dhcp35-30 ~]# glustercli volume create testing_volume --size=200
vol_e7c29062-6220-4222-b271-de75bf92bace Volume created successfully
Volume ID:  24c87364-5fa4-401a-a466-64df160d2ee0
[root@dhcp35-30 ~]# 
[root@dhcp35-30 ~]# 
[root@dhcp35-30 ~]# glustercli volume start testing_volume
volume start failed

Response headers:
X-Gluster-Cluster-Id: 3a526594-5b01-410f-a445-0a6adc2a901e
X-Gluster-Peer-Id: 7ebfbea6-59e6-494a-b157-cdf35603e624
X-Request-Id: d35c3da1-644c-4fd9-893a-7c96e625da7d

Response body:
volume not found

[root@dhcp35-30 ~]# glustercli volume start vol_e7c29062-6220-4222-b271-de75bf92bace
Volume vol_e7c29062-6220-4222-b271-de75bf92bace started successfully
```
### Expected Behaviour
The volume name given to the smart volume to be taken as volume name for further operations

rishubhjain:
@aravindavk I think this issue can be closed now.

aravindavk:
@Shrivaibavi please close this issue if the latest build fixes this issue.

PR #1245, opened 2018-09-17
Make Volume name mandatory in Volume create request

aravindavk:
- CSI driver also passes Volume name to Volume create API
- Mandatory Volume name also helps to ignore
  duplicate(accidental/network failure) Volume create
  API requests(Discussed here
  https://github.com/gluster/glusterd2/issues/919#issuecomment-401022413)
- Applications can generate name and can pass to API, not much benefit
  of auto generating name from glusterd2
- Supporting both the syntax in CLI adds confusion

### DIFF
--- a/e2e/smartvol_ops_test.go
+++ b/e2e/smartvol_ops_test.go
@@ -229,19 +229,6 @@ func testSmartVolumeDistributeDisperse(t *testing.T) {
 	checkZeroLvs(r)
 }
 
-func testSmartVolumeWithoutName(t *testing.T) {
-	r := require.New(t)
-
-	createReq := api.VolCreateReq{
-		Size: 20,
-	}
-	volinfo, err := client.VolumeCreate(createReq)
-	r.Nil(err)
-
-	r.Nil(client.VolumeDelete(volinfo.Name))
-	checkZeroLvs(r)
-}
-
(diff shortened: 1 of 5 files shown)


atinmu:
A basic question - as in the legacy case (Heketi + GD1) , volume is created with a convention vol_<UUID> (UUID is generated by heketi and passed to cli/glusterd) , with GCS, who would take up the responsibility of forming up the volume name, is it CSI driver or someone else?

aravindavk:
> A basic question - as in the legacy case (Heketi + GD1) , volume is created with a convention vol_ (UUID is generated by heketi and passed to cli/glusterd) , with GCS, who would take up the responsibility of forming up the volume name, is it CSI driver or someone else?

Name is required field in Volume claim of CSI driver(https://github.com/gluster/gluster-csi-driver/blob/master/pkg/glusterfs/controllerserver.go#L41). CSI driver collects the name from claim and passes it to glusterd2.(Ref: https://github.com/gluster/gluster-csi-driver/blob/master/pkg/glusterfs/controllerserver.go#L94)

rishubhjain:
Fixes https://github.com/gluster/glusterd2/issues/1244

aravindavk:
Since Volume name is required field in CSI driver, whoever calls the CSI driver provides the volume name

PR #1247, opened 2018-09-18
Refactor and improve pmap package

prashanthpai:
The earlier pmap implementation in Go was non-idiotomatic and mostly a
verbatim copy of glusterd1's pmap implementation. This refactor of pmap
package improves it in the following ways:

* Redesgined pmap data structure and methods with brick multiplexing
  and API consumption in mind.
* Store net.Conn object instead of arbitrary and non-typesafe Xprt
  interface{}
* Prune pmap registry on brick process TCP disconnection which should
  clean stale ports on ungraceful kill of brick processes.
* Always recheck if foreign ports are free when allocating a new port.
* Store whether individual bricks have signed in and export this info
  during statedump.
* Explicitly initialize (involves binding to thousands of ports) pmap
  registry during glusterd2 startup and not via implicit init(). This
  will ensure we don't do this when running unit tests.
* Add some unit tests.


glusterd2/pmap/registry.go
```diff
@@ -0,0 +1,207 @@
...
+// Allocate finds a free port and returns the port number. The
+// allocated free port will be in leased state until the brick
+// process it was leased to sends a SIGN IN request to glusterd2.
+func (r *registryType) Allocate(brickpath string) int {
+	r.Lock()
+	defer r.Unlock()
+
+	var port int
+	for p := r.basePort; p <= gfPortMax; p++ {
+		if r.ports[p].State == GfPmapPortFree || r.ports[p].State == GfPmapPortForeign {
```
kshlm: Shouldn't you be skipping PortForeign here?

glusterd2/pmap/registry.go
```diff
@@ -0,0 +1,207 @@
...
+				} else {
+					r.ports[p].State = GfPmapPortForeign
+					r.PortUnlock(p)
+				}
+			}
+		}
+	}
+
+	if port == 0 {
+		panic("registryAlloc() we ran out of free ports!")
```
kshlm: Don't panic. Return an error so that the brick doesn't start and a proper error is returned to the user.

glusterd2/pmap/registry.go
```diff
@@ -0,0 +1,207 @@
...
+	GfPmapPortFree PortState = iota
+	GfPmapPortForeign
+	GfPmapPortLeased
+	GfPmapPortInUse
+)
+
+type portStatus struct {
+	Port  int       `json:"port"`
+	State PortState `json:"state"`
+	Conn  net.Conn  `json:"-"`
```
kshlm: A suggestion here. Let me know what you think. We could have `portStatus` also track the bricks status. And in place of the brickpath->brickStatus map, we have a brickpath->port and conn->port maps. This would allow quick lookups later on, avoiding iteration, especially when dealing with a large number of bricks.

So we'd have,

```go
type portStatus struct {
// Don't need the Port number in the status. This will be in the correctly indexed position in the ports slice.
  State PortState
  Conn net.Conn
  Bricks map[string]bool
}

type pmapRegistry struct {
  .
  .
  bricks map[string]int
  conns map[net.Conn]int
 .
 .
}
```

Having a conn->port map will make it much simpler to delete whole connections now.
```go
func (r *PmapRegistry) RemoveByConn (conn net.Conn) error {
  p := r.conns[conn]

  delete(r.conns, conn)

  for brick, _ := range p.Bricks {
    delete(r.bricks, brick)
  }
  p.Status = PortFree
  p.Conn = nil
  p.Bricks = make(map[string]bool)
}
```

glusterd2/pmap/registry.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+type brickStatus struct {
+	Port     int  `json:"port"`
+	SignedIn bool `json:"signed-in"`
+}
+
+type registryType struct {
```
kshlm: Why not call this `PmapRegistry`?

glusterd2/pmap/registry.go
```diff
@@ -0,0 +1,207 @@
...
+
+	if port < 0 || port > gfPortMax {
+		return fmt.Errorf("registry.Add(): invalid port %d", port)
+	}
+
+	r.Lock()
+	defer r.Unlock()
+
+	r.ports[port].State = GfPmapPortInUse
+	r.ports[port].Conn = conn
```
kshlm: Since it's possible that multiple bricks are multiplexed on a single conn, ensure that the conn passed to the function, is the same as before.

glusterd2/pmap/registry.go
```diff
@@ -0,0 +1,207 @@
...
+	return -1, fmt.Errorf("SearchByBrickPath: port for brick %s not found", brickpath)
+}
+
+// RemoveByConn deletes port map entry by brick process's TCP connection.
+// There will be only one TCP connection per brick process, regardless of
+// number of bricks in the process.
+func (r *registryType) RemovePortByConn(conn net.Conn) {
+
+	if conn == nil {
+		panic("RemovePortByConn(): conn passed is nil")
```
kshlm: Again, better to return an error here, than panic.

glusterd2/pmap/registry.go
prashanthpai: I did think about `conn->port` mapping earlier and removing `Port` field. Had decided to do that later. Now that you've pointed it out, I'll update it in the same PR :+1:

glusterd2/pmap/registry.go
prashanthpai: `PortForeign` was marked only during glusterd2's startup. That would've changed by a later time. In e2e test scenario with multiple gd2 instances, a `GfPmapPortForeign` could also mean that it's being used by another glusterd2 instance.

glusterd2/pmap/registry.go
prashanthpai: I just remembered why I marked this as `panic`. The call flow upwards reaches daemon interface's `Args() []string` method which doesn't return error. An alternative is to assign port right when `brick.Glusterfsd` object gets created, which isn't clean.

To avoid making invasive changes outside of pmap package, I'll leave it as is for now. When glusterfs change https://review.gluster.org/14670 gets in, this port allocation code will be disposed off anyway.

glusterd2/pmap/registry.go
prashanthpai: Can do :+1: However, the call stack upwards has no provision to return the error anywhere (to user or brick). Will have the caller log it though.

prashanthpai:
Sample output:

```sh
[ppai@gd2-1 glusterd2]$ glustercli volume create test --replica 2 127.0.0.1:/export/brick{1..4}/data --force
test Volume created successfully
Volume ID:  3bb93af9-6661-44d2-8b00-78e9a96cf6ae

[ppai@gd2-1 glusterd2]$ glustercli volume start test
Volume test started successfully

[ppai@gd2-1 glusterd2]$ curl -s -X GET http://127.0.0.1:24007/statedump | grep pmap
"pmap": [{"port":49152,"state":3,"bricks":{"/export/brick1/data":true}},{"port":49153,"state":3,"bricks":{"/export/brick2/data":true}},{"port":49154,"state":3,"bricks":{"/export/brick3/data":true}},{"port":49155,"state":3,"bricks":{"/export/brick4/data":true}}],
```

```sh
[ppai@gd2-1 glusterd2]$ glustercli volume stop test
Volume test stopped successfully

[ppai@gd2-1 glusterd2]$ curl -s -X GET http://127.0.0.1:24007/statedump | grep pmap
"pmap": null,
```

```sh
[ppai@gd2-1 glusterd2]$ glustercli volume start test
Volume test started successfully

[ppai@gd2-1 glusterd2]$ sudo pkill -9 glusterfsd

[ppai@gd2-1 glusterd2]$ curl -s -X GET http://127.0.0.1:24007/statedump | grep pmap
"pmap": null,
```

```sh
[ppai@gd2-1 glusterd2]$ glustercli volume start test --force
Volume test started successfully

[ppai@gd2-1 glusterd2]$ curl -s -X GET http://127.0.0.1:24007/statedump | grep pmap
"pmap": [{"port":49152,"state":3,"bricks":{"/export/brick1/data":true}},{"port":49153,"state":3,"bricks":{"/export/brick2/data":true}},{"port":49154,"state":3,"bricks":{"/export/brick3/data":true}},{"port":49155,"state":3,"bricks":{"/export/brick4/data":true}}],
```

glusterd2/pmap/registry.go
```diff
@@ -0,0 +1,207 @@
...
+	gfIanaPrivPortsStart = 49152
+	gfPortMax            = 65535
+)
+
+// PortState represents the type of state of the port
+type PortState int32
+
+// List of states an individual port can be in
+const (
+	GfPmapPortFree PortState = iota
```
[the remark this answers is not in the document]
prashanthpai: Done

glusterd2/pmap/registry.go
prashanthpai: Done but the error is not propagated.

glusterd2/pmap/registry.go
prashanthpai: Done.

Removed `Conn` field from portStatus as it wasn't being used anywhere.

glusterd2/pmap/registry.go
prashanthpai: Done

glusterd2/pmap/registry.go
prashanthpai: That didn't work. We have to store the last/latest connection. There's no guarantee that sign out shall come on the same TCP connection as sign in.

glusterd2/pmap/registry.go
prashanthpai: Done

glusterd2/pmap/registry.go
```diff
@@ -0,0 +1,226 @@
...
+	if port < 0 || port > PortMax {
+		return fmt.Errorf("registry.Remove(): invalid port %d", port)
+	}
+
+	r.Lock()
+	defer r.Unlock()
+
+	delete(r.bricks, brickpath)
+	if _, ok := r.ports[port].Bricks[brickpath]; ok {
+		delete(r.ports[port].Bricks, brickpath)
```
kshlm: If it was the last brick being removed, shouldn't you be resetting the port?

prashanthpai: The last brick being removed will eventually disconnect, at which point, `RemovePortByConn` (see line 171 and 172) will reset the port.

PR #1256, opened 2018-09-29
Plugin architecture for brick Provisioners

aravindavk:
DO NOT MERGE!

WIP: Testing and adding interface for snapshot is in progress

Code is refactored to create Provisioner interface and lvm
provisioner is now a plugin

To create a new Provisioner, implement the following interface
and add it to `glusterd2/provisioners/provisioner.go`

```
type Provisioner interface {
	// Register will be called when user registers a device or directory.
	// PV and VG will be created in case of lvm plugin
	Register(devpath string) error
	// AvailableSize returns available size of the device
	AvailableSize(devpath string) (uint64, uint64, error)
	// Unregister will be called when device/directory needs to be removed
	Unregister(devpath string) error
	// CreateBrick creates the brick volume
	CreateBrick(devpath, brickid string, size uint64, bufferFactor float64) error
	// CreateBrickFs will create the brick filesystem
	CreateBrickFS(devpath, brickid, fstype string) error
	// CreateBrickDir will create the brick directory
	CreateBrickDir(brickPath string) error
	// MountBrick will mount the brick
	MountBrick(devpath, brickid, brickPath string) error
	// UnmountBrick will unmount the brick
	UnmountBrick(devpath, brickid string) error
	// RemoveBrick will remove the brick
	RemoveBrick(devpath, brickid string) error
}
```


plugins/lvmprovisioner/provisioner.go
```diff
@@ -0,0 +1,133 @@
...
+	"errors"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/gluster/glusterd2/pkg/fsutils"
+	"github.com/gluster/glusterd2/pkg/lvmutils"
+)
+
+var mountOpts = "rw,inode64,noatime,nouuid"
```
amarts: We also need 'discard' as a mount option? So that actual backend gets free'd up ?

pkg/api/volume_req.go
```diff
@@ -2,20 +2,17 @@ package api
...
-	Mountdir       string `json:"mount-dir,omitempty"`
-	DevicePath     string `json:"device-path,omitempty"`
-	MntOpts        string `json:"mnt-opts,omitempty"`
-	FsType         string `json:"fs-type,omitempty"`
+	Type       string `json:"type"`
+	PeerID     string `json:"peerid"`
+	Path       string `json:"path"`
+	Name       string `json:"name,omitempty"`
+	Size       uint64 `json:"size,omitempty"`
+	VgID       string `json:"vg-id,omitempty"`
```
amarts: do we need this any more?

aravindavk:
- Snapshot related code is not refactored in this PR, I will add Snapshot related changes in separate PR.
- This PR also includes device API refactor

Please review

plugins/device/deviceutils/errors.go
```diff
@@ -0,0 +1,10 @@
+package deviceutils
```
Madhu-1: instead of having a separate file, can we make use of `pkg/error/error.go`?

plugins/device/deviceutils/store-utils.go
```diff
@@ -1,159 +1,114 @@
...
 		return err
 	}
 
-	peerInfo.Metadata["_devices"] = string(deviceJSON)
-	err = peer.AddOrUpdatePeer(peerInfo)
+	return nil
+}
+
+// UpdateDeviceFreeSize updates the actual available size of VG
+func UpdateDeviceFreeSize(peerid, device string, size uint64, extentSize uint64) error {
```
Madhu-1: `size uint64, extentSize uint64` can be made as `size , extentSize uint64`

plugins/device/transaction.go
```diff
@@ -21,39 +22,47 @@ func txnPrepareDevice(c transaction.TxnCtx) error {
...
 	}
-	vgName := strings.Replace("vg"+device, "/", "-", -1)
-	err = deviceutils.CreateVG(device, vgName)
-	if err != nil {
-		c.Logger().WithError(err).WithField("device", device).Error("Failed to create volume group")
-		errPV := deviceutils.RemovePV(device)
-		if errPV != nil {
-			c.Logger().WithError(err).WithField("device", device).Error("Failed to remove physical volume")
+
+	var provisioner provisioners.Provisioner
```
Madhu-1: this also can be avoided if we have this code in separate function

plugins/device/deviceutils/errors.go
aravindavk: not added in `pkg/error` since this is in `plugins` right now

glusterd2/commands/volumes/volume-smartvol-txn.go
```diff
@@ -21,74 +20,76 @@ func txnPrepareBricks(c transaction.TxnCtx) error {
...
+			err = provisioner.CreateBrickDir(b.Path)
 			if err != nil {
 				c.Logger().WithError(err).WithField(
 					"path", b.Path).Error("failed to create brick directory in mount")
 				return err
 			}
 
-			// Update current Vg free size
-			err = deviceutils.UpdateDeviceFreeSize(gdctx.MyUUID.String(), b.VgName)
+			availableSize, extentSize, err := provisioner.AvailableSize(b.Device)
```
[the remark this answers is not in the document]
aravindavk: ok

glusterd2/brick/types.go
```diff
@@ -33,6 +33,7 @@ type MountInfo struct {
 // Brickinfo is the static information about the brick
 type Brickinfo struct {
 	ID             uuid.UUID
+	Name           string
```
prashanthpai: What is stored in this variable ?

prashanthpai:
@rafikc30 Can you please review the LVM related changes ?
@rishubhjain Can you please review device related changes ?

rafikc30:
@prashanthpai I will review by eod

pkg/lvmutils/utils.go
```diff
@@ -16,6 +15,16 @@ const (
 	chunkSize         = "1280k"
 )
 
+// MbToKb converts Value from Mb to Kb
```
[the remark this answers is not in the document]
aravindavk: I left it since this will go away once we convert everything to bytes.

pkg/fsutils/utils.go
```diff
@@ -0,0 +1,31 @@
+package fsutils
+
+import (
+	"syscall"
+
+	"github.com/gluster/glusterd2/pkg/utils"
+)
+
+// MakeXfs creates XFS filesystem
```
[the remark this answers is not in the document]
aravindavk: ack

pkg/fsutils/utils.go
```diff
@@ -0,0 +1,31 @@
...
+	return utils.ExecuteCommandRun("mkfs.xfs",
+		"-i", "size=512",
+		"-n", "size=8192",
+		dev,
+	)
+}
+
+// Mount mounts the brick LV
+func Mount(dev, mountdir, options string) error {
+	return utils.ExecuteCommandRun("mount",
```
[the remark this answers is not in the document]
aravindavk: ack

plugins/device/rest.go
```diff
@@ -41,6 +41,18 @@ func deviceAddHandler(w http.ResponseWriter, r *http.Request) {
...
 	defer txn.Done()
 
+	_, err = deviceutils.GetDevice(peerID, req.Device)
+	if err == nil {
+		logger.WithError(err).WithField("device", req.Device).Error("Device already exists")
+		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, "device already exists")
+		return
+	}
+
+	if err != deviceutils.ErrDeviceNotFound {
```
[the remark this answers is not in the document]
aravindavk: ok

plugins/device/rest.go
```diff
@@ -87,35 +86,48 @@ func deviceAddHandler(w http.ResponseWriter, r *http.Request) {
...
-		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, "failed to get peer from store")
+		logger.WithError(err).WithField("peerid", peerID).Error("Failed to get device from store")
+		restutils.SendHTTPError(ctx, w, http.StatusInternalServerError, "failed to get device from store")
 		return
 	}
 
 	// FIXME: Change this to http.StatusCreated when we are able to set
 	// location header with a unique URL that points to created device.
-	restutils.SendHTTPResponse(ctx, w, http.StatusOK, peerInfo)
+	restutils.SendHTTPResponse(ctx, w, http.StatusOK, deviceInfo)
```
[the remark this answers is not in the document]
aravindavk: sure

plugins/device/rest.go
```diff
@@ -87,35 +86,48 @@ func deviceAddHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	err = txn.Ctx.Set("provisioner", &req.Provisioner)
```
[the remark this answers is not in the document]
aravindavk: ok, I will merge multiple Ctx.Set into one

plugins/device/transaction.go
```diff
@@ -21,39 +22,47 @@ func txnPrepareDevice(c transaction.TxnCtx) error {
...
+	var err error
+	if provisionerType == "" {
+		provisioner = provisioners.GetDefault()
+	} else {
+		provisioner, err = provisioners.Get(provisionerType)
+		if err != nil {
```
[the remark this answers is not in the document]
aravindavk: ok

pkg/api/volume_req.go
```diff
@@ -2,20 +2,16 @@ package api
...
-	DevicePath     string `json:"device-path,omitempty"`
-	MntOpts        string `json:"mnt-opts,omitempty"`
-	FsType         string `json:"fs-type,omitempty"`
+	Type       string `json:"type"`
+	PeerID     string `json:"peerid"`
+	Path       string `json:"path"`
+	Name       string `json:"name,omitempty"`
+	Size       uint64 `json:"size,omitempty"`
+	Mountdir   string `json:"mount-dir,omitempty"`
+	Device     string `json:"device,omitempty"`
```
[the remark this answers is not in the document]
aravindavk: I need to think about different naming for this. Device is main device where VG is created in case of lvm. DevicePath is path of LV where brick needs to be mounted. Is `RootDevice` or `ParentDevice` sounds good instead of just `Device`?

glusterd2/utils/mount.go
```diff
@@ -46,6 +47,28 @@ func MountLocalBricks() error {
 					continue
 				}
 
+				if v.Provisioner != "" {
+					provisioner, err := provisioners.Get(v.Provisioner)
```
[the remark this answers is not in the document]
aravindavk: Currently Snapshot Provisioned volumes are not included in this Provisioner interface. Mounting Snapshot provisioned volume or restored volume support needs to be handled.

glusterd2/brick/types.go
aravindavk: `<volname>_s<subvol-number>_b<brick-number>` similar to subvolume name

glusterd2/provisioners/provisioners.go
```diff
@@ -0,0 +1,54 @@
...
+type Provisioner interface {
+	// Register will be called when user registers a device or directory.
+	// PV and VG will be created in case of lvm plugin
+	Register(devpath string) error
+	// AvailableSize returns available size of the device
+	AvailableSize(devpath string) (uint64, uint64, error)
+	// Unregister will be called when device/directory needs to be removed
+	Unregister(devpath string) error
+	// CreateBrick creates the brick volume
+	CreateBrick(devpath, brickid string, size uint64, bufferFactor float64) error
```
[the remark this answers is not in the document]
aravindavk: makes sense, I will merge the functions into one. `MountBrick` is also required as separate since bricks needs to be mounted on glusterd start/node reboot.

```
Register
AvailableSize
Unregister
CreateBrick
MountBrick
UnmountBrick
RemoveBrick
```

`CreateBrick` also mounts the brick. but exported `MountBrick` can be used separately when required.

glusterd2/utils/mount.go
```diff
@@ -46,6 +47,28 @@ func MountLocalBricks() error {
...
+					if err != nil {
+						log.WithError(err).WithFields(log.Fields{
+							"volume":      v.Name,
+							"provisioner": v.Provisioner,
+						}).Error("unable to get provisioner")
+						continue
+					}
+
+					err = provisioner.MountBrick(b.Device, b.Name, b.Path)
```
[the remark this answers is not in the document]
aravindavk: that is to temporarily support mounting snapshot provisioned bricks.

pkg/lvmutils/utils.go
```diff
@@ -73,8 +72,8 @@ func GetVgAvailableSize(vgname string) (uint64, uint64, error) {
 	return extentSize * freeExtents, extentSize, nil
 }
 
-// GetPoolMetadataSize calculates the thin pool metadata size based on the given thin pool size
-func GetPoolMetadataSize(poolsize uint64) uint64 {
+// GetTpMetadataSize calculates the thin pool metadata size based on the given thin pool size
+func GetTpMetadataSize(poolsize uint64) uint64 {
```
rafikc30: we should have a minimum value for metadata size, for example, if a user creates with a pool size of 100MB, then metadata will be 0.5MB. We had a recent issue in a user setup where metadata became full because of small size

plugins/lvmprovisioner/provisioner.go
```diff
@@ -0,0 +1,133 @@
...
+
+// MountBrick mounts the brick to the given brick path
+func (p Provisioner) MountBrick(devpath, brickid, brickPath string) error {
+	mountdir := path.Dir(brickPath)
+	brickdev := getBrickDev(devpath, brickid)
+	err := os.MkdirAll(mountdir, os.ModeDir|os.ModePerm)
+	if err != nil {
+		return err
+	}
+	return fsutils.Mount(brickdev, mountdir, mountOpts)
```
rafikc30: I think mount can be variable. Different provisioners of the same type may need to mount with different mount options. For example, in the case of a snapshot, we have to mount using the same options that the parent bricks are used. In this case, the default value won't help

plugins/lvmprovisioner/provisioner.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/gluster/glusterd2/pkg/fsutils"
+	"github.com/gluster/glusterd2/pkg/lvmutils"
+)
+
+var mountOpts = "rw,inode64,noatime,nouuid"
+
+// Provisioner represents lvm provisioner plugin
+type Provisioner struct{}
+
+func getVgName(devpath string) string {
```
rafikc30: can we store the values like vgname, lvname, device name etc. As of now it return constant values, which is not the case with snapshot. 

Or is it possible to move the Mountinfo variable in the brickinfo to provisioner, and return this values based on the values stored

plugins/lvmprovisioner/provisioner.go
```diff
@@ -0,0 +1,133 @@
...
+	err := os.MkdirAll(mountdir, os.ModeDir|os.ModePerm)
+	if err != nil {
+		return err
+	}
+	return fsutils.Mount(brickdev, mountdir, mountOpts)
+}
+
+// UnmountBrick unmounts the brick
+func (p Provisioner) UnmountBrick(brickPath string) error {
+	mountdir := path.Dir(brickPath)
```
rafikc30: This hold true only for smartvol, to extend to the other provisioners, we need to take the mount root from brickpath

glusterd2/provisioners/provisioners.go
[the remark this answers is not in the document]
aravindavk: yeah possible to move bricks mount to volume start. With this change, we should only mount the bricks if `volume.State=="Started"` during glusterd2 start/restart

pkg/lvmutils/utils.go
aravindavk: I will look into that. But this is calculated based on given pool size. 

Refer https://access.redhat.com/documentation/en-us/red_hat_gluster_storage/3.3/html-single/administration_guide/#Brick_Configuration 

Minimum metadata size required is 0.5% and Max upto 16GB

aravindavk: please provide more details about the issue. This function is moved from `plugins/deviceutils` to `pkg/lvmutils`

plugins/lvmprovisioner/provisioner.go
aravindavk: Snapshot mount should be different interface function in my opinion. Bricks of Cloned volume will use this function but not snapshot bricks.

plugins/lvmprovisioner/provisioner.go
aravindavk: I am trying to keep lvm specific things outside the device info. Non LVM use case will not have Vg. For example, loopmount devices provisioner

atinmu:
@rafikc30 have you managed to find time to review this code?

plugins/lvmprovisioner/provisioner.go
rafikc30: Any particular reason why snapshot volume should have a separate mount interface. There is nothing special for snapshot mount. IMHO, we can just take brickinfo.MountInfo as an argument . Becuase this strcture is contains information exclusive to mount,
eg :glusterd2/volume/fs_utils.go: MountDirectory

rafikc30: Any particular reason why snapshot should have separate mount interface. Snapshot mount has nothing special comared to the normal mount. IMHO, we can take brickinfo.MountInfo as an argument, because it is a strcture contains xclusive information about mount. And that is common to any provisoner. Something like glusterd2/volume/fs_utils.go:MountDirectory

plugins/lvmprovisioner/provisioner.go
rafikc30: yea, we can have multiple levels. Totally depends on how users provisioned it. So we use this function volume.GetBrickMountRoot to find that

aravindavk:
Removing GCS tracking label from pull request

pkg/api/volume_req.go
```diff
@@ -2,20 +2,16 @@ package api
...
-	Size           uint64 `json:"size,omitempty"`
-	VgID           string `json:"vg-id,omitempty"`
-	Mountdir       string `json:"mount-dir,omitempty"`
-	DevicePath     string `json:"device-path,omitempty"`
-	MntOpts        string `json:"mnt-opts,omitempty"`
-	FsType         string `json:"fs-type,omitempty"`
+	Type       string `json:"type"`
+	PeerID     string `json:"peerid"`
+	Path       string `json:"path"`
+	Name       string `json:"name,omitempty"`
```
rishubhjain: I guess we do not  store the name of brick in the BrickInfo. Do we really require this?

plugins/device/api/resp.go
```diff
@@ -2,17 +2,18 @@ package api
...
 // Info represents structure in which devices are to be store in Peer Metadata
 type Info struct {
-	Name          string `json:"name"`
-	State         string `json:"state"`
-	VgName        string `json:"vg-name"`
-	AvailableSize uint64 `json:"available-size"`
-	ExtentSize    uint64 `json:"extent-size"`
-	Used          bool   `json:"used"`
-	PeerID        string `json:"peer-id"`
+	Device        string    `json:"device"`
```
rishubhjain: @aravindavk I would like to know why Vgname was removed, it would be useful to know the VG of device via the device struct itself.

plugins/device/rest.go
```diff
@@ -87,35 +86,48 @@ func deviceAddHandler(w http.ResponseWriter, r *http.Request) {
...
+	restutils.SendHTTPResponse(ctx, w, http.StatusOK, deviceInfo)
 }
 
 func deviceListHandler(w http.ResponseWriter, r *http.Request) {
 
 	ctx := r.Context()
 	logger := gdctx.GetReqLogger(ctx)
 	peerID := mux.Vars(r)["peerid"]
-	if uuid.Parse(peerID) == nil {
+	if peerID != "" && uuid.Parse(peerID) == nil {
```
rishubhjain: If the peerID is empty then uuid.Parse() will return nil, so IMO we don't have to put an additional check.

aravindavk:
Updates: #1418

plugins/lvmprovisioner/provisioner.go
JohnStrunk: On LVM it returns space to the TP, and on loop, it'll return space to the outer file system... both are desirable.

PR #1259, opened 2018-10-04
Refactor flags/filters applied to set volume options

### LINKED ISSUE #1078: cli/rest: Rename option name that sets the advanced flag
--advanced in the CLI during volume create doesn't intuitively and clearly indicate that it's for the volume options. A new user may read this as creation of an "advanced volume" which isn't true.

prashanthpai:
'--advanced' in the CLI and REST API during volume create doesn't intuitively
and clearly indicate that it's for the volume options. A new user may read this
as creation of an "advanced volume" which isn't true.

The flags have now been prepended with 'allow' in the fields of request structs.
The API has been renamed as 'allow-<type>-options'.

Also, created 'VolOptionFlags' struct for flags of options and embeded it in
all request structs.

Closes #1078 

### DIFF
--- a/pkg/api/volume_req.go
+++ b/pkg/api/volume_req.go
@@ -62,12 +62,18 @@ type VolCreateReq struct {
 	VolOptionReq
 }
 
+// VolOptionFlags is set of flags that allow/disallow setting certain kinds
+// of volume options.
+type VolOptionFlags struct {
+	AllowAdvanced     bool `json:"allow-advanced-options,omitempty"`
+	AllowExperimental bool `json:"allow-experimental-options,omitempty"`
+	AllowDeprecated   bool `json:"allow-deprecated-options,omitempty"`
+}
+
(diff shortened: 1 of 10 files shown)


amarts:
Also, need to consider '--category' flag we recently introduced in glusterfs. It would help in writing a quick tool to list all the options which are in 'tech-preview' or 'maintained' or 'experimental'...

pkg/api/volume_req.go
```diff
@@ -62,12 +62,18 @@ type VolCreateReq struct {
 	VolOptionReq
 }
 
+// VolOptionFlags is set of flags that allow/disallow setting certain kinds
+// of volume options.
+type VolOptionFlags struct {
+	AllowAdvanced     bool `json:"allow-advanced-options,omitempty"`
```
prashanthpai: @Madhu-1 @shtripat Please make a note of this API change for python mgmt bindings and tendrl/WA

PR #1264, opened 2018-10-08
Notify firewalld over dbus to open/close brick ports

### LINKED ISSUE #447: Add a general dbus interface
We need to use dbus for
* notifying firewalld about port changes
* notify NFS Ganesha of new exports

prashanthpai:
Firewalld is now notified over dbus to open TCP port when a brick signs
in. Correspondingly, the firewalld is notified to close the TCP port
when a brick signs out - gracefully or otherwise.

During glusterd2 shutdown, firewalld is NOT notified to remove port.
This is to ensure that the client can talk to bricks even when
glusterd2 is down.

The package github.com/godbus/dbus has been added as a new dependency
and is available as a RPM package in centos7.

~~Caveats:~~
All firewalld configuration changes are (by choice) runtime changes i.e
these are not permanent. Reloading firewalld configuration will result
in ~~current runtime configuration being wiped out~~ glusterd2
adding the ports again. Glusterd2 detects firewalld reload via dbus
signal.

TODO (future):
Currently firewalld port changes are applied to the default zone. This can be
easily made configurable later if needed.

Fixes #447 


prashanthpai:
retest this please

prashanthpai:
@amarts 

> Can this ever be added to e2e tests?

It can be but it's not straight-forward and hence haven't done it as it will bring in a hard dependency on firewalld. Admin can choose to statically configure a range of ports as well.

> Let me know how to test this!

Sample output:

```sh
# Prerequisite: Start firewalld before starting glusterd2
# sudo systemctl start firewalld
# sudo systemctl enable firewalld

[ppai@gd2-1 glusterd2]$ sudo firewall-cmd --list-ports

[ppai@gd2-1 glusterd2]$ glustercli volume start test
Volume test started successfully

[ppai@gd2-1 glusterd2]$ glustercli volume status test
Volume : test
+--------------------------------------+-----------+---------------------+--------+-------+------+
|               BRICK ID               |   HOST    |        PATH         | ONLINE | PORT  | PID  |
+--------------------------------------+-----------+---------------------+--------+-------+------+
| 354fd1c6-7053-470b-aeaf-a8b6a0679806 | 127.0.0.1 | /export/brick1/data | true   | 44793 | 2316 |
| 56aa0297-e25b-4160-859d-dbae81fbef89 | 127.0.0.1 | /export/brick2/data | true   | 37237 | 2340 |
| a0760d82-9387-42c3-a326-137f365b16c9 | 127.0.0.1 | /export/brick3/data | true   | 33133 | 2366 |
| fb01d613-d4a3-4893-9669-4a97f5603cfc | 127.0.0.1 | /export/brick4/data | true   | 41505 | 2390 |
+--------------------------------------+-----------+---------------------+--------+-------+------+

[ppai@gd2-1 glusterd2]$ sudo firewall-cmd --list-ports
44793/tcp 37237/tcp 33133/tcp 41505/tcp

[ppai@gd2-1 glusterd2]$ glustercli volume stop test
Volume test stopped successfully

[ppai@gd2-1 glusterd2]$ sudo firewall-cmd --list-ports

```

amarts:
Thanks for the output and commands! This is neat! Please add it in user documentation!

@kshlm @JohnStrunk  does anything need to change in gcs installation? Do we take care of firewall config in GCS?

pkg/firewalld/firewalld.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"errors"
+	"strconv"
+
+	"github.com/godbus/dbus"
+)
+
+const (
+	fInterface = "org.fedoraproject.FirewallD1"
```
aravindavk: will this change in case of centos/rhel?

prashanthpai: No. It will be the same.

I just confirmed it on a RHEL 7.3 machine with `man firewalld.dbus` and ran a sample dbus program too.

```sh
[root@dhcp42-40 ~]# cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 7.3 (Maipo)
[root@dhcp42-40 ~]# man firewalld.dbus | grep org.fedoraproject.FirewallD1.zone.Methods.addPort
               org.fedoraproject.FirewallD1.zone.Methods.addPort.
[root@dhcp42-40 ~]# 
```

```
[root@dhcp42-40 ~]# ./dbus 
public
```

aravindavk:
retest this please

pkg/firewalld/firewalld.go
```diff
@@ -0,0 +1,92 @@
...
+		return errors.New("invalid port or protocol")
+	}
+
+	if dbusObj == nil || !isRunning {
+		return nil
+	}
+
+	portStr := strconv.Itoa(port)
+
+	if err := dbusObj.Call(fInterface+".zone.addPort", 0, zone, portStr, string(protocol), 0).Store(&zone); err != nil {
```
Madhu-1: how about just returning `err` here?

pkg/firewalld/firewalld.go
```diff
@@ -0,0 +1,92 @@
...
+		return errors.New("invalid port or protocol")
+	}
+
+	if dbusObj == nil || !isRunning {
+		return nil
+	}
+
+	portStr := strconv.Itoa(port)
+
+	if err := dbusObj.Call(fInterface+".zone.removePort", 0, zone, portStr, string(protocol)).Store(&zone); err != nil {
```
Madhu-1: same here also

pkg/firewalld/firewalld.go
```diff
@@ -0,0 +1,92 @@
...
+var (
+	dbusObj   dbus.BusObject
+	isRunning bool
+)
+
+// AddPort removes a port from runtime configuration of firewalld.
+// If zone is set to empty string, it is applied to the default zone.
+func AddPort(zone string, port int, protocol Protocol) error {
+
+	if port <= 0 || (protocol != ProtoTCP && protocol != ProtoUDP) {
```
Madhu-1: can we do both `max` and `min` port number validation here?

glusterd2/pmap/registry.go
```diff
@@ -148,9 +170,10 @@ func init() {
 	}
 
 	registry = &pmapRegistry{
-		Ports:  make(map[int]brickSet),
-		bricks: make(map[string]int),
-		conns:  make(map[net.Conn]int),
+		Ports:           make(map[int]brickSet),
+		bricks:          make(map[string]int),
+		conns:           make(map[net.Conn]int),
+		notifyFirewalld: true,
```
Madhu-1: is `notifyFirewalld` is configurable?

prashanthpai: No. Not right now. It can be added in future if required.

This is also useful in unit tests of pmap package. It's set to false there. We do not want unit tests to actually configure firewalld.

pkg/firewalld/firewalld.go
prashanthpai: It's already done in pmap package which is brick specific. firewalld package is a generic package.

pkg/firewalld/firewalld.go
prashanthpai: Will do :+1: Some linters catches that.

pkg/firewalld/firewalld.go
prashanthpai: Will do :+1:

pkg/firewalld/firewalld.go
prashanthpai: Done.

pkg/firewalld/firewalld.go
prashanthpai: Done.

JohnStrunk:
> @kshlm @JohnStrunk does anything need to change in gcs installation? Do we take care of firewall config in GCS?

I don't think firewall needs to be configured in kube env as long as we're not using host network stack (and current GCS deploy isn't). Node port (service) or Route could be used to gain access from outside kube, and I would expect that to handle firewall for us (if necessary).

JohnStrunk:
> Caveats:
> All firewalld configuration changes are (by choice) runtime changes i.e
> these are not permanent. Reloading firewalld configuration will result
> in the current runtime configuration being wiped out.

@prashanthpai This makes me nervous. What if the admin upgrades a package on the host (firewall or other) that results in a firewalld restart? How will the admin know this is the cause of their failure, and how will they recover? Or better yet, How can they avoid causing the failure?

prashanthpai:
> How will the admin know this is the cause of their failure, and how will they recover? Or better yet, How can they avoid causing the failure?

Recovering currently involves restarting glusterd2. On glusterd2 restart, bricks sign in, which will result in glusterd2 updating firewalld.

An alternative is to manually and statically configure range of ports to be open in firewalld. That's how it's done today and prior to this PR.

> Node port (service) or Route could be used to gain access from outside kube, and I would expect that to handle firewall for us (if necessary).

Bricks are free to bind on any available free port provided by the kernel in the ephemeral port range. How does the "Node port (service) or Route" get to know about it ?

JohnStrunk:
> > How will the admin know this is the cause of their failure, and how will they recover? Or better yet, How can they avoid causing the failure?
> 
> Recovering currently involves restarting glusterd2. On glusterd2 restart, bricks sign in, which will result in glusterd2 updating firewalld.
> 
> An alternative is to manually and statically configure range of ports to be open in firewalld. That's how it's done today and prior to this PR.

Is there a way to automatically tie this in w/ the restart of firewalld such that admins don't accidentally break their cluster?

Would a viable alternative be to make the rules persistent and have gd2 reconcile when it starts and when bricks (de-)register. I.e., keep a list of the ports we opened, and if there's not a brick there currently, close it and remove it from the opened list?
It's not perfect, but I think it's likely to lead to less catastrophic breakage.

JohnStrunk:
> > Node port (service) or Route could be used to gain access from outside kube, and I would expect that to handle firewall for us (if necessary).
> 
> Bricks are free to bind on any available free port provided by the kernel in the ephemeral port range. How does the "Node port (service) or Route" get to know about it ?

Yeesh... I wasn't thinking about the "any available free port". I think this means if we need to access the gluster cluster from outside the kube cluster, we'll need the firewalld functionality, but instead of talking to firewalld, it'll need to adjust [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) objects.

For normal operation, internal-only access is sufficient, but there are environments where admins want to access the cluster from outside, such as using a legacy backup tool to create backups of PV data. I'm ok if we don't have a solution to this part of it currently, but we should get an issue opened. Backup is important.

prashanthpai:
> Would a viable alternative be to make the rules persistent and have gd2 reconcile when it starts and when bricks (de-)register. I.e., keep a list of the ports we opened, and if there's not a brick there currently, close it and remove it from the opened list?

I considered this earlier. However, list of ports opened is dynamic information, something that glusterd2 prefers not to store persistently (in etcd or otherwise). Hence, the old list of ports to be invalidated in firewalld isn't available when glusterd2 restarts. If a brick goes down when glusterd2 isn't up, there's no way glusterd2 can know about the port to be invalidated.

> Is there a way to automatically tie this in w/ the restart of firewalld such that admins don't accidentally break their cluster?

I'm hoping that there's a provision for a dbus signal notification on firewalld reload. I'll check on that. That seems like a better solution.

 > Yeesh... I wasn't thinking about the "any available free port". I think this means if we need to access the gluster cluster from outside the kube cluster, we'll need the firewalld functionality, but instead of talking to firewalld, it'll need to adjust Ingress objects.

Sure. If brick mux is going be the default at some point, there's likely gonna be one brick process which can bind on a fixed port.

prashanthpai:
> I'm hoping that there's a provision for a dbus signal notification on firewalld reload. I'll check on that. That seems like a better solution.

glusterd2 now detects firewalld reload via dbus signal and reconciles the ports.

glusterd2/main.go
```diff
@@ -130,6 +132,13 @@ func main() {
 	super.ServeBackground()
 	super.Add(servers.New())
 
+	// Start dbus connection (optional for notifying firewalld)
+	if err := firewalld.Init(); err != nil {
+		log.WithError(err).Warn("firewalld.Init() failed")
+	}
+
+	pmap.Init()
```
Madhu-1: question:
do we need to do `pmap.Init()` even if firewall initialization fails?

Madhu-1: okay if initialization of firewalld fails I think even `firewalld.AddPort` won't work
what I was thinking if firewalld initialization fails, why want we set `notifyFirewalld` to `false` after pmap

glusterd2/pmap/registry.go
```diff
@@ -139,18 +163,41 @@ func (r *pmapRegistry) Remove(port int, brickpath string, conn net.Conn) error {
 	return nil
 }
 
+func (r *pmapRegistry) reconcileFirewalld() {
+	sigCh := make(chan *dbus.Signal, 10)
```
Madhu-1: can you  add a comment about why  channel  length is taken as 10

pkg/firewalld/firewalld.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+// Protocol type represents the type of protocol
+type Protocol string
+
+// Supported protocol types
+const (
+	ProtoTCP         Protocol = "tcp"
+	ProtoUDP         Protocol = "udp"
+	dbusSignalChSize          = 10
```
Madhu-1: where  `dbusSignalChSize` is used, am not able to find reference to it

glusterd2/pmap/registry.go
prashanthpai: It is arbitrary. The dbus API requires it to be buffered.

From:
https://godoc.org/github.com/godbus/dbus#Conn.Signal

"The caller has to make sure that ch is sufficiently buffered; if a message arrives when a write to c is not possible, it is discarded."

pkg/firewalld/firewalld.go
prashanthpai: This was intended to be used in reconcileFirewalld(). Will fix that.

glusterd2/main.go
prashanthpai: > think even firewalld.AddPort won't work what I was thinking

That is expected behaviour and is handled inside `firewalld.AddPort`.

> why want we set notifyFirewalld to false after pmap

`notifyFirewalld` is the config option, not a flag to decide to go ahead with dbus notification. There are separate guards for that. The local variable `isRunning` in firewalld package takes care of what you are looking for.

Issue #1000, opened 2018-07-12
Arbiter brick size calculation for Intelligent Volume Provisioning


aravindavk:
Existing feature creates arbiter brick same size as replica brick. Introduce configurable option for average file size with good default value. Apply the formula while planning arbiter brick( https://access.redhat.com/documentation/en-us/red_hat_gluster_storage/3.3/html/administration_guide/creating_arbitrated_replicated_volumes#sect-Arbiter_Requirements )

aravindavk:
Issue #851 is split into multiple issues. This is one among them

Issue #1135, opened 2018-08-06
Volume create should accept Size in bytes


kshlm:
The K8S CSI specifies size in bytes. Which needs to be converted before being passed along to GD2 by the CSI driver. This conversion should not be needed IMO.

In addition to the volume create API, we should ensure that within GD2, in the device management and provisioning we only deal with bytes as well. This is the base unit used by the underlying tools we use (lvm, xfs-tools etc.) and which gives the most accurate control and representation.

When displaying the values to the user, the CLI or another UI can convert it into human friendly format.

prashanthpai:
The sooner we get this cleaned up the better, from API stability standpoint. Bumping up the priority accordingly.

atinmu:
@rishubhjain this is targeted for GCS/0.3 milestone. Where are we w.r.t the implementation?

PR #1267, opened 2018-10-08
Arbiter brick size calculation for smart volume creation.

harigowtham:
This patch creates the arbiter brick for the smart volume as per the
calculation:

brick size = 4 KB * ( size in KB of largest data brick in volume or
replica set / average file size in KB)

### DIFF
--- a/glusterd2/bricksplanner/subvoltype_replicate.go
+++ b/glusterd2/bricksplanner/subvoltype_replicate.go
@@ -2,7 +2,11 @@ package bricksplanner
 
 import (
 	"github.com/gluster/glusterd2/pkg/api"
+	"github.com/gluster/glusterd2/plugins/device/deviceutils"
 )
+
+// defaultLeastArbiterSize is the size (in KB) the arbiter brick will be assigned to if the brick size is less than 100M.
+const defaultLeastArbiterSize = 100000
 
 type replicaSubvolPlanner struct {
 	subvolSize       uint64
@@ -17,8 +21,23 @@ func (s *replicaSubvolPlanner) Init(req *api.VolCreateReq, subvolSize uint64) {
 	s.replicaCount = req.ReplicaCount
 	s.arbiterCount = req.ArbiterCount
 	s.brickSize = s.subvolSize
-	// TODO: Calculate Arbiter brick size
-	s.arbiterBrickSize = s.subvolSize
+	// TODO: the size is calculated in KB, should be changed as per the default.
+	// default avgFileSize needs to be changed from 1M to 64K as well.
+	// Now we are receiving 1M as default from cli so having it as 1M here as well,
+	var avgFileSize uint64 = 1024
+	if req.AverageFileSize != 0 {
+		avgFileSize = deviceutils.MbToKb(req.AverageFileSize)
+	}
+	arbiterSize := uint64((4.0) * (float64(subvolSize) / float64(avgFileSize)))
+	// Assigning arbiter brick size to be bricksize if its lesser than 100M
+	if arbiterSize < defaultLeastArbiterSize {
+		if defaultLeastArbiterSize > subvolSize {
+			arbiterSize = subvolSize
+		} else {
+			arbiterSize = defaultLeastArbiterSize
+		}
+	}
+	s.arbiterBrickSize = arbiterSize
 }
 
 func (s *replicaSubvolPlanner) BricksCount() int {
--- a/glusterd2/commands/volumes/volume-smartvol-txn.go
+++ b/glusterd2/commands/volumes/volume-smartvol-txn.go
@@ -59,7 +59,13 @@ func txnPrepareBricks(c transaction.TxnCtx) error {
 			}
 
 			// Make Filesystem
-			err = deviceutils.MakeXfs(b.DevicePath)
+			mkfsOpts := []string{}
+			if b.Type == "arbiter" {
+				mkfsOpts = []string{"-i", "size=512", "-n", "size=8192", "-i", "maxpct=0"}
+			} else {
+				mkfsOpts = []string{"-i", "size=512", "-n", "size=8192"}
+			}
+			err = deviceutils.MakeXfs(b.DevicePath, mkfsOpts...)
 			if err != nil {
 				c.Logger().WithError(err).WithField("dev", b.DevicePath).Error("mkfs.xfs failed")
 				return err
--- a/plugins/device/deviceutils/utils.go
+++ b/plugins/device/deviceutils/utils.go
@@ -81,7 +81,10 @@ func GetPoolMetadataSize(poolsize uint64) uint64 {
 	metadataSize := uint64(float64(poolsize) * 0.005)
 	if metadataSize > GbToKb(maxMetadataSizeGb) {
 		metadataSize = GbToKb(maxMetadataSizeGb)
+	} else if metadataSize == 0 {
+		metadataSize = 64
 	}
+
 	return metadataSize
 }
 
@@ -108,12 +111,11 @@ func CreateLV(vgname, tpname, lvname string, lvsize uint64) error {
 }
 
 // MakeXfs creates XFS filesystem
-func MakeXfs(dev string) error {
+func MakeXfs(dev string, mkfsOpts ...string) error {
+	mkfsOpts = append([]string{dev}, mkfsOpts...)
 	// TODO: Adjust -d su=<>,sw=<> based on RAID/JBOD
 	return utils.ExecuteCommandRun("mkfs.xfs",
-		"-i", "size=512",
-		"-n", "size=8192",
-		dev,
+		mkfsOpts...,
 	)
 }
 
--- a/glustercli/cmd/volume-create.go
+++ b/glustercli/cmd/volume-create.go
@@ -39,6 +39,7 @@ var (
 	flagCreateSnapshotEnabled       bool
 	flagCreateSnapshotReserveFactor float64 = 1
 	flagCreateSubvolZoneOverlap     bool
+	flagAverageFileSize             string
 
 	volumeCreateCmd = &cobra.Command{
 		Use:   "create <volname> [<brick> [<brick>]...|--size <size>]",
@@ -83,6 +84,7 @@ func init() {
 	volumeCreateCmd.Flags().BoolVar(&flagCreateSnapshotEnabled, "enable-snapshot", false, "Enable Volume for Gluster Snapshot")
 	volumeCreateCmd.Flags().Float64Var(&flagCreateSnapshotReserveFactor, "snapshot-reserve-factor", 1, "Snapshot Reserve Factor")
 	volumeCreateCmd.Flags().BoolVar(&flagCreateSubvolZoneOverlap, "subvols-zones-overlap", false, "Brick belonging to other Sub volume can be created in the same zone")
+	volumeCreateCmd.Flags().StringVar(&flagAverageFileSize, "average-file-size", "1M", "Average size of the files")
 
 	volumeCmd.AddCommand(volumeCreateCmd)
 }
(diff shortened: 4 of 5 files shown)


glusterd2/commands/volumes/volume-smartvol-txn.go
```diff
@@ -48,15 +48,35 @@ func txnPrepareBricks(c transaction.TxnCtx) error {
...
-				c.Logger().WithError(err).WithFields(log.Fields{
-					"vg-name": b.VgName,
-					"tp-name": b.TpName,
-					"lv-name": b.LvName,
-					"size":    b.Size,
-				}).Error("lvcreate failed")
-				return err
+			if b.Type == "arbiter" {
+				// All the calculations should be in MB
+				arbiterSize := (4 / 1024) * (b.Size / req.AverageFileSize)
```
aravindavk: Move arbiter brick calculations here `$SOURCE/glusterd2/bricksplanner/subvoltype_replicate.go`

glusterd2/commands/volumes/volume-smartvol-txn.go
```diff
@@ -48,15 +48,35 @@ func txnPrepareBricks(c transaction.TxnCtx) error {
...
+				// Assigning arbiter brick size to be 20Mb if its lesser than 20Mb
+				if arbiterSize < 20 {
```
Madhu-1: * is there any specific reason to select `average size` as 20 MB?
in heketi we use `64KB` [heketi-avegare-size](https://github.com/heketi/heketi/blob/63154832abfe16d7d9ebd2f5ded798b0f6fed725/apps/glusterfs/volume_entry.go#L51)

* I think we should not set arbiter size to a default value if its less than 20MB. we should make volume create to fail.

aravindavk: 20 MB is not average file size here. If the calculated arbiter brick size is less than 20MB, then create brick with 20MB. I think default size of file is not yet considered in this PR

Madhu-1: if the calculated size is less than `20MB` don't we need to fail it?

aravindavk: No. It is like minimum size required to create lv and xfs is 20MB(I think we can increase the default to 100MB). 

Maximum arbiter brick size possible = Size of data brick

glusterd2/commands/volumes/volume-smartvol-txn.go
Madhu-1: one more question, what happens if the `req.AverageFileSize` is 0 from rest call?

JohnStrunk:
Caveat: I have not looked in detail at the code.
Are you ensuring there are sufficient inodes available on the small arbiter volume? We should make sure the arbiter volume is created with a comparable number to the full-sized volume. (IIRC, heketi ran into this problem).

glusterd2/commands/volumes/volume-smartvol-txn.go
harigowtham: We have a default value for average file size as 10M. This value is used if the command doesn't have the average file size argument in the creation of replica volume with arbiter brick.
The arbiter brick is always calculated as per this patch. The only change done is, If the arbiter brick size is less than 20M then the size of the brick is itself used as the arbiter brick size.

I haven't failed the create if the arbiter size is calculated to be less than 20M instead I have used it as the size of the brick and passed the creation.

Regarding the average file size, I used 10M as I didn't find any recommended value. As the sizeToMb returns zero for values less than 1M I have changed the average file size to 1M which wont error out and will be the closest to 64K. This is a TODO. Once we change the default value to a unit lesser than M the support for K can be done.

glusterd2/commands/volumes/volume-smartvol-txn.go
harigowtham: done.

harigowtham: Nice catch. Have fixed this case.

harigowtham:
> Caveat: I have not looked in detail at the code.
> Are you ensuring there are sufficient inodes available on the small arbiter volume? We should make sure the arbiter volume is created with a comparable number to the full-sized volume. (IIRC, heketi ran into this problem).

For smaller volumes, the arbiter size is taken as the size of brick itself.
the smallest size of arbiter brick can be 20M. I hope this will give it sufficient inodes on the arbiter brick.

JohnStrunk:
By default, for small volumes, xfs permits up to 25% of volume capacity to be inodes. In the case of arbiter bricks, it's basically all inodes. Suggest `-i maxpct=0` should be passed to `mkfs.xfs` to remove the cap on inode percentage.

See:
- https://github.com/heketi/heketi/issues/1296
- https://bugzilla.redhat.com/show_bug.cgi?id=1612058

glusterd2/bricksplanner/subvoltype_replicate.go
```diff
@@ -44,3 +46,18 @@ func (s *replicaSubvolPlanner) BrickType(idx int) string {
...
+// ArbiterBrickSizeCalculation returns the arbiter brick size for that particular volume
+func ArbiterBrickSizeCalculation(brickSize uint64, avgFileSize uint64) (uint64, error) {
+	if avgFileSize == 0 {
+		return 0, errs.New("average file size is passed as zero")
+	}
+	// All the calculations should be in MB
+	arbiterSize := (4 / 1024) * (brickSize / avgFileSize)
+	// Assigning arbiter brick size to be bricksize if its lesser than 20M
+	// TODO: the size is calculated in MB, should be changed as per the default.
+	if arbiterSize < 20 {
```
aravindavk: Do not hard code 20 here. Use the constant defined. I feel we should increase the minimum volume size to `100MB`

glusterd2/bricksplanner/subvoltype_replicate.go
```diff
@@ -44,3 +46,18 @@ func (s *replicaSubvolPlanner) BrickType(idx int) string {
 func init() {
 	subvolPlanners["replicate"] = &replicaSubvolPlanner{}
 }
+
+// ArbiterBrickSizeCalculation returns the arbiter brick size for that particular volume
+func ArbiterBrickSizeCalculation(brickSize uint64, avgFileSize uint64) (uint64, error) {
```
aravindavk: Do not add new function here, add calculation where following `TODO: Calculate Arbiter brick size` is present in this file.

glusterd2/commands/volumes/volume-smartvol-txn.go
```diff
@@ -48,15 +49,33 @@ func txnPrepareBricks(c transaction.TxnCtx) error {
...
-			err = deviceutils.CreateLV(b.VgName, b.TpName, b.LvName, b.Size)
-			if err != nil {
-				c.Logger().WithError(err).WithFields(log.Fields{
-					"vg-name": b.VgName,
-					"tp-name": b.TpName,
-					"lv-name": b.LvName,
-					"size":    b.Size,
-				}).Error("lvcreate failed")
-				return err
+			if b.Type == "arbiter" {
```
aravindavk: this condition is not required since Size already updated in brick request

glusterd2/commands/volumes/volume-smartvol-txn.go
```diff
@@ -48,15 +49,33 @@ func txnPrepareBricks(c transaction.TxnCtx) error {
...
+				err = deviceutils.CreateLV(b.VgName, b.TpName, b.LvName, b.Size)
+				if err != nil {
+					c.Logger().WithError(err).WithFields(log.Fields{
+						"vg-name": b.VgName,
+						"tp-name": b.TpName,
+						"lv-name": b.LvName,
+						"size":    b.Size,
+					}).Error("lvcreate failed")
+					return err
+				}
```
aravindavk: Add mkfs option if b.Type is arbiter where `deviceutils.MakeXfs` is called, as @JohnStrunk suggested

glusterd2/bricksplanner/subvoltype_replicate.go
```diff
@@ -44,3 +46,18 @@ func (s *replicaSubvolPlanner) BrickType(idx int) string {
...
+	if avgFileSize == 0 {
```
aravindavk: If avgFileSize is zero then consider default value `64K`

harigowtham: As all the values are calculated in M, if we specify it in K then it will be ~0.064M which will be considered as zero because the size is an uint64. we need these to be changed once the default size is in bytes.

aravindavk: define constant of 1M for now. Once this issue https://github.com/gluster/glusterd2/issues/1135 fixed, we can change that to 64K

Madhu-1:
@harigowtham can you address the Review comments on this PR?

Madhu-1:
@harigowtham can this be done by today?
if not possible, we need to move to GCS0.3

glusterd2/commands/volumes/volume-smartvol-txn.go
```diff
@@ -60,10 +60,18 @@ func txnPrepareBricks(c transaction.TxnCtx) error {
 			}
 
 			// Make Filesystem
-			err = deviceutils.MakeXfs(b.DevicePath)
-			if err != nil {
-				c.Logger().WithError(err).WithField("dev", b.DevicePath).Error("mkfs.xfs failed")
-				return err
+			if b.Type == "arbiter" {
+				err = deviceutils.MakeXfsArbiter(b.DevicePath)
```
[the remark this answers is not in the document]
harigowtham: done

harigowtham:
@Madhu-1 Have updated the patch. Sorry couldn't reply earlier.

aravindavk:
Fixes: #1000

aravindavk:
Removing GCS tracking label from pull request and adding it to respective issue(#1000)

glusterd2/bricksplanner/subvoltype_replicate.go
```diff
@@ -17,8 +20,19 @@ func (s *replicaSubvolPlanner) Init(req *api.VolCreateReq, subvolSize uint64) {
...
+	// default avgFileSize needs to be changed from 1M to 64K as well.
+	// Now we are receiving 1M as default from cli so having it as 1M here as well,
+	var avgFileSize uint64 = 1024
+	if req.AverageFileSize != 0 {
+		avgFileSize = deviceutils.MbToKb(req.AverageFileSize)
+	}
+	arbiterSize := uint64((4.0) * (float64(subvolSize) / float64(avgFileSize)))
+	// Assigning arbiter brick size to be bricksize if its lesser than 100M
+	if arbiterSize < defaultLeastArbiterSize {
+		arbiterSize = subvolSize
```
aravindavk:
```
if arbiterSize < defaultLeastArbiterSize {
     if defaultLeastArbiterSize > subvolSize {
        arbiterSize = subvolSize
    } else {
        arbiterSize = defaultLeastArbiterSize
    }
}
```

glusterd2/commands/volumes/volume-smartvol-txn.go
```diff
@@ -60,7 +60,13 @@ func txnPrepareBricks(c transaction.TxnCtx) error {
...
 
 			// Make Filesystem
-			err = deviceutils.MakeXfs(b.DevicePath)
+			mkfsOpts := []string{}
+			if b.Type == "arbiter" {
+				mkfsOpts = []string{"-i", "size=512", "-n", "size=8192", "-i", "maxpct=0", b.DevicePath}
+			} else {
+				mkfsOpts = []string{"-i", "size=512", "-n", "size=8192", b.DevicePath}
+			}
+			err = deviceutils.MakeXfs(mkfsOpts...)
```
aravindavk: Mixing devicePath with options is not looking good, retain first argument as devicePath  `deviceutils.MakeXfs(b.DevicePath, mkfsOpts...)`

plugins/device/deviceutils/utils.go
```diff
@@ -108,12 +111,10 @@ func CreateLV(vgname, tpname, lvname string, lvsize uint64) error {
 }
 
 // MakeXfs creates XFS filesystem
-func MakeXfs(dev string) error {
+func MakeXfs(mkfsOpts ...string) error {
```
aravindavk: see comment above

Madhu-1:
@harigowtham can you fix minor comments?

Madhu-1:
@harigowtham ping

plugins/device/deviceutils/utils.go
harigowtham: done

glusterd2/commands/volumes/volume-smartvol-txn.go
harigowtham: done

glusterd2/bricksplanner/subvoltype_replicate.go
harigowtham: done

harigowtham:
@Madhu-1 I have updated the patch.

glusterd2/bricksplanner/subvoltype_replicate.go
```diff
@@ -2,8 +2,11 @@ package bricksplanner
 
 import (
 	"github.com/gluster/glusterd2/pkg/api"
+	"github.com/gluster/glusterd2/plugins/device/deviceutils"
 )
 
+const defaultLeastArbiterSize = 100000
```
Madhu-1: add comment for this variable

Issue #376, opened 2017-09-22
Revisit option name that turns on xlators


prashanthpai:
The xlator option table can also contain options that load (add to graph) or turn on an xlator. For example, to turn on the barrier xlator, as of today, one has to set option named `barrier.barrier` to on. This should rather simply be just `barrier`. IOW, we need a way to identify such options. One approach is to let xlators themselves define which option in option table is a "gobal switch" resulting in the xlator itself being turned on. We can use the newly introduced [flags](https://github.com/gluster/glusterfs/blob/master/libglusterfs/src/options.h#L110) field in the `volume_option_t` struct.

amarts:
I was thinking of solving it by approach mentioned in  https://github.com/gluster/glusterfs/issues/304 Can you check if that solves your problem?

aravindavk:
With the templates support for Volgen(https://github.com/gluster/glusterd2/pull/1229). A Xlator can be enabled by setting `[<graph-name>].<option-name>=on` in Volinfo. (Volume Set/Reset need to be accommodated once template support merges)

Some Xlators understands enable by having a option in the graph as below,

    volume gv7-changelog
        type features/changelog
        option changelog on
        subvolumes gv7-changetimerecorder
    end-volume

And some of the Xlators doesn't need any option to say it is enabled. If xlator present in volfile means enabled. This can be handled by setting `EnabledByOption` field in Volfile template.

aravindavk:
With Glusterfs Patch https://review.gluster.org/21813 and Glusterd2 patch https://github.com/gluster/glusterd2/pull/1370 we can enable/disable xlators if option name with same name as xlator exists in respective options table.

PR #1282, opened 2018-10-15
Add group profile for transactional DB workload

### LINKED ISSUE #1250: group profile for transactional db workload
group profile infrastructure already exists in GD2, however we'd need to have a profile for db workload tunings to be included for GD2 which should be as follows:

performance.open-behind=on
performance.write-behind=off
performance.stat-prefetch=off
performance.quick-read=off
performance.strict-o-direct=on
performance.read-ahead=off
performance.io-cache=off
performance.readdir-ahead=off
performance.client-io-threads=on
server.event-threads=4
client.event-threads=4
performance.read-after-open=yes

### Observed behavior
No db profile is currently present

### Expected/desired behavior
DB profile should be present

- Glusterd2 version used (e.g. v4.1.0 or master): 
- Operating system used: 
- Glusterd2 compiled from sources, as a package (rpm/deb), or container: 
- Using External ETCD: (yes/no, if yes ETCD version):

### Other useful information

- ETCD configuration

### Useful commands

- To get glusterd2 version
    ```
    glusterd2 --version
    ```
- To get ETCD version
    ```
    etcd --version
    ```
- To get output of statedump
    ```
    ```

prashanthpai:
For this to work, it needs support from volgen for providing a way to
turn an xlator on or off by its name.

Fixes #1250 
Updates #376 

### DIFF
--- a/glusterd2/commands/volumes/grouped-options.go
+++ b/glusterd2/commands/volumes/grouped-options.go
@@ -129,6 +129,28 @@ var defaultGroupOptions = map[string]*api.OptionGroup{
 			{Name: "features.shard", OnValue: "on"},
 			{Name: "user.cifs", OnValue: "off"}},
 		"Enable this profile, if the Gluster Volume is used to store virtual machines"},
+	"profile.db": {"profile.db",
+		[]api.VolumeOption{
+			// enable/disable xlators (depends on volgen support PR #1229)
+			{Name: "open-behind", OnValue: "on"},
+			{Name: "write-behind", OnValue: "off"},
+			{Name: "quick-read", OnValue: "off"},
+			{Name: "read-ahead", OnValue: "off"},
+			{Name: "io-cache", OnValue: "off"},
+			{Name: "readdir-ahead", OnValue: "off"},
+			// as per new volgen, this has been named 'md-cache' instead of its
+			// legacy name 'stat-prefetch'
+			{Name: "md-cache", OnValue: "off"},
+			// FIXME: how should this option be named ? fuse.io-threads ?
+			// https://github.com/gluster/glusterd2/issues/376#issuecomment-424970778
+			{Name: "client-io-threads", OnValue: "on"},
+			// xlator options
+			{Name: "server.event-threads", OnValue: "4"},
+			{Name: "client.event-threads", OnValue: "4"},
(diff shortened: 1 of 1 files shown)


glusterd2/commands/volumes/grouped-options.go
```diff
@@ -129,6 +129,28 @@ var defaultGroupOptions = map[string]*api.OptionGroup{
...
+			{Name: "quick-read", OnValue: "off"},
+			{Name: "read-ahead", OnValue: "off"},
+			{Name: "io-cache", OnValue: "off"},
+			{Name: "readdir-ahead", OnValue: "off"},
+			// as per new volgen, this has been named 'md-cache' instead of its
+			// legacy name 'stat-prefetch'
+			{Name: "md-cache", OnValue: "off"},
+			// FIXME: how should this option be named ? fuse.io-threads ?
+			// https://github.com/gluster/glusterd2/issues/376#issuecomment-424970778
+			{Name: "client-io-threads", OnValue: "on"},
```
prashanthpai: Do not merge this until this is finalised.

prashanthpai: @aravindavk Thoughts on this ?

aravindavk: once template based volgen PR merges, xlator can be enabled/disabled like this. But only if the xlator name exists in the template.

prashanthpai: In this specific case, io-threads can be loaded on the brick (default) and also on the client (additional and optional). If a user has to turn on io-threads on the client, what will the option look like ?

Madhu-1:
@aravindavk can we take this once we get `volgen` patch in?

glusterd2/commands/volumes/grouped-options.go
prashanthpai: @aravindavk Can you please clarify ?

aravindavk:
Sent glusterfs patch to include a option in xlator option table to enable/disable xlators. https://review.gluster.org/21813

This PR also needs changes to some of the names used here, I will submit new PR to address those issues.

aravindavk:
PR sent to address the option names and fix a few validation issues related to enable/disable xlators. https://github.com/gluster/glusterd2/pull/1370

PR #1290, opened 2018-10-17
Add qsh to the list of required claims.

pdrakeweb:
qsh was added as a claim that must be included but was not added to the required claims validation.  This results in a confusing error message: 'invalid qsh claim in token' when no qsh claim is included.  With this PR, the error response will include the more informative error message: 'Token missing qsh Claim'.

### DIFF
--- a/glusterd2/middleware/auth.go
+++ b/glusterd2/middleware/auth.go
@@ -18,7 +18,7 @@ const (
 )
 
 var (
-	requiredClaims = []string{"iss", "exp"}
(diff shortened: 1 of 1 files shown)


Madhu-1:
@centos-ci ok to test

Madhu-1:
@pdrakeweb Thanks for the PR, @prashanthpai can you trigger @centos-ci

aravindavk:
add to whitelist

Issue #219, opened 2017-01-13
Brick multiplexing support


prashanthpai:
Placeholder issue to track adding support for brick multiplexing.

purpleidea:
On Fri, Jan 13, 2017 at 1:14 AM, Prashanth Pai <notifications@github.com>
wrote:

> Placeholder issue to track adding support for brick multiplexing.



Is there a core (glusterfs) upstream ticket for this issue or any
design/docs page about it? Thanks!

prashanthpai:
@purpleidea Here's everything that's out there:

* [Feature page](https://github.com/gluster/glusterfs-specs/blob/master/under_review/multiplexing.md) and tracker [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1385758)
* [Detailed walthrough](https://docs.google.com/document/d/1yOoLv1zMCShd8PUFnGGKzZsOpshvFmgJxIwLERhErds/edit#heading=h.vam0sl52njw) of the implementation
* [Implementation](http://review.gluster.org/#/c/14763/) posted for review

purpleidea:
@prashanthpai Great links, thanks!

atinmu:
@raghavendra-talur I'd like you to pick this up for GCS-Sprint1.

atinmu:
@raghavendra-talur Circling back to see if the first cut document is ready for review?

prashanthpai:
I have opened the following separate issues for pending stuff (not GCS focus) not addressed by PR #219:

* Extend brick multiplexing to volume expand (add brick) operation: https://github.com/gluster/glusterd2/issues/1303
* Implement 'cluster.max-bricks-per-process' option: https://github.com/gluster/glusterd2/issues/1304

atinmu:
This can't be completed by GCS/0.3 release. Moving it to 0.4.

amarts:
Note that relevant GlusterFS patch is now merged!

PR #1293, opened 2018-10-18
Improve getting/setting cluster options

prashanthpai:
* Take lock when setting/updating cluster option.
 * Add support for custom validation function.
 * Fix crash in GetClusterOption() due to nil pointer access.

The first consumer of these fixes and improvements shall be brick multiplexing.
Updates #219


prashanthpai:
> LGTM, do we have any e2e tests to cover this one?

No. This is mostly setting up infra and refactor. The features (such as brickmux) that actually use these cluster options are yet to be implemented.

PR #1300, opened 2018-10-26
Tracing: Trace volume option set command.

### LINKED ISSUE #1299: Tracing: Trace volume option set command.
Add instrumentation to trace the volume option set command. Add relevant metadata to the spans like request-id, volume name and the options being set along with the new values.

### Observed behavior
NA

### Expected/desired behavior
span must also show the requestID, volume name and the options being set along with new values as part of the request on the tracing UI.

- Operating system used: fedora27
- Using External ETCD: (yes/no, if yes ETCD version): no

### Other useful information

- ETCD configuration

### Useful commands

- To get glusterd2 version
    ```
    glusterd2 --version
    ```
- To get ETCD version
    ```
    etcd --version
    ```
- To get output of statedump
    ```
    ```

sseshasa:
Add instrumentation to trace volume option set transaction. Add relevant
attributes to the trace like request ID, volume name and the options along
with the values to be set on the volume. This is to help in debugging in
case the command fails for some reason. The instrumentation to trace steps
of this transaction is already in place in the common transaction and
steps framework code.

Closes  #1299


glusterd2/commands/volumes/volume-option.go
```diff
@@ -262,6 +265,18 @@ func volumeOptionsHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Add relevant attributes to the span
+	var optionToSet string
+	for option, value := range req.Options {
+		optionToSet += option + "=" + value + ":"
```
Madhu-1: question:
is it good to separate options by comma other than colon?

sseshasa: Earlier I was just including the option name and so I chose colon so that distinguishing the dots in the option name from the comma separator might be a bit difficult. But since I also include the value now, I guess comma should be fine. I will change it.

Issue #1306, opened 2018-10-31
heal info fails when one brick is killed in a volume


vpandey-RH:
Heal Info fails when one of the bricks of a volume is not online.

### Observed behavior
strconv.ParseInt: parsing "-": invalid syntax

### Expected/desired behavior
Heal Info should successfully reflect heal related information for all online bricks and proper message like "Transport endpoint not connectd for the killed brick"

### Details on how to reproduce (minimal and precise)
- Create a replica volume
- Start the volume and check heal info on the volume
- Now kill one of the bricks of the volume
- Run heal info command again

### Information about the environment:

- Glusterd2 version used (e.g. v4.1.0 or master): 
- Operating system used: any
- Glusterd2 compiled from sources, as a package (rpm/deb), or container: source
- Using External ETCD: (yes/no, if yes ETCD version): external
- If container, which container image: 
- Using kubernetes, openshift, or direct install: 
- If kubernetes/openshift, is gluster running inside kubernetes/openshift or outside: 

### Other useful information

- glusterd2 config files from all nodes (default /etc/glusterd2/glusterd2.toml)
- glusterd2 log files from all nodes (default /var/log/glusterd2/glusterd2.log)
- ETCD configuration
- Contents of `uuid.toml` from all nodes (default /var/lib/glusterd2/uuid.toml)
- Output of `statedump` from any one of the node

### Useful commands

- To get glusterd2 version
    ```
    glusterd2 --version
    ```
- To get ETCD version
    ```
    etcd --version
    ```
- To get output of statedump
    ```
    curl http://glusterd2-IP:glusterd2-Port/statedump
    ```

PR #1310, opened 2018-10-31
Heal-info command fails when any one of the bricks is offline

vpandey-RH:
Fixes [#1306](https://github.com/gluster/glusterd2/issues/1306)

### DIFF
--- a/plugins/glustershd/api/resp.go
+++ b/plugins/glustershd/api/resp.go
@@ -12,15 +12,20 @@ type FileGfID struct {
 
 // BrickHealInfo represents brick details for Heal Info command
 type BrickHealInfo struct {
-	HostID                 string     `xml:"hostUuid,attr" json:"host-id"`
-	Name                   string     `xml:"name" json:"name"`
-	Status                 string     `xml:"status" json:"status"`
-	TotalEntries           *int       `xml:"totalNumberOfEntries" json:"total-entries,omitempty"`
-	EntriesInHealPending   *int       `xml:"numberOfEntriesInHealPending" json:"entries-in-heal-pending,omitempty"`
-	EntriesInSplitBrain    *int       `xml:"numberOfEntriesInSplitBrain" json:"entries-in-split-brain,omitempty"`
-	EntriesPossiblyHealing *int       `xml:"numberOfEntriesPossiblyHealing" json:"entries-possibly-healing,omitempty"`
-	Entries                *int       `xml:"numberOfEntries" json:"entries,omitempty"`
-	Files                  []FileGfID `xml:"file" json:"file-gfid,omitempty"`
+	HostID                    string     `xml:"hostUuid,attr" json:"host-id"`
+	Name                      string     `xml:"name" json:"name"`
+	Status                    string     `xml:"status" json:"status"`
+	TotalEntriesRaw           string     `xml:"totalNumberOfEntries" json:"-"`
+	EntriesInHealPendingRaw   string     `xml:"numberOfEntriesInHealPending" json:"-"`
+	EntriesInSplitBrainRaw    string     `xml:"numberOfEntriesInSplitBrain" json:"-"`
+	EntriesPossiblyHealingRaw string     `xml:"numberOfEntriesPossiblyHealing" json:"-"`
+	EntriesRaw                string     `xml:"numberOfEntries" json:"-"`
+	TotalEntries              *int64     `json:"total-entries,omitempty"`
+	EntriesInHealPending      *int64     `json:"entries-in-heal-pending,omitempty"`
+	EntriesInSplitBrain       *int64     `json:"entries-in-split-brain,omitempty"`
+	EntriesPossiblyHealing    *int64     `json:"entries-possibly-healing,omitempty"`
+	Entries                   *int64     `json:"entries,omitempty"`
+	Files                     []FileGfID `xml:"file" json:"file-gfid,omitempty"`
 }
 
 // HealInfo represents structure of stdout while running glfsheal binary
--- a/plugins/glustershd/rest.go
+++ b/plugins/glustershd/rest.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os/exec"
 	"path"
+	"strconv"
 	"strings"
 
 	"github.com/gluster/glusterd2/glusterd2/gdctx"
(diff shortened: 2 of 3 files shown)


Madhu-1:
add to whitelist

Madhu-1:
ok to test

plugins/glustershd/api/resp.go
```diff
@@ -2,6 +2,7 @@ package api
 
 import (
 	"encoding/xml"
+	//"github.com/pborman/uuid"
```
Madhu-1: remove this line

plugins/glustershd/api/resp.go
```diff
@@ -15,11 +16,11 @@ type BrickHealInfo struct {
 	HostID                 string     `xml:"hostUuid,attr" json:"host-id"`
 	Name                   string     `xml:"name" json:"name"`
 	Status                 string     `xml:"status" json:"status"`
-	TotalEntries           *int       `xml:"totalNumberOfEntries" json:"total-entries,omitempty"`
-	EntriesInHealPending   *int       `xml:"numberOfEntriesInHealPending" json:"entries-in-heal-pending,omitempty"`
-	EntriesInSplitBrain    *int       `xml:"numberOfEntriesInSplitBrain" json:"entries-in-split-brain,omitempty"`
-	EntriesPossiblyHealing *int       `xml:"numberOfEntriesPossiblyHealing" json:"entries-possibly-healing,omitempty"`
-	Entries                *int       `xml:"numberOfEntries" json:"entries,omitempty"`
+	TotalEntries           *string    `xml:"totalNumberOfEntries" json:"total-entries,omitempty"`
```
Madhu-1: why these are converted into a string? can you update the commit message with more info

aravindavk: keeping it `int` may be good in the response. Use `string` while parsing xml output and add validation to check `-`

```
TotalEntriesRaw *string `xml:"totalNumberOfEntries" json:"-"`
TotalEntries *int `json:"total-entries,omitempty"`
```

and update,

```
if healinfo.TotalEntriesRaw == "-" {
    healinfo.TotalEntries = -1
} else {
    healinfo.TotalEntries = parseint(healinfo.TotalEntriesRaw)
}
```

vpandey-RH: @aravindavk  But we are not using these entries in any kind of mathematical operations, these are used only while displaying heal info, so I am not sure if adding so many if conditions just to handle this is okay. Its not just TotalEntries there are other fields that I will need to handle which will increase the lines of code and can be avoided here.

plugins/glustershd/api/resp.go
```diff
@@ -12,15 +12,20 @@ type FileGfID struct {
...
-	EntriesPossiblyHealing *int       `xml:"numberOfEntriesPossiblyHealing" json:"entries-possibly-healing,omitempty"`
-	Entries                *int       `xml:"numberOfEntries" json:"entries,omitempty"`
-	Files                  []FileGfID `xml:"file" json:"file-gfid,omitempty"`
+	HostID                    string     `xml:"hostUuid,attr" json:"host-id"`
+	Name                      string     `xml:"name" json:"name"`
+	Status                    string     `xml:"status" json:"status" json:"-"`
+	TotalEntriesRaw           string     `xml:"totalNumberOfEntries" json:"-"`
+	EntriesInHealPendingRaw   string     `xml:"numberOfEntriesInHealPending" json:"-"`
+	EntriesInSplitBrainRaw    string     `xml:"numberOfEntriesInSplitBrain" json:"-"`
+	EntriesPossiblyHealingRaw string     `xml:"numberOfEntriesPossiblyHealing" json:"-"`
```
rishubhjain: I feel the field names are pretty big. I am not sure if its a good way to program in golang. I will like to know the opinion of @Madhu-1 and @aravindavk .

plugins/glustershd/rest.go
```diff
@@ -63,6 +64,73 @@ func getHealInfo(volname string, option string) (string, error) {
...
+				*healInfo.Bricks[i].TotalEntries = -1
+			} else {
+				*healInfo.Bricks[i].TotalEntries, err = strconv.ParseInt(healInfo.Bricks[i].TotalEntriesRaw, 10, 64)
+				if err != nil {
+					return healInfo, err
+				}
+			}
+		}
+
+		if healInfo.Bricks[i].EntriesInHealPendingRaw != "" {
```
aravindavk: create a function to avoid duplicates.

```
func convertToNumber(value string) int64 {

    if value == "" {
        return 0
    }

    if value == "-" {
        return -1
    }
    val, err := strconv.ParseInt(value, 10, 64)
    if err != nil {
        return -1
    }
    return val
}

File: plugins/glustershd/api/resp.go [hunk 5]
```diff
@@ -12,15 +12,20 @@ type FileGfID struct {
...
-	Status                 string     `xml:"status" json:"status"`
-	TotalEntries           *int       `xml:"totalNumberOfEntries" json:"total-entries,omitempty"`
-	EntriesInHealPending   *int       `xml:"numberOfEntriesInHealPending" json:"entries-in-heal-pending,omitempty"`
-	EntriesInSplitBrain    *int       `xml:"numberOfEntriesInSplitBrain" json:"entries-in-split-brain,omitempty"`
-	EntriesPossiblyHealing *int       `xml:"numberOfEntriesPossiblyHealing" json:"entries-possibly-healing,omitempty"`
-	Entries                *int       `xml:"numberOfEntries" json:"entries,omitempty"`
-	Files                  []FileGfID `xml:"file" json:"file-gfid,omitempty"`
+	HostID                    string     `xml:"hostUuid,attr" json:"host-id"`
+	Name                      string     `xml:"name" json:"name"`
+	Status                    string     `xml:"status" json:"status" json:"-"`
```
aravindavk: Is `status` not required in json output?

vpandey-RH: Sorr, my bad. It is required.

plugins/glustershd/rest.go
vpandey-RH: Done

plugins/glustershd/api/resp.go
vpandey-RH: Done

atinmu:
@vpandey-RH @Madhu-1 What's pending on this?

Madhu-1:
@aravindavk PTAL

PR #1317, opened 2018-11-06
Remove sdfs xlator

Madhu-1:


Madhu-1:
@aravindavk CI passed, do we have any other way to find out what caused this issue?

aravindavk:
Merging this PR, if fails with any other xlator then we need to find the issue. 

One suspect is `getXlatorsDir()` returning wrong xlator directory. https://github.com/gluster/glusterd2/blob/master/glusterd2/xlator/load.go#L153-L167

Issue #912, opened 2018-06-25
gluster-block integration with GD2


atinmu:
With GD2 we need to come up with an integration method on how gluster-block can be setup.

aravindavk:
I will take this up for "GCS-Sprint2"

atinmu:
This doesn't look like a candidate for GCS/0.2. @amarts @aravindavk Please move it to GCS/0.3 ?

obnoxxx:
Related / broader discussion:
* https://github.com/gluster/gcs/issues/40
* https://github.com/gluster/gluster-csi-driver/issues/10

atinmu:
gluster-block implementation (RWO) will be based on the loopback devices which is currently work in progress. This issue was to track the iSCSI/TCMU based gluster-block integration which is not a priority for GCS.

PR #1319, opened 2018-11-08
WIP: Managing Gluster block volumes

aravindavk:
Updates: #912

### DIFF
--- a/doc/design/block-volumes-integration.md
+++ b/doc/design/block-volumes-integration.md
@@ -1,0 +1,122 @@
+# Managing Gluster Block Volumes
+
+Following new APIs will be introduced to support Gluster block Volumes.
+
+## New Config and Environment variable options and its default values
+
+Environment variables will get more priority than defined in config
+file.
+
+```
+block-hosting-volume-size          = 500G
+block-hosting-volume-type          = Replica
+block-hosting-volume-replica-count = 3
+auto-create-block-hosting-volumes  = false
+```
+
+Note: Environment variables will starts with `GD2_`, for example
+`GD2_BLOCK_HOSTING_VOLUME_SIZE`
+
+## Create a Block Volume
+
+```
+POST /v1/blockvolumes
+```
+
+gluster-block project provides CLI command to create, delete and list
+the block volumes. Glusterd2 will provide wrapper around these CLI
+commands to provide block volumes.
+
+Request format:
+
+
+```golang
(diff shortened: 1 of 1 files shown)


atinmu:
Great start! While this document captures on the internals of the APIs and how the business logic will be implemented based on different conditions on available block volumes, size et all, it'd be awesome if we can have a design flow (mainly a diagram) which can illustrate the workflow above and underneath the GD2 layers to have a better grasp on the design for the beginners.

Thoughts?

amarts:
> Thoughts?

I guess, in none of these APIs, there is any role of GD2 at all. This provides 1-1 mapping between gluster-block-cli to a gd2 plugin API. (As already mentioned in doc).

Only change is, if there is no block-hosting-volume present, then it would create one. (And also see if the volume is full, create one more).

atinmu:
> I guess, in none of these APIs, there is any role of GD2 at all. This provides 1-1 mapping between gluster-block-cli to a gd2 plugin API. (As already mentioned in doc).

Probably it's unfair to say that GD2 plugin API is not GD2's responsibility :-)

aravindavk:
> Probably it's unfair to say that GD2 plugin API is not GD2's responsibility :-)

As @amarts mentioned, this feature can be developed without changing glusterd2. Except the block hosting volume management all other steps can be done with Gluster mount(Can be outside the cluster).

Also, if we replace block cli commands with `mount`, `mkdir` and xfs `quota`, then it will become flex volume plugin. (If we replace block cli command with mount, loop device create then it will become loop dev plugin)

amarts:
> Probably it's unfair to say that GD2 plugin API is not GD2's responsibility :-)

Yes, and hence the discussion is going on @ gluster/gcs#40 

Ideally, we would have done this as part of CSI driver itself, but as requested, if we keep the API similar to heketi in gd2, then we can actually write a single CSI driver which would work for both stack. Which means, it is much easier to validate if g-b works as same as current stack too.

This document, and work would be phase#1 of the work, and once things stabilize, and more well integrated, we may still have to move it out to CSI driver.

amarts:
Update for those who are reviewing, at the end of discussion in a meeting today, thought of doing all these as a separate project (or as a sub-process) of gluster-block came up. Which means that gluster-block plugin in plan itself can work as a solution to gluster/gluster-block#43 and also would solve the issue of having 'gluster-block CLI' as a separate process. 

Happy to hear feedback.

PR #1308, opened 2018-10-31
Include gluster-prometheus-exporter rpm in glusterd2 container

aravindavk:
- Depends on rpm spec file PR
  https://github.com/gluster/gluster-prometheus/pull/26
- Nightly rpm building is not yet setup
- Depends of environment variable related PR
  https://github.com/gluster/gluster-prometheus/pull/66

Signed-off-by: Aravinda VK <avishwan@redhat.com>

PR #1323, opened 2018-11-12
nightly-container: Use 'command' instead of 'service'

kshlm:
I had asked for the usage of the 'service' module to enable services in
the Ansible provisioner. Forgetting that 'service' module does not work
with Ansible-Buildah as systemd isn't running in the build container
(there is even a comment about this).

This commit changes back to using the 'command' module.


kshlm:
@aravindavk I made a mistake while reviewing #1308. This fixes the problem I caused.

Issue #1322, opened 2018-11-12
glusterd2 is failing on all nodes when force start the volume.


Shrivaibavi:
### Description
Created a replica 3 volume and started it. Added-bricks to the volume and everything was fine. Just tried to froce start the volume, the glusterd2 went into failed state on all the 3 nodes. When i tried to reproduce i saw glusterd2 failing on 2 nodes and etcd is running fine.

### Observed behavior

```
[root@dhcp35-30 ~]# glustercli volume create volrep2 --replica 3 10.70.35.30:/bricks/brick0/volrep2 10.70.35.240:/bricks/brick0/volrep2 10.70.35.106:/bricks/brick0/volrep2 --create-brick-dir
volrep2 Volume created successfully
Volume ID:  8cd54f95-8835-421a-9607-526b919301f2
[root@dhcp35-30 ~]# 
[root@dhcp35-30 ~]# 
[root@dhcp35-30 ~]# 
[root@dhcp35-30 ~]# glustercli volume start volrep2
Volume volrep2 started successfully
[root@dhcp35-30 ~]# 
[root@dhcp35-30 ~]# 
[root@dhcp35-30 ~]# 
[root@dhcp35-30 ~]# glustercli volume info volrep2

Volume Name: volrep2
Type: Replicate
Volume ID: 8cd54f95-8835-421a-9607-526b919301f2
State: Started
Transport-type: tcp
Options:
Number of Bricks: 3
Brick1: 10.70.35.30:/bricks/brick0/volrep2
Brick2: 10.70.35.240:/bricks/brick0/volrep2
Brick3: 10.70.35.106:/bricks/brick0/volrep2
[root@dhcp35-30 ~]# 
[root@dhcp35-30 ~]# 
[root@dhcp35-30 ~]# 
[root@dhcp35-30 ~]# glustercli volume status volrep2
Volume : volrep2
+--------------------------------------+--------------+------------------------+--------+-------+-------+
|               BRICK ID               |     HOST     |          PATH          | ONLINE | PORT  |  PID  |
+--------------------------------------+--------------+------------------------+--------+-------+-------+
| 8fe6858f-ce09-4aad-b0f9-e880e7e8ab7b | 10.70.35.30  | /bricks/brick0/volrep2 | true   | 46247 |  6844 |
| 6794973c-6b74-4dc6-8cd7-4671337afc2f | 10.70.35.240 | /bricks/brick0/volrep2 | true   | 42387 |  7072 |
[...]

PR #1337, opened 2018-11-23
Continue mounting all localBricks by logging and not returning on error

vpandey-RH:
Updates [#1322](https://github.com/gluster/glusterd2/issues/1322)

ISSUE -
MountLocalBricks() and MountVolumeBricks() are called during GD2 startup.
Both these methods are responsible for mounting bricks for all the volumes in the store.
MountLocalBricks() exits as soon as there is some issue with mounting any volume's localBricks because of
which other volumes' localBricks(which are yet to be processed) are skipped for mounting.
MountVolumeBricks() exits as soon as any one of the brick of that volume is unable to be mounted because of which
other localBricks of that volume(which are yet to be processed), are skipped for mounting.

FIX-
Modify both the methods to not exit after an error but log the error and continue the mounting process for othe volumes and localBricks


glusterd2/volume/fs_utils.go
```diff
@@ -293,7 +294,8 @@ func MountVolumeBricks(volinfo *Volinfo) error {
 
 	for _, b := range volinfo.GetLocalBricks() {
 		if err := MountBrickDirectory(volinfo, &b, mtab); err != nil {
```
Madhu-1: `MountVolumeBricks` will also get called while taking the snapshot clone. 
continue during  error scenarios may affect the functionality, @rafikc30 PTAL

vpandey-RH:
@rafikc30  Need your opinion on this.

glusterd2/volume/fs_utils.go
rafikc30: As @Madhu-1 suggested during a snapshot clone, if the mount failed, then there is no point in continuing the operation. So I would suggest having a rapper function like `RestartMountVolumeBricks` or have a flag passed to the function like
```
func MountVolumeBricks (volinfo *Volinfo, skipError bool) {
        mtab, err := GetMounts()
        if err != nil {
                return err 
        }

        for _, b := range volinfo.GetLocalBricks() {
                if err := MountBrickDirectory(volinfo, &b, mtab); err != nil && !skipError {
                        return err 
                }
        }
   return nil
}
```

aravindavk: @rafikc30 if one brick is corrupt why blocking the mount of other bricks(which may belong to other subvolume)?

vpandey-RH:
@Madhu-1  @rafikc30  PTAL.

aravindavk:
retest this please

Issue #1161, opened 2018-08-13
Arbiter volume configuration changed after add-brick


ksandha:
1) Create a arbiter volume
```
[root@dhcp35-88 ~]# glustercli volume create newvol2 replica 2 arbiter 1 10.70.35.88:/bricks/brick1/newvol 10.70.35.181:/bricks/brick1/newvol 10.70.35.26:/bricks/brick1/newvol -v
newvol2 Volume created successfully
Volume ID:  69a0d70f-f1b4-4654-a4d2-6c176db64995

```
2)  check volume info ; arbiter volume created.
```
[root@dhcp35-88 ~]# glustercli volume info newvol2

Volume Name: newvol2
Type: Replicate
Volume ID: 69a0d70f-f1b4-4654-a4d2-6c176db64995
State: Created
Transport-type: tcp
Options:
    replicate.arbiter-count: 1
Number of Bricks: 3
Brick1: 10.70.35.88:/bricks/brick1/newvol
Brick2: dhcp35-181.lab.eng.blr.redhat.com:/bricks/brick1/newvol
Brick3: dhcp35-26.lab.eng.blr.redhat.com:/bricks/brick1/newvol (arbiter)

```
3) add bricks to the arbiter volume.
```
[root@dhcp35-88 ~]# glustercli volume add-brick newvol2 10.70.35.88:/bricks/brick2/newvol2 10.70.35.181:/bricks/brick2/newvol2 10.70.35.26:/bricks/brick2/newvol2 -v
newvol2 Volume expanded successfully

```
4) Check info of the volume.
```
[root@dhcp35-88 ~]# glustercli volume info newvol2

Volume Name: newvol2
Type: Replicate
Volume ID: 69a0d70f-f1b4-4654-a4d2-6c176db64995
State: Created
Transport-type: tcp
Options:
    replicate.arbiter-count: 1
Number of Bricks: 2 x (1 + 1) = 5
Brick1: 10.70.35.88:/bricks/brick1/newvol
Brick2: dhcp35-181.lab.eng.blr.redhat.com:/bricks/brick1/newvol
Brick3: dhcp35-26.lab.eng.blr.redhat.com:/bricks/brick1/newvol (arbiter)
Brick4: 10.70.35.88:/bricks/brick2/newvol2
Brick5: dhcp35-181.lab.eng.blr.redhat.com:/bricks/brick2/newvol2
[root@dhcp35-88 ~]# 

```

[...]

PR #1351, opened 2018-11-28
Fix expanding arbiter volume

aravindavk:
Arbiter count was not considered when a Volume is expanded.

Fixes: #1161

### DIFF
--- a/glusterd2/commands/volumes/volume-expand-txn.go
+++ b/glusterd2/commands/volumes/volume-expand-txn.go
@@ -41,7 +41,7 @@ func expandValidatePrepare(c transaction.TxnCtx) error {
 	if req.ReplicaCount == 0 {
 		newReplicaCount = volinfo.Subvols[0].ReplicaCount
 	}
-	if (len(req.Bricks)+len(volinfo.GetBricks()))%newReplicaCount != 0 {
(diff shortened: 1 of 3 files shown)


aravindavk:
> @aravindavk Are we looking to add e2e tests to verify this ? Also, I noticed the ''Use:'' section in case of volume expand CLI doesn't contain a few flags usage like --replica, which can be a bit confusing for users, can we modify it in the same PR ?

I will add e2e tests and update the CLI help

aravindavk:
Added e2e test and changed help message. @vpandey-RH @Madhu-1 please review

glusterd2/commands/volumes/volume-expand-txn.go
```diff
@@ -207,13 +207,29 @@ func updateVolinfoOnExpand(c transaction.TxnCtx) error {
 	} else {
 		// Create new Sub volumes with given bricks
 		subvolIdx := len(volinfo.Subvols)
-		for i := 0; i < len(newBricks)/newReplicaCount; i++ {
-			idx := i * newReplicaCount
+		bricksCount := newReplicaCount + volinfo.Subvols[0].ArbiterCount
+		for i := 0; i < len(newBricks)/bricksCount; i++ {
```
Madhu-1: can we have `len(newBricks)` calculation outside to avoid computing of this value each time.

aravindavk: done

PR #1357, opened 2018-11-30
gd2 plugin: added a plugin for block volume management

oshankkumar:
- added APIs for creation,deleting and listing block volumes.
 - added pluggable interface for block volume providers.

Refer Design Doc: #1319


Madhu-1:
add to whitelist

Madhu-1:
CI failure
```
20:58:20 Building glusterd2 v6.0-dev.54.gita525bc5
20:58:21 plugins/blockvolume/blockprovider/gluster-block/glusterblock.go:10:2: cannot find package "github.com/gluster/gluster-block-restapi/client" in any of:
20:58:21 	/root/go/src/github.com/gluster/glusterd2/vendor/github.com/gluster/gluster-block-restapi/client (vendor tree)
20:58:21 	/usr/local/go/src/github.com/gluster/gluster-block-restapi/client (from $GOROOT)
20:58:21 	/root/go/src/github.com/gluster/gluster-block-restapi/client (from $GOPATH)
20:58:21 plugins/blockvolume/blockprovider/gluster-block/glusterblock.go:11:2: cannot find package "github.com/gluster/gluster-block-restapi/pkg/api" in any of:
20:58:21 	/root/go/src/github.com/gluster/glusterd2/vendor/github.com/gluster/gluster-block-restapi/pkg/api (vendor tree)
20:58:21 	/usr/local/go/src/github.com/gluster/gluster-block-restapi/pkg/api (from $GOROOT)
20:58:21 	/root/go/src/github.com/gluster/gluster-block-restapi/pkg/api (from $GOPATH)
```

nigelbabu:
retest this please

plugins/blockvolume/blockprovider/gluster-block/glusterblock.go
```diff
@@ -0,0 +1,314 @@
...
+		HaCount:            blockVolOpts.Ha,
+		AuthEnabled:        blockVolOpts.Auth,
+		FullPrealloc:       blockVolOpts.FullPrealloc,
+		Size:               uint64(size),
+		Storage:            blockVolOpts.Storage,
+		RingBufferSizeInMB: blockVolOpts.RingBufferSizeInMB,
+		Hosts:              hosts,
+	}
+
+	resp, err := g.client.CreateBlockVolume(volInfo.Name, name, req)
```
amarts: We need to make sure we have the BlockHostingVolume is up and running. Because, if it is not present, then this will fail for sure. If it is already taken care by the events above, then all good.

aravindavk:
I don't see `Gopkg.toml` changes which mentions the new dependency

oshankkumar:
> I don't see `Gopkg.toml` changes which mentions the new dependency

I am getting error like this while updating the dependencies using `dep ensure`

```
Solving failure: No versions of github.com/gluster/gluster-block-restapi met constraints:
	v0.2: Could not introduce github.com/gluster/gluster-block-restapi@v0.2, as it has a dependency on github.com/sirupsen/logrus with constraint ^1.2.0, which has no overlap with existing constraint ~1.0.3 from (root)
```

@aravindavk @Madhu-1 can you please suggest how to resolve this

Madhu-1:
@oshankkumar update `github.com/sirupsen/logrus` version  to "~1.2.0" in toml file to fix this issue

plugins/blockvolume/blockprovider/gluster-block/glusterblock.go
aravindavk: Line 149 is checking that Hosting Volume is in Started State

plugins/blockvolume/blockprovider/gluster-block/glusterblock.go
```diff
@@ -0,0 +1,325 @@
...
+// GetBlockVolume gives info about a gluster block volume
+func (g *GlusterBlock) GetBlockVolume(name string) (blockprovider.BlockVolume, error) {
+	var (
+		blockVolume           blockprovider.BlockVolume
+		availableBlockVolumes = g.BlockVolumes()
+	)
+
+	for _, blockVol := range availableBlockVolumes {
+		if blockVol.Name() == name {
+			blockVolume = blockVol
```
aravindavk: break?

plugins/blockvolume/blockprovider/gluster-block/volume.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	vInfo.Metadata["_block-hosting-volume-auto-created"] = "yes"
+	return vInfo, nil
+}
+
+// StartBlockHostingVolume starts a gluster volume
+func StartBlockHostingVolume(name string) error {
```
aravindavk: Is this different from regular Volume start. If not can we reuse the code

plugins/blockvolume/blockprovider/gluster-block/volume.go
```diff
@@ -0,0 +1,163 @@
...
+	status, err := volumecommands.CreateVolume(context.Background(), *req)
+	if err != nil || status != http.StatusCreated {
+		return nil, err
+	}
+	vInfo, err := volume.GetVolume(req.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	vInfo.Metadata["_block-hosting-volume-auto-created"] = "yes"
```
aravindavk: Using `_` prefix will make it impossible to manually create block hosting volume and use that for block file creation.(Since Metadata keys starts with `_` are restricted for API users). We can skip using prefix or we can expose BlockHosting=Yes/No flag in Volume Edit API

glusterd2/commands/volumes/volume-create.go
```diff
@@ -100,86 +101,93 @@ func registerVolCreateStepFuncs() {
...
+
 	}
 
 	if err := validateOptions(req.Options, req.VolOptionFlags); err != nil {
-		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, err)
-		return
-	}
+		return http.StatusBadRequest, err
 
-	// Include default Volume Options profile
```
aravindavk: this new change is lost in Rebase?

glusterd2/commands/volumes/volume-create.go
```diff
@@ -100,86 +101,93 @@ func registerVolCreateStepFuncs() {
...
 	}
 
-	if err := validateVolCreateReq(&req); err != nil {
-		restutils.SendHTTPError(ctx, w, http.StatusBadRequest, err)
+	if status, err := CreateVolume(ctx, req); err != nil {
+		restutils.SendHTTPError(ctx, w, status, err)
 		return
 	}
 
-	if containsReservedGroupProfile(req.Options) {
```
Madhu-1: this check also lost

plugins/blockvolume/api/types.go
```diff
@@ -0,0 +1,38 @@
+package api
+
+// BlockVolumeInfo represents block volume info
+type BlockVolumeInfo struct {
+	// HostingVolume name is optional
+	HostingVolume string `json:"hosting_volume"`
```
Madhu-1: don't use underscore in json name

Madhu-1: can this be omitempty?

Madhu-1: am i missing anything, dont we  need to consider size for blockhosting volume.

plugins/blockvolume/api/types.go
```diff
@@ -0,0 +1,38 @@
...
+// BlockVolumeCreateRequest represents req Body for Block vol create req
+type BlockVolumeCreateRequest struct {
+	*BlockVolumeInfo
+	Clusters []string `json:"clusters,omitempty"`
+	Auth     bool     `json:"auth,omitempty"`
+}
+
+// BlockVolumeCreateResp represents resp body for a Block Vol Create req
+type BlockVolumeCreateResp struct {
+	*BlockVolumeInfo
```
Madhu-1: as you are using `BlockVolumeInfo` `size` cannot be omitempty for this volume response

plugins/blockvolume/api/types.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// BlockVolumeListResp represents resp body for a Block List req
+type BlockVolumeListResp []BlockVolumeInfo
+
+// BlockVolumeGetResp represents a resp body for Block Vol Get req
+type BlockVolumeGetResp struct {
+	*BlockVolumeInfo
+	Hosts    []string `json:"hosts"`
+	Password string   `json:"password,omitempty"`
```
Madhu-1: dont we need username here?

plugins/blockvolume/blockprovider/blockvolume_provider.go
```diff
@@ -0,0 +1,59 @@
...
+	HostAddresses() []string
+	IQN() string
+	Username() string
+	Password() string
+	Size() uint64
+}
+
+// RegisterBlockProvider will register a block provider
+func RegisterBlockProvider(name string, f ProviderFunc) {
+	providersMutex.Lock()
```
Madhu-1: can you explain whats the advantage of taking lock here and in below functions.

plugins/blockvolume/api/types.go
oshankkumar: we don't get username details  from `gluster-block info` command

plugins/blockvolume/blockprovider/blockvolume_provider.go
oshankkumar: just to make sure providerFactory map is not accessed parallelly by some other goroutine

plugins/blockvolume/blockprovider/gluster-block/volume.go
oshankkumar: done

plugins/blockvolume/api/types.go
oshankkumar: Since we are getting only list of blockvolume name from `gluster-block list` command.
So for a blockvolume list response, we are only showing blockvolume name and hosting volume name, so if I remove `omitempty` then size will be displayed as 0 in response.

oshankkumar: for a BlockVolumeList response `size` is not a mandatory field

plugins/blockvolume/blockprovider/gluster-block/glusterblock.go
oshankkumar: added

poornimag:
As discussed, could you please move the block host volume management to block provider instead of gluster-block, so that the block host volume management code can be reused by all providers?

oshankkumar:
> As discussed, could you please move the block host volume management to block provider instead of gluster-block, so that the block host volume management code can be reused by all providers?

moved the common code to a utils package

aravindavk:
Block provider details can be handled in URL to support different providers. `POST /v1/subvolumes/{provider}`. (**Note:** Name "subvolumes" here may add confusion wrt subvol in gluster. Any suggestions for the name?)

Create

```
POST /subvolumes/gluster-block         To provide block volumes using gluster-block
POST /subvolumes/loop                      To provide block volumes using loop devices
POST /subvolumes/dir                         To provide volumes using subdir approach
```

Similarly for other URLS

Same ReST handler receives all the requests and host volume management part will remain same. Instead of calling `g.client.CreateBlockVolume(volInfo.Name, name, req)` in Line 164(https://github.com/gluster/glusterd2/pull/1357/files#diff-3a743ffc1638946e95edba8bc6e26c50R164) it will call `provider.CreateBlockVolume(args...)`

In handler,

```
providerName := mux.Vars(r)["provider"]
provider := getProvider(providerName)  // This gets the interface implementation of respective provider
```

@oshankkumar @poornimag @Madhu-1 @amarts thoughts?

Madhu-1:
@aravindavk @amarts how about keeping the URL as constant something like `v1/blockvolume` and pass the provider as part of request body?

aravindavk:
> @aravindavk @amarts how about keeping the URL as constant something like v1/blockvolume and pass the provider as part of request body?

- It will not look good for other URLs like `GET /blockvolumes?provider=<provider>`
- Wrong error messages if provider is not specified. For example `POST /blockvolumes` for create with loop as parameter in body. But `GET /blockvolumes` may not show if that is not the default(It will show not found but not as invalid provider)

amarts:
Thanks for the reviews Michael.

> How well is the code protected against concurrent block-volume create operations if at all? (Reading I think it isn't but maybe it is hidden.) What I'm referring to here is: If there are multiple concurrent block volume create requests when there is no block-hosting volume available, each of the requests would create a block-hosting volume if done separately, but when done concurrently we should coordinate the creation of block-hosting volumes and normally only create ONE block hosting volume for all those block-volume create requests.

Yes, this is a must for the project, to call the plugin ready. And is the main reason to move the logic to glusterd2 from that of CSI. While this functionality is not complete, the PR itself will allow us to get started from other front, and hence the push for the 'functionality'.

plugins/blockvolume/blockprovider/gluster-block/glusterblock.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	resp, err := g.client.CreateBlockVolume(hostVolume, name, req)
+	if err != nil {
+		return nil, err
+	}
+
+	volInfo.Metadata["block-hosting-available-size"] = fmt.Sprintf("%d", int64(availableSizeInBytes)-size)
+
+	if err := volume.AddOrUpdateVolume(volInfo); err != nil {
```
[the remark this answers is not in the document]
oshankkumar: added cluster wide lock

plugins/blockvolume/blockprovider/gluster-block/glusterblock.go
```diff
@@ -0,0 +1,259 @@
...
+	req := &api.BlockVolumeDeleteReq{
+		UnlinkStorage: blockVolOpts.UnlinkStorage,
+		Force:         blockVolOpts.ForceDelete,
+	}
+
+	if err := g.client.DeleteBlockVolume(hostVol, name, req); err != nil {
+		return err
+	}
+
+	if err := utils.ResizeBlockHostingVolume(hostVol, blockInfo.Size); err != nil {
```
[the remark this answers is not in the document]
oshankkumar: done

oshankkumar:
@obnoxxx @amarts by taking a cluster-wide lock (etcd lock) will prevent in creating multiple block hosting volume if concurrent request arrives.

glusterd2/commands/volumes/volume-create.go
oshankkumar: fixed

plugins/blockvolume/hostvol_manager.go
```diff
@@ -0,0 +1,163 @@
...
+// If volume name is not provided then it will create a gluster volume with default size for hosting gluster block.
+func (g *glusterVolManager) GetOrCreateHostingVolume(name string, minSizeLimit int64) (*volume.Volinfo, error) {
+	var (
+		volInfo      *volume.Volinfo
+		volCreateReq = g.hostVolOpts.PrepareVolumeCreateReq()
+		ctx          = gdctx.WithReqLogger(context.Background(), log.StandardLogger())
+		clusterLocks = transaction.Locks{}
+	)
+
+	if err := clusterLocks.Lock(path.Join(globalLockID, name)); err != nil {
```
aravindavk: Cluster wide lock with volume name will not help. If two concurrent requests gets different volume name then that will succeed and we end up creating two host volumes. Since CreateVolume anyways takes lock on Volume name, can we only take cluster wide lock with globalLockID only?

oshankkumar: I have a doubt here, if hostvolume name is present in request body, it means that volume already exists and we should not create it.
As mentioned in design doc we should only create hostvolume if hostvolume name  is not provided in req body.

oshankkumar: yes, I am not using the generated name, in taking lock.

pkg/size/size.go
```diff
@@ -0,0 +1,187 @@
+package size
```
Madhu-1: why do we require this package?
support only bytes as input, then we don't need this package.

aravindavk: This is a nice package, but I agree with @Madhu-1 API only need to support bytes

plugins/blockvolume/blockprovider/blockvolume_provider.go
```diff
@@ -0,0 +1,61 @@
...
+	Size() uint64
+	HaCount() int
+}
+
+// RegisterBlockProvider will register a block provider
+func RegisterBlockProvider(name string, f ProviderFunc) {
+	providersMutex.Lock()
+	defer providersMutex.Unlock()
+	if _, found := providerFactory[name]; found {
+		log.WithField("name", name).Error("block provider was registered twice")
```
Madhu-1: add return if name found

plugins/blockvolume/blockprovider/gluster-block/config.go
```diff
@@ -0,0 +1,23 @@
...
+	HostAddress string
+	User        string
+	Secret      string
+	CaCertFile  string
+	Insecure    bool
+}
+
+// ApplyFromConfig sets the ClientConfig options from various config sources
+func (c *ClientConfig) ApplyFromConfig(conf *viper.Viper) {
+	c.CaCertFile = conf.GetString("gluster-block-cacert")
```
Madhu-1: we need a configuration as listed below

* read from the parameter passed why starting glusterd2
* read from the glusterd2 toml file
* read from ENV variable

plugins/blockvolume/blockprovider/gluster-block/glusterblock.go
```diff
@@ -0,0 +1,272 @@
...
+	if err != nil {
+		return nil, err
+	}
+	gb.client = gbClient
+
+	return gb, nil
+}
+
+// CreateBlockVolume will create a gluster block volume with given name and size having `hostVolume` as hosting volume
+func (g *GlusterBlock) CreateBlockVolume(name string, size int64, hosts []string, hostVolume string, options ...blockprovider.BlockVolOption) (blockprovider.BlockVolume, error) {
```
Madhu-1: if you have more than 5 input parameters, why can't we have a structure pointer as an input?

plugins/blockvolume/blockprovider/gluster-block/glusterblock.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+
+	defer clusterLocks.UnLock(context.Background())
+
+	availableSizeInBytes, err := strconv.Atoi(volInfo.Metadata["block-hosting-available-size"])
+	if err != nil {
+		return nil, err
+	}
+
+	if int64(availableSizeInBytes) < size {
```
Madhu-1: typecast to uint64, and use size as uint64

plugins/blockvolume/blockprovider/gluster-block/glusterblock.go
```diff
@@ -0,0 +1,272 @@
...
+
+	for _, blockVol := range blockVols {
+		if blockVol.Name() == name {
+			hostVol = blockVol.HostVolume()
+			break
+		}
+	}
+
+	if hostVol == "" {
+		return errors.New("block volume not found")
```
Madhu-1: host volume of block volume not found?

plugins/blockvolume/blockprovider/gluster-block/glusterblock.go
```diff
@@ -0,0 +1,272 @@
...
+	if err != nil {
+		return err
+	}
+
+	req := &api.BlockVolumeDeleteReq{
+		UnlinkStorage: blockVolOpts.UnlinkStorage,
+		Force:         blockVolOpts.ForceDelete,
+	}
+
+	if err := g.client.DeleteBlockVolume(hostVol, name, req); err != nil {
```
Madhu-1: question: during the 1st REST call block volume deletion was successful and updating block hosting failed. please add  a TODO to consider this scenario

Madhu-1: same for create also.

plugins/blockvolume/blockprovider/gluster-block/glusterblock.go
```diff
@@ -0,0 +1,272 @@
...
+		return err
+	}
+
+	if err := utils.ResizeBlockHostingVolume(hostVol, blockInfo.Size); err != nil {
+		log.WithError(err).Error("error in resizing the block hosting volume")
```
Madhu-1: missing return err

plugins/blockvolume/blockprovider/gluster-block/glusterblock.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+
+	return nil
```
Madhu-1: make it as a return err

pkg/size/size.go
oshankkumar: @aravindavk @Madhu-1 form gluster-block info we get the response in string format, we need to parse the string representation, so I thought a separate package will be helpful in future also.

plugins/blockvolume/blockprovider/gluster-block/glusterblock.go
oshankkumar: maybe this check is little misleading , if hostVol is empty, it means we don't have details of request block.

plugins/blockvolume/blockprovider/gluster-block/glusterblock.go
oshankkumar: all error scenario has been covered, except updating the volume details. that's why I returned nil.

oshankkumar: > make it as a return err

done

oshankkumar:
retest this please

atinmu:
@aravindavk @Madhu-1 can you folks take a look at the latest version? I believe all comments are being incorporated now?

oshankkumar:
> @aravindavk @Madhu-1 can you folks take a look at the latest version? I believe all comments are being incorporated now?

all review comments has been addressed.

amarts:
what more is pending on this ?

glusterd2.toml.example
```diff
@@ -3,3 +3,15 @@ peeraddress = ":24008"
 clientaddress = ":24007"
 #restauth enables/disables REST authentication in glusterd2
 #restauth = true
+
+#[gluster-block-hosting-vol-options]
+block-hosting-volume-size = 20971520
+auto-create-block-hosting-volumes = true
+block-hosting-volume-replica-count = 1
```
aravindavk: Change this to 3

glusterd2.toml.example
```diff
@@ -3,3 +3,15 @@ peeraddress = ":24008"
...
+#block-hosting-volume-type = "Distribute"
```
aravindavk: Change this to Replicate

Madhu-1:
@oshankkumar please resolve merge conflicts

glusterd2.toml.example
```diff
@@ -3,3 +3,15 @@ peeraddress = ":24008"
 clientaddress = ":24007"
 #restauth enables/disables REST authentication in glusterd2
 #restauth = true
+
+#[gluster-block-hosting-vol-options]
+block-hosting-volume-size = 20971520
```
Madhu-1: add a comment what is this size in Gib for more readability

glusterd2.toml.example
```diff
@@ -3,3 +3,15 @@ peeraddress = ":24008"
...
+auto-create-block-hosting-volumes = true
+block-hosting-volume-replica-count = 1
+#block-hosting-volume-type = "Distribute"
+
+#[gluster-block-client-config]
+gluster-block-hostaddr = "192.168.122.16:8081"
+#gluster-block-cacert = "/path/to/ca.crt"
+#gluster-block-user = "username_for_rest_authentication"
+#gluster-block-secret = "secret_for_rest_authentication"
```
Madhu-1: add a new line

Madhu-1:
@oshankkumar regenerate `endpoints.md` file

plugins/blockvolume/blockprovider/gluster-block/glusterblock.go
```diff
@@ -0,0 +1,274 @@
...
+		Hosts:              hosts,
+	}
+
+	if err := clusterLocks.Lock(hostVolume); err != nil {
+		log.WithError(err).Error("error in acquiring cluster lock")
+		return nil, err
+	}
+	defer clusterLocks.UnLock(context.Background())
+
+	volInfo, err := volume.GetVolume(hostVolume)
```
Madhu-1: can we move this logic to some other function, so this can be reused for checking block hosting volume?

oshankkumar: done

Madhu-1:
@aravindavk PTAL.

PR #1363, opened 2018-12-05
Generate Client Volfiles on demand

aravindavk:
Currently client Volfiles are stored in etcd, but this increases the
consumption of etcd storage space. With this PR, client volfiles are
generated on the fly when mount request comes.


glusterd2/commands/volumes/commands.go
```diff
@@ -136,18 +136,6 @@ func (c *Command) Routes() route.Routes {
 			Version:     1,
 			RequestType: utils.GetTypeString((*api.VolStatedumpReq)(nil)),
 			HandlerFunc: volumeStatedumpHandler},
-		route.Route{
```
Madhu-1: if I want to get the volfiles for some debugging purpose, how can I do it?

glusterd2/servers/sunrpc/handshake_prog.go
```diff
@@ -140,41 +140,66 @@ func (p *GfHandshake) ServerGetspec(args *GfGetspecReq, reply *GfGetspecRsp) err
...
+			volinfo, err = volume.GetVolume(volfileID)
+			if err != nil {
+				log.WithError(err).WithField(
+					"volfile", volfileID,
+				).Error("failed to get volume info")
+				goto Out
+			}
+		}
+
+		tmpl, err := volgen.GetTemplateFromVolinfo(volinfo, "client")
```
Madhu-1: is there any performance import if we generate volfiles on demand?

glusterd2/commands/volumes/commands.go
aravindavk: from mount log file

Madhu-1: question:
is it possible to generate a volfiles and return in response for this HTTP request?
will it be helpful for us later?

aravindavk:
Not seeing much difference in time for mount:

Fetch Volfile from etcd:

```
13:38:33     --- PASS: TestSnapshot/Mount (0.27s)
13:38:33     --- PASS: TestSnapshot/MountRestoredVolume (0.20s)
13:39:11     --- PASS: TestVolume/Mount (0.28s)
```

Generate client volfile on demand

```
12:34:28     --- PASS: TestSnapshot/Mount (0.26s)
12:34:28     --- PASS: TestSnapshot/MountRestoredVolume (0.20s)
12:35:05     --- PASS: TestVolume/Mount (0.28s)
```

glusterd2/servers/sunrpc/handshake_prog.go
aravindavk: https://github.com/gluster/glusterd2/pull/1363#issuecomment-444382322

glusterd2/commands/volumes/commands.go
aravindavk: Yes for client volfiles, need to get Volume info and generate the volfile on request. We can add that later if required. Since it generates fresh volfile on every request this may not be useful for debugging. Mount log file will be source of truth for actual volfile content used by glusterfs process.

Alternatively, We can dump the generated volfile to disk for debugging. (Note: Volfiles will be on disk only if a client(mount) connected to that glusterd2)

PR #1371, opened 2018-12-06
Fix #1321: remove session references

### LINKED ISSUE #1321: Glusterd2 consuming huge amount of memory when a brick is down
### Observed behavior

glusterd2 process consumes 82% of memory.

```
 PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND                                                                                                  
24616 root      20   0 15.179g 3.054g   4552 S   1.7 82.5   3:08.14 glusterd2                                                                                                
  783 ovirtag+  20   0  547296   7800   3496 S   0.7  0.2  43:59.27 python                                                                                                   
  436 root      20   0       0      0      0 S   0.3  0.0   2:17.23 xfsaild/dm-0                                                                                             
24870 root      20   0 1185580   7208    168 S   0.3  0.2   0:04.67 glusterfsd                                                                                               
25003 root      20   0  626548   8060   1008 S   0.3  0.2   0:37.50 glusterfs  

```

### Expected/desired behavior

Memory should not increase to 82%

1) Start glusterd2
2) Create a replica volume
3) Bring down one brick
4) Observe the memory usage

- Operating system used: fedora
- Glusterd2 compiled from sources, as a package (rpm/deb), or container: source install
- Using External ETCD: (yes/no, if yes ETCD version):

bserdar:
acceptLoop keeps a list of all sessions, and each session keeps a reference to the memory block allocated as a read buffer. This reference is there even if the session closes, so as long as acceptLoop runs, memory keeps growing. This should fix it by doing two things: 1) When session closes, remove the reference to the buffer. But that will still continue to leak sessions, because nothing removes elements from the session list. So, 2) remove the logic that keep track of sessions, instead, run two goroutines for each session, one to keep the session and run the server, and the other to terminate the first one when needed.

I don't have the infra to verify this, so let me know how it goes.

### DIFF
--- a/glusterd2/servers/sunrpc/server.go
+++ b/glusterd2/servers/sunrpc/server.go
@@ -132,17 +132,10 @@ func (s *SunRPC) acceptLoop(stopCh chan struct{}, l net.Listener, wg *sync.WaitG
 		"transport": ltype})
 	logger.WithField("address", l.Addr().String()).Info("started server")
 
-	sessions := make([]rpc.ServerCodec, 50)
 	for {
 		select {
 		case <-stopCh:
 			logger.Debug("stopped accepting new connections")
-			logger.Debug("closing client connections")
-			for _, c := range sessions {
-				if c != nil {
-					c.Close()
-				}
-			}
 			return
 		default:
 		}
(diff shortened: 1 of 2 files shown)


rishubhjain:
@bserdar Does this PR helps reduce memory leak?

bserdar:
There are two leaks, the larger one is related to the read buffer in ServerCodec. It is not freed, because a reference to the buffer is kept in the servercodec, acceptLoop keeps a list of all servercodecs even if they're closed. so, the 1-liner in servercodec.go should remove the read buffer reference when the session is closed, and the two goroutines in the acceptLoop should eliminate the need to keep all the sessions.

rishubhjain:
Fixes:https://github.com/gluster/glusterd2/issues/1321 @harigowtham Please check

atinmu:
@prashanthpai if possible, please review!

glusterd2/servers/sunrpc/server.go
```diff
@@ -132,17 +132,10 @@ func (s *SunRPC) acceptLoop(stopCh chan struct{}, l net.Listener, wg *sync.WaitG
 		"transport": ltype})
 	logger.WithField("address", l.Addr().String()).Info("started server")
 
-	sessions := make([]rpc.ServerCodec, 50)
 	for {
 		select {
 		case <-stopCh:
```
harigowtham: doesn't this implementation make this code useless?

bserdar: The select on stopCh is still required to return from the acceptLoop. There is no other return from this function. If stopCh is closed, acceptLoop will terminate, along with all the sessions that are still active.

harigowtham: Got the return in the func confused with the return for this infinte loop.

atinmu:
Functionality wise changes look fine.

@aravindavk @madhu-1 please review?

aravindavk:
add to whitelist

aravindavk:
retest this please

aravindavk:
Thanks @bserdar for the PR. Two go routine approach is nice and helpful.

PR #1377, opened 2018-12-10
Fix go vet issues

Madhu-1:


pkg/logging/logging.go
```diff
@@ -87,7 +87,7 @@ func Init(logdir string, logFileName string, logLevel string, verboseLogEntry bo
 		logFile, err := openLogFile(logFilePath)
 		if err != nil {
 			setLogOutput(os.Stderr)
-			log.WithError(err).Debug("Failed to open log file %s", logFilePath)
+			log.WithError(err).Debug("Failed to open log file ", logFilePath)
```
rishubhjain: Will it print the file path?

Madhu-1: yes

Madhu-1: no use of using `%s` at the end  of the statement, if it was in the middle it would be  useful, 
because of this go vet was giving error.

PR #1379, opened 2018-12-10
Fix nightly-rpms

kshlm:


scripts/build.sh
```diff
@@ -32,7 +32,7 @@ GIT_SHA=${GIT_SHA:-$(git rev-parse --short HEAD || echo "undefined")}
 GIT_SHA_FULL=${GIT_SHA_FULL:-$(git rev-parse HEAD || echo "undefined")}
 
 LDFLAGS="-X ${REPO_PATH}/version.GlusterdVersion=${VERSION} -X ${REPO_PATH}/version.GitSHA=${GIT_SHA}"
-LDFLAGS+=" -B 0x${GIT_SHA_FULL}"
+LDFLAGS+=" -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \n')"
```
atinmu: Could you explain what this change is all about?

kshlm: The commit message explains this. Previously, we were using git hash as build-id which gave same build-id for both glusterd2 and glustercli binaries. A new change in rpmbuild started marking packages with binaries with same build-ids as failures. So this change.

atinmu:
We will have to wait for https://review.gluster.org/#/c/glusterfs/+/21843/ to get into glusterfs repo and a glusterfs build might be required to get the smoke passed.

atinmu:
retest this please

atinmu:
retest this please

scripts/build.sh
aravindavk: Question: Can we add prefix/suffix to differenciate the build ids for glustercli and glusterd2 instead of using random ID

atinmu:
@aravindavk Apologies that I missed out your comment before merging this change. If you’ve any objection to the current change a follow up PR can be sent.

Issue #999, opened 2018-07-12
Auto distribute count for Intelligent Volume Provisioning


aravindavk:
Current dynamic Volume create API accepts distribute count as input and defaults to 1 if no distribute count is specified. New feature is required to find distribute count automatically based on given Volume size and available size.

If Distribute count is not specified and disk space is not enough to create volume, try again with increased distribute count till all the bricks can be planned(till max distribute count)

aravindavk:
Issue #851 is split into multiple issues. This is one among them

aravindavk:
Alternate idea compared to Iterative method till max distribute count: New API tunable `max-brick-size`.

    if (requested-size > max-brick-size) 
    {
        distribution_count = int(requested-size / max-brick-size);
        if (requested-size % max-brick-size > 0)
        {
            distribution_count++;
        }
    }

JohnStrunk:
I like the notion of having `max-brick-size` as it makes efficient bin packing easier when item size is constrained.

PR #1386, opened 2018-12-12
Auto distribute count based on Max brick size

aravindavk:
While creating auto provisioned volume, support added to automatically
calculate the distribute count based on max brick size specified in
the request.

For example, below command creates 2x3(Distributed replicate) volume

```
glustercli volume create gv1 --replica 3 --size 1G \
        --max-brick-size 512M
```

Fixes: #999


glustercli/cmd/volume-create.go
```diff
@@ -102,10 +104,16 @@ func smartVolumeCreate(cmd *cobra.Command, args []string) {
 		failure("Invalid File Size specified", nil, 1)
 	}
 
+	maxBrickSize, err := sizeToBytes(flagCreateMaxBrickSize)
```
Madhu-1: if `flagCreateMaxBrickSize` has changed then do size calculation?

e2e/smartvol_ops_test.go
```diff
@@ -407,6 +407,54 @@ func testSmartVolumeDistributeDisperse(t *testing.T) {
 	checkZeroLvs(r)
 }
 
+func testSmartVolumeAutoDistributeReplicate(t *testing.T) {
```
Madhu-1: can you add some negative test cases?

aravindavk: what kind of negative tests I can add? negative max-brick-size or max-brick-size is greater than request size?

glustercli/cmd/volume-create.go
aravindavk: sizeToBytes returns zero when input is empty, no conversion/calculation done

e2e/smartvol_ops_test.go
Madhu-1: kind of invalid minimum and maximum values for  max-brick-size

aravindavk:
@Madhu-1 please review

aravindavk:
Found one more issue in Volume create flow, opened issue https://github.com/gluster/glusterd2/issues/1389

pkg/lvmutils/utils.go
```diff
@@ -345,3 +341,12 @@ func ExtendThinpool(expansionTpSizePerBrick uint64, vgName string, tpName string
 	err := utils.ExecuteCommandRun("lvextend", fmt.Sprintf("-L+%dB", expansionTpSizePerBrick), fmt.Sprintf("/dev/%s/%s", vgName, tpName))
 	return err
 }
+
+// NormalizeSize converts the value to multiples of 512
```
Madhu-1: can we make it unexported function?

glusterd2/bricksplanner/planner.go
```diff
@@ -95,6 +87,37 @@ func getBricksLayout(req *api.VolCreateReq) ([]api.SubvolReq, error) {
 		return nil, err
 	}
 
+	if req.MaxBrickSize > 0 && req.MaxBrickSize < minBrickSize {
+		return nil, errors.New("invalid max-brick-size, Minimum size required is " + strconv.Itoa(minBrickSize))
+	}
+
+	// If max Brick size is specified then decide distribute
+	// count and Volume Size based on Volume Type
+	if req.MaxBrickSize > 0 && req.Size > req.MaxBrickSize {
```
Madhu-1: what happens if `req.MaxBrickSize` if greater than `req.Size`?

aravindavk: It will be same as not passing that option

pkg/lvmutils/utils.go
aravindavk: this is used in `glusterd2/bricksplanner` as well

glusterd2/bricksplanner/planner.go
Madhu-1: as this is an input request we need to return an error to the user.

aravindavk: I don't think that is an error. It is satisfying the user's condition of max-brick-size.

e2e/smartvol_ops_test.go
```diff
@@ -407,6 +407,83 @@ func testSmartVolumeDistributeDisperse(t *testing.T) {
...
+	// Max-brick-size is more than request size
+	createReq = api.VolCreateReq{
+		Name:               smartvolname,
+		Size:               20 * gutils.MiB,
+		ReplicaCount:       3,
+		MaxBrickSize:       30 * gutils.MiB,
+		SubvolZonesOverlap: true,
+	}
+	volinfo, err = client.VolumeCreate(createReq)
+	r.Nil(err)
```
atinmu: Shouldn't this be r.NotNil (err) , how is this going to succeed if the MaxBrickSize is greater than Size ?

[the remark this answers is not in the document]
aravindavk: max subvol size is derived based on volume type and `req.MaxBrickSize`. This calculation is same for Arbiter/Replica. Special handling only for Disperse volume.

aravindavk: `MaxBrickSize > VolumeSize` condition will always satisfy. So this satisfies user's condition.

aravindavk:
> How would this max brick size be leveraged in PVC request? I see that if this isn't passed then we don't decide on the distribute count.

Two ways we can distribute now,

- Specify distribute count in the Volume create request
- Specify max-brick-size and calculate the distribute count automatically.

Max brick size can be added in Storage class definition irrespective of Volume sizes. That will be passed with every volume create request. (For example, `max-brick-size=10GiB`). If Volume requested of size less than that value will be created without distribute.

Currently `max-brick-size` takes precedence over the `distribute-count` parameter in Volume create request. That means if request contains `distribute=3`, `max-brick-size=512M` and `size=1G` then it will create volume with distribute count 2(Distribute count specified in request is ignored). Let me know if you feel this precedence needs to be reversed.(If distribute count is specified then do not consider `max-brick-size`)

atinmu:
@aravindavk I'm not sure if distributeCount and maxBricksSize parameters input value should co-exist but if they do then I agree that maxBrickSize should take precedence as distribute count should be more internal and should be managed intelligently by GD2.

atinmu:
@aravindavk @Madhu-1 don't we need a respective change in the gluster-csi-driver once this gets in? If so, I think we need an issue tracking this.

Madhu-1:
@atinmu currently we are just making use of size and volume type in volume create,we will make required changes in CSI driver.

rishubhjain:
Fixes: #1398

aravindavk:
retest this please

Issue #1391, opened 2018-12-12
Without starting volume, volume stop is success


Akarsha-rai:
### Observed behavior
When volume is in created state, stopping volume is success

### Expected/desired behavior
When volume is in created state, stopping volume should fail.

### Details on how to reproduce (minimal and precise)
1. Create a volume
```
 glustercli volume create testvol replica 3 10.70.35.121:/bricks/brick1/testvol 10.70.35.122:/bricks/brick1/testvol 10.70.35.4:/bricks/brick1/testvol 10.70.35.121:/bricks/brick2/testvol 10.70.35.122:/bricks/brick2/testvol 10.70.35.4:/bricks/brick2/testvol --create-brick-dir
testvol Volume created successfully
Volume ID:  464b4ef2-e933-4c1f-b839-1143f13c513d

glustercli volume info testvol

Volume Name: testvol
Type: Distributed-Replicate
Volume ID: 464b4ef2-e933-4c1f-b839-1143f13c513d
State: Created
Transport-type: tcp
Options:
    performance/readdir-ahead: off
    performance/write-behind: off
    cluster/replicate.self-heal-daemon: on
    performance/io-cache: off
    performance/md-cache: off
    performance/open-behind: off
    performance/quick-read: off
    performance/read-ahead: off
Number of Bricks: 2 x 3 = 6
Brick1: 10.70.35.121:/bricks/brick1/testvol
Brick2: 10.70.35.122:/bricks/brick1/testvol
Brick3: 10.70.35.4:/bricks/brick1/testvol
Brick4: 10.70.35.121:/bricks/brick2/testvol
Brick5: 10.70.35.122:/bricks/brick2/testvol
Brick6: 10.70.35.4:/bricks/brick2/testvol
```
2. Stop the volume without starting it.
```
glustercli volume stop testvol
Volume testvol stopped successfully

 glustercli volume info testvol

Volume Name: testvol
Type: Distributed-Replicate
Volume ID: 464b4ef2-e933-4c1f-b839-1143f13c513d
State: Stopped
Transport-type: tcp
Options:
    performance/write-behind: off
    cluster/replicate.self-heal-daemon: on
    performance/io-cache: off
    performance/md-cache: off
    performance/open-behind: off
    performance/quick-read: off
    performance/read-ahead: off
    performance/readdir-ahead: off
Number of Bricks: 2 x 3 = 6
Brick1: 10.70.35.121:/bricks/brick1/testvol
[...]

PR #1392, opened 2018-12-12
volume stop: return error if volume is not in stated state

Madhu-1:
add a check to test whether the volume
is in started state or not during volume stop.
Fixes: #1391


glusterd2/commands/volumes/volume-stop.go
```diff
@@ -143,6 +143,11 @@ func volumeStopHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if volinfo.State != volume.VolStarted {
```
vpandey-RH: @Madhu-1  Can we remove  if (volinfo.State == volume.VolStopped) condition ?

rishubhjain:
@Madhu-1 CI failure please check.

glusterd2/commands/volumes/volume-stop.go
Madhu-1: in both the conditions we are returning an appropriate error message, one is volume already in stopped state and another one is volume not started.

rishubhjain: It's required, if the volume is already stopped then no need of stopping it again

vpandey-RH: @Madhu-1  Can't we have a single message saying "volume is not in started state". I don't see the need for an extra check.

Madhu-1:
@aravindavk PTAL

glusterd2/commands/volumes/volume-stop.go
Madhu-1: if we send only one error message for both the scenario it might be difficult for the user to identify what has happened.
Ex: currently in CSI driver if the volume is already in the stopped state it will proceed for volume deletion.
if we give single error message it might be difficult for the consumer to take appropriate action.

aravindavk:
Conflict need to be resolved.

Madhu-1:
@aravindavk done

aravindavk:
retest this please

rishubhjain:
@Madhu-1 Can you correct your PR commit message? s/stated/started

rishubhjain:
retest this please

Madhu-1:
@rishubhjain Thanks, updated commit message

PR #1394, opened 2018-12-12
Removing volume stop from replace-brick e2e test case

rishubhjain:
Since volume was not started, volume stop was unnecessarily placed in the e2e test case.

### DIFF
--- a/e2e/smartvol_ops_test.go
+++ b/e2e/smartvol_ops_test.go
@@ -112,6 +112,10 @@ func testReplaceBrick(t *testing.T) {
 	volinfo, err = client.VolumeCreate(createReq)
 	r.Nil(err)
 
+	// Start volume
+	err = client.VolumeStart(smartvolname, true)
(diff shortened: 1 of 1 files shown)


Madhu-1:
@rishubhjain why can't you add volume start and volume stop as part of replacing brick testing
why it affected anything?

rishubhjain:
@Madhu-1 I have already covered a scenario where volume is started and stopped (https://github.com/gluster/glusterd2/blob/master/e2e/smartvol_ops_test.go#L78), so I am covering another scenario where a brick is replaced even when the volume is not started.

Madhu-1:
the above scenario covers the starting of volume and then replacing a brick, in the next scenario its good to cover starting of volume after replacing the brick. does this sound good?

rishubhjain:
@Madhu-1 Made the change.

Madhu-1:
@aravindavk PTAL

Issue #1403, opened 2018-12-13
old glusterfsd process still running after the brick mux is enabled.


Shrivaibavi:
### Observed behavior

When i did ps -ef | grep glusterfsd i saw the old glusterfsd process still running after the brick mux is enabled.

```

[root@dhcp35-30 ~]# glustercli volume list
+--------------------------------------+------+-----------------------+---------+-----------+--------+
|                  ID                  | NAME |         TYPE          |  STATE  | TRANSPORT | BRICKS |
+--------------------------------------+------+-----------------------+---------+-----------+--------+
| e42c6604-6cf7-4aac-bd5e-dfe0cd674d4f | abc  | Replicate             | Started | tcp       | 3      |
| dd797f61-0c20-4894-8447-9b734e21f63b | dif  | Replicate             | Started | tcp       | 3      |
| 7c0b0f49-9c61-4641-b688-c0a478924ba9 | xy   | Distributed-Replicate | Created | tcp       | 6      |
| e4f4ba5b-89c9-4b53-a2d3-5e7e6935804a | xyz  | Distributed-Replicate | Created | tcp       | 6      |
+--------------------------------------+------+-----------------------+---------+-----------+--------+

[root@dhcp35-30 ~]# pgrep glusterfsd
20932
26843

[root@dhcp35-30 ~]# glustercli volume status
Volume : abc
+--------------------------------------+--------------+---------------------+--------+-------+-------+
|               BRICK ID               |     HOST     |        PATH         | ONLINE | PORT  |  PID  |
+--------------------------------------+--------------+---------------------+--------+-------+-------+
| cd207ed3-5fea-42a8-94aa-a6699c70ce08 | 10.70.35.30  | /bricks/brick0/abc1 | true   | 39728 | 26843 |
| 77144bf5-58e4-4945-948c-e586c9e14c7f | 10.70.35.106 | /bricks/brick0/abc2 | true   | 45344 | 14064 |
| 9007a1a1-980c-4f04-9d1a-680782860847 | 10.70.35.240 | /bricks/brick0/abc3 | true   | 33356 | 29968 |
+--------------------------------------+--------------+---------------------+--------+-------+-------+
Volume : dif
+--------------------------------------+--------------+--------------------+--------+-------+-------+
|               BRICK ID               |     HOST     |        PATH        | ONLINE | PORT  |  PID  |
+--------------------------------------+--------------+--------------------+--------+-------+-------+
[...]

Issue #1405, opened 2018-12-13
Demultiplexing fails when doing a force start


vpandey-RH:
Consider a scenario where multiplexing is on and all bricks are multiplexed into one brick.
Now, if brick-multiplex is set as off and a force start is performed instead of a stop and a start, then we will see that demultiplexing of all bricks will fail.

The main reason for this is that force start is meant to start only those bricks which are offline and not affect started bricks. Therefore, all multiplexed and started bricks are not demultiplexed.

### Observed behavior
Multiplexing continues to prevail even when brick-multiplexing is off and volume is started forcefully.

### Expected/desired behavior
Bricks should be demultiplexed when doing a force start after setting brick-multiplex off.

### Details on how to reproduce (minimal and precise)
1- Create a distribute volume with 3 bricks
2- Set cluster.brick-multiplex off
3- Start the volume 
4- Check if all bricks are multiplexed into one
5- Set brick-multiplex off
6- Do a force start on the volume

### Information about the environment:

- Glusterd2 version used (e.g. v4.1.0 or master): 
- Operating system used: 
- Glusterd2 compiled from sources, as a package (rpm/deb), or container: 
- Using External ETCD: (yes/no, if yes ETCD version):
- If container, which container image: 
- Using kubernetes, openshift, or direct install: 
- If kubernetes/openshift, is gluster running inside kubernetes/openshift or outside: 

### Other useful information

- glusterd2 config files from all nodes (default /etc/glusterd2/glusterd2.toml)
- glusterd2 log files from all nodes (default /var/log/glusterd2/glusterd2.log)
- ETCD configuration
- Contents of `uuid.toml` from all nodes (default /var/lib/glusterd2/uuid.toml)
- Output of `statedump` from any one of the node

### Useful commands

- To get glusterd2 version
    ```
    glusterd2 --version
    ```
- To get ETCD version
    ```
    etcd --version
    ```
- To get output of statedump
    ```
    curl http://glusterd2-IP:glusterd2-Port/statedump
    ```

aravindavk:
We should not demultiplex the bricks of running volume. This will affect the high availability of the volume. Brickmux enable/disable should be considered only while starting the new bricks.

Messaging we need to give when brickmux is enabled or disabled is "All new volumes will be brickmuxed, and already running volumes will not be affected unless it is restarted"

**Note**: 

- We need to make sure all other volume operations work even though some bricks are muxed and some are not.
[...]

PR #1408, opened 2018-12-14
Kill old glusterfsds post multiplexing

vpandey-RH:
Fixes - [#1403](https://github.com/gluster/glusterd2/issues/1403)

ISSUE - Force start doesn't kill bricks and respawn but only spawns killed ones,
so in case when we set multiplex as on and then do a force start on volume,
the old glusterfsds are still running.

Fix - Look for an already running glusterfsd for the brick to be multiplexed and terminate its glusterfsd
only after a target brick for currrent brick (brick to be multiplexed) has been found.

### DIFF
--- a/glusterd2/brickmux/multiplex.go
+++ b/glusterd2/brickmux/multiplex.go
@@ -31,6 +31,27 @@ func Multiplex(b brick.Brickinfo, v *volume.Volinfo, volumes []*volume.Volinfo, 
 		return err
 	}
 
+	// Kill already running fsd for the brick in consideration
+	// Helpful in case of force start where no running glusterfsds are
+	// effected and thus even after multiplexing, old glusterfsds
+	// exist, take this action only when target brick for brick in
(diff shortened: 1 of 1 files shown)


vpandey-RH:
retest this please

aravindavk:
> Fix - Look for an already running glusterfsd for the brick to be multiplexed and terminate its glusterfsd
only after a target brick for currrent brick (brick to be multiplexed) has been found.

As discussed here https://github.com/gluster/glusterd2/issues/1405#issuecomment-447262010, Volume availability will get affected if running bricks are stopped and restarted.

vpandey-RH:
@aravindavk In this case we will be asking the brick to continue working through the glusterfsd that it will be multiplexed into, so its old glusterfsd will not be of any use. If we keep on running the old glusterfsds there could be a case where 1000 glusterfsds are running at one point in time on large setups.

atinmu:
@vpandey-RH Please abandon this PR as discussed earlier.

PR #1410, opened 2018-12-14
Fix xlator .so read failure if no options

aravindavk:
If `.options` is null in xlator_api then glusterd2
was failing. This PR adds nil check to prevent this failure.

### DIFF
--- a/glusterd2/xlator/load.go
+++ b/glusterd2/xlator/load.go
@@ -123,19 +123,22 @@ func loadXlator(xlPath string) (*Xlator, error) {
 	// Parent directory name where xlator .so file exists
 	xl.Category = filepath.Base(filepath.Dir(xlPath))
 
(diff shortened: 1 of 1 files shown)


glusterd2/xlator/load.go
```diff
@@ -123,19 +123,22 @@ func loadXlator(xlPath string) (*Xlator, error) {
...
+	// If no options exists in xlator_api table
+	if p != nil {
+		soOptions := (*[maxOptions]C.volume_option_t)(p)
+		for _, option := range soOptions {
 
-		// identify sentinel NULL key which marks the end of options
-		if option.key[0] == nil {
-			break
-		}
+			// identify sentinel NULL key which marks the end of options
```
rishubhjain: Spaces instead of tab

Issue #1389, opened 2018-12-12
minimum brick size validation during Volume create


aravindavk:
Auto provisioned volume create validates the minimum volume size but does not validate the minimum brick size. Following command fails 

```
glustercli volume create gv4 --size 500M --distribute 50 --subvols-zones-overlap
```

Validation required after calculating brick size

aravindavk:
@harigowtham do you want to take this up?

harigowtham:
@aravindavk. Yes, I'll take it up.

PR #1417, opened 2018-12-17
minimum brick size validation during Volume create

harigowtham:
Auto provisioned volume does not validate the minimum brick size.
If the size is too small then the volume creation errors out.

The validation is done after the bricksize calculation.

Fixes: #1389

### DIFF
--- a/glusterd2/bricksplanner/planner.go
+++ b/glusterd2/bricksplanner/planner.go
@@ -136,6 +136,9 @@ func getBricksLayout(req *api.VolCreateReq) ([]api.SubvolReq, error) {
 		for j := 0; j < subvolplanner.BricksCount(); j++ {
 			eachBrickSize := subvolplanner.BrickSize(j)
 			brickType := subvolplanner.BrickType(j)
+			if eachBrickSize < minBrickSize {
+				return nil, errors.New("brick size is too small")
+			}
 			eachBrickTpSize := uint64(float64(eachBrickSize) * req.SnapshotReserveFactor)
 
 			bricks = append(bricks, api.BrickReq{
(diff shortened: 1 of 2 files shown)


glusterd2/bricksplanner/planner.go
```diff
@@ -116,6 +116,9 @@ func getBricksLayout(req *api.VolCreateReq) ([]api.SubvolReq, error) {
 	subvolSize := req.Size
 	if numSubvols > 1 {
 		subvolSize = subvolSize / uint64(numSubvols)
+		if subvolSize < minBrickSize {
```
aravindavk: Brick size in case of disperse Volume will be less than subvolsize.(that is `subvolsize / disperse_data_count`)

harigowtham: this situation is taken care of.

atinmu:
@harigowtham Please update the commit message with the github issue link (#1389)

harigowtham:
> @harigowtham Please update the commit message with the github issue link (#1389)

Yes @atinmu, I have updated it in the commit message. As I missed it in the first push, the first comment in the conversation is not reflecting it.

atinmu:
Fixes #1389

glusterd2/bricksplanner/planner.go
```diff
@@ -136,6 +136,9 @@ func getBricksLayout(req *api.VolCreateReq) ([]api.SubvolReq, error) {
 		for j := 0; j < subvolplanner.BricksCount(); j++ {
 			eachBrickSize := subvolplanner.BrickSize(j)
 			brickType := subvolplanner.BrickType(j)
+			if eachBrickSize < minBrickSize {
+				return nil, errors.New("Brick size is too small")
```
aravindavk: start errorstring with lower case

harigowtham: done

PR #1426, opened 2018-12-19
cluster-lock: fix getting cluster wide etcd lock

oshankkumar:
Instead of using a global etcd session create a new etcd session
everytime a lock is created.
If we use existing session every time when a lock is created then
etcd lock will use same leaseId. It leads to faulty lock behaviour.


kshlm:
ok to test

kshlm:
add to whitelist

Issue #1431, opened 2018-12-20
When Peer probed using API's, the online parameter shows "False"


Shrivaibavi:
### Observed behavior
```
[root@dhcp35-30 ~]# cat peer_add.json 
{ 
"addresses": ["dhcp35-240.lab.eng.blr.redhat.com"]
}

[root@dhcp35-30 ~]# curl -X POST http://localhost:24007/v1/peers --data @peer_add.json -H 'Content-Type: application/json'
{"id":"b7bd3edf-bbaf-4dec-af86-3f369ad1fb8e","name":"dhcp35-240.lab.eng.blr.redhat.com","peer-addresses":["dhcp35-240.lab.eng.blr.redhat.com:24008","10.70.35.240:24008"],"client-addresses":["127.0.0.1:24007","10.70.35.240:24007"],"online":false,"metadata":{"_zone":"b7bd3edf-bbaf-4dec-af86-3f369ad1fb8e"}}
```

### Expected/desired behavior

As the peer probe is successful, Can we expect the online parameter to show as True ?

### Information about the environment:

- Glusterd2 version used (e.g. v4.1.0 or master): 
- Operating system used: 
- Glusterd2 compiled from sources, as a package (rpm/deb), or container: 
- Using External ETCD: (yes/no, if yes ETCD version):
- If container, which container image: 
- Using kubernetes, openshift, or direct install: 
- If kubernetes/openshift, is gluster running inside kubernetes/openshift or outside: 

### Other useful information

- glusterd2 config files from all nodes (default /etc/glusterd2/glusterd2.toml)
- glusterd2 log files from all nodes (default /var/log/glusterd2/glusterd2.log)
- ETCD configuration
- Contents of `uuid.toml` from all nodes (default /var/lib/glusterd2/uuid.toml)
- Output of `statedump` from any one of the node

### Useful commands

- To get glusterd2 version
    ```
    glusterd2 --version
    ```
- To get ETCD version
    ```
    etcd --version
    ```
- To get output of statedump
    ```
    curl http://glusterd2-IP:glusterd2-Port/statedump
    ```

PR #1433, opened 2018-12-24
Add online status to peer add response

Madhu-1:
Fixes: #1431

### DIFF
--- a/glusterd2/commands/peers/addpeer.go
+++ b/glusterd2/commands/peers/addpeer.go
@@ -141,6 +141,7 @@ func createPeerAddResp(p *peer.Peer) *api.PeerAddResp {
 	return &api.PeerAddResp{
 		ID:              p.ID,
 		Name:            p.Name,
+		Online:          true,
 		PeerAddresses:   p.PeerAddresses,
 		ClientAddresses: p.ClientAddresses,
 		Metadata:        p.Metadata,
--- a/e2e/peer_ops_test.go
+++ b/e2e/peer_ops_test.go
@@ -51,6 +51,7 @@ func TestAddRemovePeer(t *testing.T) {
 
 	peerinfo, err := client.PeerAdd(peerAddReq)
 	r.Nil(err)
+	r.True(peerinfo.Online)
 
 	_, err = client.GetPeer(peerinfo.ID.String())
 	r.Nil(err)


glusterd2/commands/peers/addpeer.go
```diff
@@ -141,6 +141,7 @@ func createPeerAddResp(p *peer.Peer) *api.PeerAddResp {
 	return &api.PeerAddResp{
 		ID:              p.ID,
 		Name:            p.Name,
+		Online:          true,
```
oshankkumar: I think instead of hardcoding online as true, can we check liveness status of peer?

atinmu: Shouldn't we display this in the cli as well?

Madhu-1: @oshankkumar This is during the peer add, I don't think we need to check the liveness status.
if the node is available then only we can make that peer add is successful.

do we need to check liveness status during peer add also?

Madhu-1: @atinmu showing this field in peer add response won't make any sense in my opinion.
let me know your thoughts on this.

atinmu: Then why is it different to be shown in the API response especially we know that online field will be anyways true. To me, uniformity on the response should be maintained as much as possible in both API & cli response.

vpandey-RH: @Madhu-1  If we are considering that there is no possibility that node will go down between the time we call joinCluster() and returning the response of peer add, then hard coding true is fine, even when large number of peers are added parallely.

Madhu-1: we are having one common structure for peer add, peer edit and get peer response.
we are making use of the online field in peer list and peer get cli command.

aravindavk: I agree with @Madhu-1, if Peer add successful then that means Peer is online. Peer can go down after peer add, but that status can be obtained from Peer status.

Madhu-1: if glusterd2 goes down in between it will take 5 seconds to make the peer state as offline.
even if you check the liveness check it won't make much difference, liveness timeout is around 5 seconds.

even if you get so many requests at a time, gd2 won't take 5 seconds to respond back.

for the corner cases, I have added the check, but it won't make much difference, for me, it will take some more time to send back the response as it includes the get request from ETCD

Madhu-1: let me know it make sense to add the check

aravindavk: Not required I think, If we say peer add is successful and online as false then again it adds confusion as now.

Issue #1303, opened 2018-10-31
Extend brick multiplexing to volume expand (add brick) operation


prashanthpai:
Brick multiplexing needs to be extended to volume expand operation (add brick). PR #1301 implements brick multiplexing for volume start, volume stop, volume status operations.

Related issue #219

PR #1434, opened 2018-12-24
Bmux support for volume expand

vpandey-RH:
Fixes [#1303](https://github.com/gluster/glusterd2/issues/1303)

Issue -
Add bmux support for volume-expand feature

Fix -
During starting the newly added bricks, check if bmux is enabled. If it is enabled
then traverse over all the local bricks present in the newly added bricks list and
 look for a target brick from any of the started volumes(which have a brick on the node on which the current brick exists) present in the cluster.
If a target brick is not found due to any reason then start a new seperate glusterfsd for that particular brick.


atinmu:
@vpandey-RH I'd like to have an issue tagged to this commit.

Issue #1436, opened 2018-12-26
Peer add failed when the removed peer is added back to the trusted pool


PrasadDesala:
Peer add failed when the removed peer is added back to the trusted pool.

### Observed behavior

[root@gluster-kube1-0 /]# glustercli peer add b6d7a82f-2a8d-4ce6-9d51-a404aec74997
Peer add failed

Response headers:
X-Gluster-Cluster-Id: 796a0a22-29c9-48b4-b26a-0153b8850a58
X-Gluster-Peer-Id: d0576643-b50e-4a95-831b-1222aa04c5dd
X-Request-Id: a2af2579-d305-4c50-8eef-3ae7c5a40e99

Response body:
failed to send join cluster request

### Expected/desired behavior

Peer add should succeed

### Details on how to reproduce (minimal and precise)
1) Have a 3 node gcs setup with all the 3 nodes in trusted pool.
2) Remove a node(n3)  from the trusted pool "glustercli peer remove 5187223f-d7a7-4b42-ab39-5afc4036e344"
3) Now, add back the node (n3) to the trusted pool (ID of n3 has changed)
glustercli peer add b6d7a82f-2a8d-4ce6-9d51-a404aec74997

### Information about the environment:

- Glusterd2 version used (e.g. v4.1.0 or master): v6.0-dev.93.gitd4ca2cb
- Operating system used: CentOS 7.5.1804
- Glusterd2 compiled from sources, as a package (rpm/deb), or container:  Container
- Using External ETCD: (yes/no, if yes ETCD version): Yes version: 3.3.8
- If container, which container image: 
- Using kubernetes, openshift, or direct install: 
- If kubernetes/openshift, is gluster running inside kubernetes/openshift or outside: Kubernetes

Seeing below error in glusterd2 logs:

time="2018-12-26 12:18:39.376744" level=error msg="failed RPC call" error="rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = \"transport: Error while dialing dial tcp: lookup b6d7a82f-2a8d-4ce6-9d51-a404aec74997 on 10.233.0.3:53: no such host\"" remote="b6d7a82f-2a8d-4ce6-9d51-a404aec74997:24008" rpc=PeerService.Join source="[peer-rpc-clnt.go:47:peers.(*peerSvcClnt).JoinCluster]"
[...]

PR #1438, opened 2018-12-26
check connectivity to peer before sending join cluster request.

oshankkumar:
Fixes: #1436

### DIFF
--- a/e2e/peer_ops_test.go
+++ b/e2e/peer_ops_test.go
@@ -100,3 +100,38 @@ func TestAddRemovePeer(t *testing.T) {
 	err = client.PeerRemove(g2.PeerID())
 	r.Nil(err)
 }
+
+func TestPeerAddFailure(t *testing.T) {
+	r := require.New(t)
+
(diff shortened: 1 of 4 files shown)


rishubhjain:
@oshankkumar Thanks for the PR, if possible please add a small e2e test case for peer add failure when peerID is provided instead of peer address (not a priority.).

oshankkumar:
> @oshankkumar please check why we have not received any error while parsing remote address at https://github.com/gluster/glusterd2/blob/master/glusterd2/commands/peers/addpeer.go#L60

we are just splitting host and port, and if port is missing then we set default port in remote address. Since a hostname could be anything not necessarily an IP addr, so it won't return any error even if we pass any random string as the hostname(in this case it is receiving a uuid ).

oshankkumar:
> @oshankkumar Thanks for the PR, if possible please add a small e2e test case for peer add failure when peerID is provided instead of peer address (not a priority.).

@rishubhjain  added test case for this scenario.

rishubhjain:
retest this please

Issue #1445, opened 2018-12-28
Modifed volume set options are not reflecting in "glustercli volume get <volname> all"


PrasadDesala:
I have changed the default value of distribute.min-free-disk from 10% to 60%. The new values are reflected in the glustercli volume info output but the changes are not reflected in the output of "glustercli volume get <volname> all".

Volume Name: pvc-d16a6e66-0a70-11e9-8594-525400e31a9f
Type: Replicate
Volume ID: 47c213ce-24ff-4fd6-bac2-ebbf5cc69dce
State: Started
Capacity: 1.0 GiB
Transport-type: tcp
Options:
    cluster/replicate.self-heal-daemon: on
    debug/io-stats.count-fop-hits: on
    performance/io-cache.io-cache: off
    performance/quick-read.quick-read: off
    performance/read-ahead.read-ahead: off
    performance/write-behind.write-behind: off
    cluster/distribute.min-free-disk: 60%
    debug/io-stats.latency-measurement: on
    performance/md-cache.md-cache: off
    performance/open-behind.open-behind: off
    performance/readdir-ahead.readdir-ahead: off
Number of Bricks: 3
Brick1: gluster-kube2-0.glusterd2.gcs:/var/run/glusterd2/bricks/pvc-d16a6e66-0a70-11e9-8594-525400e31a9f/subvol1/brick1/brick
Brick2: gluster-kube3-0.glusterd2.gcs:/var/run/glusterd2/bricks/pvc-d16a6e66-0a70-11e9-8594-525400e31a9f/subvol1/brick2/brick
Brick3: gluster-kube1-0.glusterd2.gcs:/var/run/glusterd2/bricks/pvc-d16a6e66-0a70-11e9-8594-525400e31a9f/subvol1/brick3/brick

[root@gluster-kube1-0 /]# glustercli volume get pvc-d16a6e66-0a70-11e9-8594-525400e31a9f all
+-------------------------------+----------+--------+---------------+--------------+
|             NAME              | MODIFIED | VALUE  | DEFAULT VALUE | OPTION LEVEL |
+-------------------------------+----------+--------+---------------+--------------+
| nufa.lookup-unhashed          | no       | on     | on            | Basic        |
| nufa.min-free-disk            | no       | 10%    | 10%           | Basic        |
| nufa.min-free-inodes          | no       | 5%     | 5%            | Basic        |
| nufa.rebalance-stats          | no       | off    | off           | Basic        |
[...]

PR #1446, opened 2018-12-29
Fixing volume set options functionality

rishubhjain:
Fixes:https://github.com/gluster/glusterd2/issues/1445
Signed-off-by: rishubhjain <rishubhjain47@gmail.com>

### DIFF
--- a/glusterd2/commands/volumes/volume-options-get.go
+++ b/glusterd2/commands/volumes/volume-options-get.go
@@ -2,6 +2,7 @@ package volumecommands
 
 import (
 	"net/http"
+	"strings"
 
 	"github.com/gluster/glusterd2/glusterd2/options"
 	restutils "github.com/gluster/glusterd2/glusterd2/servers/rest/utils"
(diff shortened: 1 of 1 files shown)


rishubhjain:
@aravindavk Please check

glusterd2/commands/volumes/volume-options-get.go
```diff
@@ -78,7 +78,7 @@ func createVolumeOptionsGetResp(volinfo *volume.Volinfo) *api.VolumeOptionsGetRe
 		// now return all options
 		for _, opt := range xl.Options {
 			for _, k := range opt.Key {
-				optName := xl.ID + "." + k
+				optName := xl.Category + "/" + xl.ID + "." + k
```
aravindavk: The change is good for most cases. But if a option is set with the prefix of template name then that will not be catch. For example `glustercli volume set gv1 brick.changelog.rollover-time 30`. This will be stored in volinfo as `brick.features/changelog.rollover-time`. In this case above comparison is not sufficient. `createVolumeOptionGetResp` function needs modification to handle this case.

atinmu: @rishubhjain can you please refresh the PR with the suggested changes?

PR #1447, opened 2018-12-30
txn Engine: refactor code and fix a minor issue in step synchronization

oshankkumar:
If a tranction got failed on some node, then all nodes who are waiting
for txn to get synchronized should cancel the ongoing txn on their respective
nodes and return.


oshankkumar:
retest this please

oshankkumar:
@kshlm can you please review this.

glusterd2/transactionv2/executor.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// executorImpl is a concrete implementation of Executor
+type executorImpl struct {
+	txnManager  TxnManager
+	stepManager StepManager
+	selfNodeID  uuid.UUID
+}
+
+// IsTxnPending returns true if the give txn is in pending state
```
rishubhjain: s/give/given

glusterd2/transactionv2/executor.go
```diff
@@ -0,0 +1,177 @@
...
+
+	logger.WithField("stepname", step.DoFunc).Debug("running step func on node")
+
+	// a synchronized step is executed only after all previous steps
+	// have been completed successfully by all involved peers.
+	if step.Sync {
+		logger.WithField("stepname", step.DoFunc).Debug("synchronizing txn step")
+		if err := e.stepManager.SyncStep(ctx, stepIndex, txn); err != nil {
+			logger.WithFields(log.Fields{
+				"error":    err,
```
rishubhjain: trivial: I think you can log this with  `.WithError(err)` just to maintain uniformity

oshankkumar: since i am using more than one field in log, so using WithError seems to be redundant, because WithError just add one field.

rishubhjain: I agree, but I think its a part of the convention that we follow. `logger.WithError(err).WithFields(log.Field("stepname": step.DoFunc)..Error("encounter an error in synchronizing txn step")`
should also work. This is not very important, you can skip it if you want.

oshankkumar:
retest this please

PR #1454, opened 2019-01-02
txn Lock: acquire cluster locks while creating a txn

oshankkumar:
- all cluster locks should be aquired before starting a txn.
   We should acquire locks at the time of creating a txn.

 - use namespaced etcd client for creating a etcd Session in
   transaction.Lock and CleanupHandler


oshankkumar:
@kshlm please review this

PR #1455, opened 2019-01-02
Improve brick mux e2e test suite

vpandey-RH:
Handle cases causing spurious failures in brickmux tests. Increase waiting time for newly added bricks to SignIn in case of volume-expand and
add sleep wherever a new gd2 instance is re-spawned to give enough time
to bricks to sign in properly and multiplex.


vpandey-RH:
retest this please

vpandey-RH:
retest this please

vpandey-RH:
retest this please

vpandey-RH:
retest this please

e2e/brickmux_test.go
```diff
@@ -255,6 +255,8 @@ func TestBrickMux(t *testing.T) {
 	r.Nil(err)
 	r.True(gd.IsRunning())
 
+	time.Sleep(5 * time.Second)
```
atinmu: What's the justification of waiting for 5 seconds here?

vpandey-RH: @atinmu  Oh yeah, forgot to put a comment here. I have put a 5 sec sleep here just to give the bricks sometime to SignIn after a new gd2 instance has spawned.

Madhu-1: is there a way to remove this timeout, can we do something in during volume create?
IMO doing sleep in a test case is not a good idea.

e2e/brickmux_test.go
```diff
@@ -400,7 +402,7 @@ func TestBrickMux(t *testing.T) {
 			index = index + 1
 		}
 		// Wait for added bricks' glusterfsds to Sign In
-		time.Sleep(5 * time.Millisecond)
+		time.Sleep(70 * time.Millisecond)
```
Madhu-1: @vpandey-RH will this timeout will behave differently in different environments based on the network?

e2e/brickmux_test.go
vpandey-RH: @Madhu-1  I know that doing a sleep is not the best idea here but the issue here is that the SignIn done by bricks sometime is not the quickest, sometimes they signin instantly, sometimes it takes time to signin and that's why we keep on seeing spurious failures in brickmux tests. Now the only solution to this is to make SignIn process as quick as possible.

@atinmu  I had a discussion with @aravindavk sometime back about whether we can somehow trigger a SignIn from GD2 side as soon as a glusterfsd spawns.

e2e/brickmux_test.go
vpandey-RH: @Madhu-1 I opted for this timeout just to be safe that the test should pass almost everytime. I am still retesting this and would like to keep trigerring tests atleast for today.

atinmu:
@vpandey-RH What we can do here is to keep on checking for the online or port flag till it gets updated at an interval of 1 second.

rishubhjain:
> @vpandey-RH What we can do here is to keep on checking for the online or port flag till it gets updated at an interval of 1 second.

But we should limit the number of checks as well.

atinmu:
> But we should limit the number of checks as well.

A snippet of how this is done at glusterfs tests:

TEST $CLI_1 peer probe $H3;                                                        
EXPECT_WITHIN $PROBE_TIMEOUT 2  peer_count

The above probes a new peer and then continue to check for peer count to become 2 at every seconds till the threshold PROBE_TIMEOUT number of seconds expire. So yes, there has to be a limit. But with this, you don't predefine a sleep for certain seconds and the overall delay depends on how fast the system works.

vpandey-RH:
retest this please

atinmu:
@vpandey-RH This PR is a candidate for GCS/0.5 release. I see the changes suggested at https://github.com/gluster/glusterd2/pull/1455#issuecomment-451065338 hasn't been implemented yet. What's the plan?

atinmu:
@vpandey-RH can you please take out DO NOT MERGE message from the commit?

atinmu:
Fixes https://github.com/gluster/glusterd2/pull/1455

Issue #1459, opened 2019-01-03
glustershd needs to be restarted for volume start/stop operations


atinmu:
Based on code reading, it was found that glustershd wasn't restarted on volume start/stop operations which was resulting shd process to not having the latest volfile and it was continuing to attempt to connect to the stale bricks and GD2 logs were flooding with 

time="2019-01-03 15:43:09.300668" level=info msg="client disconnected" address="172.17.0.4:996" server=sunrpc source="[server.go:109:sunrpc.(*SunRPC).pruneConn]"
time="2019-01-03 15:43:09.815217" level=info msg="client connected" address="172.17.0.3:996" server=sunrpc source="[server.go:148:sunrpc.(*SunRPC).acceptLoop]" transport=tcp
time="2019-01-03 15:43:09.818169" level=error msg="registry.SearchByBrickPath() failed for brick" brick=/tmp/b1 error="SearchByBrickPath: port for brick /tmp/b1 not found" source="[rpc_prog.go:104:pmap.(*GfPortmap).PortByBrick]"

PR #1460, opened 2019-01-03
restart glustershd on volume start/stop operation

atinmu:
glustershd process needs to be restarted for every volume start/stop
operations to ensure that glustershd can run with the latest volfiles.
Currently due to the lack of client graph switch issue, a reconfigure
isn't possible for such case with out a restart.

volume set operation is still a TODO item and this PR doesn't cover
this.

Fixes: #1459


atinmu:
I'm seeing performance implications because of this. Please don't merge it.

PR #1464, opened 2019-01-07
Skip lvs command during thinpool delete

aravindavk:
If a Volume is cloned from another volume, the bricks of cloned volume
will also belong to same LV thinpool as the original Volume. So before
removing the thinpool a check was made to confirm if number of Lvs in
that thinpool is zero. This check was causing hang when parallel
Volume delete commands were issued.

With this PR, number of Lvs check is removed, instead of that captured
the failure of thinpool delete and handled it gracefully.

This PR also adds support for gracefully delete the volume if lv or
thinpool already deleted by previous failed transaction or manual delete.


aravindavk:
I will add tests by today

glusterd2/volume/volume-utils.go
```diff
@@ -309,26 +310,25 @@ func CleanBricks(volinfo *Volinfo) error {
...
-					"vg-name": vgname,
-					"tp-name": tpname,
-				}).Error("thinpool remove failed")
-				return err
-			}
+		err = lvmutils.RemoveLV(vgname, tpname, false)
+		// Do not remove Thinpool if the dependent Lvs exists
+		// Ignore the lvremove command failure if the reason is
+		// dependent Lvs exists
+		if err != nil && !lvmutils.IsDependentLvsError(err) && !lvmutils.IsLvNotFoundError(err) {
```
kshlm: Why is LV not found error being ignored?

pkg/lvmutils/utils.go
```diff
@@ -149,9 +149,39 @@ func UnmountLV(mountdir string) error {
 	return syscall.Unmount(mountdir, syscall.MNT_FORCE)
 }
 
+// IsDependentLvsError returns true if the error related to dependent Lvs exists
+func IsDependentLvsError(err error) bool {
+	if err == nil {
+		return false
+	}
+	return strings.Contains(err.Error(), "dependent volume(s). Proceed? [y/n]")
```
kshlm: A slight concern comparing strings here would be that string could be different for non-english locales.

kshlm: Tested this. Tried setting LANG to fr_FR.UTF-8 (french) and es_ES.UTF-8 (spanish). The output was always english. The lvm2 tools only have english strings. So a different locale should not affect this.

glusterd2/volume/volume-utils.go
aravindavk: for repeat-ability. Previously we observed the volume delete failures due to previous transaction failures.(If LVs deleted but crashed before removing from store)

pkg/lvmutils/utils.go
aravindavk: Makes sense, unfortunately lvremove command is not providing different error codes. Worst case can happen if we fail to catch this failure is failing Delete call itself with respective error printed.

aravindavk:
retest this please

aravindavk:
added tests, please review @kshlm @Madhu-1 @rishubhjain

e2e/smartvol_ops_test.go
```diff
@@ -510,6 +510,69 @@ func testSmartVolumeAutoDistributeDisperse(t *testing.T) {
 	checkZeroLvs(r)
 }
 
+func testSmartVolumeWhenCloneExists(t *testing.T) {
+	r := require.New(t)
+
+	smartvolname := "svol1"
```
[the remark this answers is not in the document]
aravindavk: for that I was getting error like file name is too long for brick volfile.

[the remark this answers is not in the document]
Madhu-1: @aravindavk then we need to add a check in volume create for max char in volume name isn't it?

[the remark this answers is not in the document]
Madhu-1: if yes, then that can be done in seperate PR.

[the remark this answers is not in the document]
aravindavk: it was not due to long volume name, brick volfile path of a snapshot volume was exceeding the limit. (Example file path which I copied from previous run `/tmp/gd2_func_test/TestSmartVolume/w3/volfiles/TestSmartVolume-Smartvol_delete_when_clone_exists-s1.77b9da5c-6d00-49f8-b532-3af2297eb523.tmp-gd2_func_test-TestSmartVolume-w3-run-gluster-snaps-TestSmartVolume-Smartvol_delete_when_clone_exists-s1-TestSmartVolume-Smartvol_delete_when_clone_exists-subvol1-brick1-brick.vol`)

aravindavk:
retest this please

aravindavk:
retest this please

glusterd2/brick/types.go
```diff
@@ -32,6 +32,13 @@ type MountInfo struct {
 	MntOpts        string
 }
 
+type DeviceInfo struct {
```
kshlm: Linter is failing here. Add a description  for this struct.

aravindavk: fixed. Also fixed issues related to deviceinfo detail update during snapshot create and clone

Issue #1468, opened 2019-01-08
Fix spurious socket connect failure issue in brick multiplexing stop code path


atinmu:
While sending parallel volume requests some of the volume stop operations were failing with the brick socket file reporting EOF.

PR #1469, opened 2019-01-08
brick-mux : fix spurious socket connect failure issue during volume stop

atinmu:
Change the logic of using the parent socket file to send the detach
request by picking it up from /proc/<pid>/cmdline where pid is the
process id of the parent brick. During Multiplex, this would avoid
creating a hard link of a socket file.

Credits : avishwan@redhat.com for helping with the utility function

Fixes: #1468


rishubhjain:
retest this please

rishubhjain:
retest this please

PR #1478, opened 2019-01-12
Skip local bricks path validation if auto provisioned volume

aravindavk:
Brick paths are generated while creating smart volume using volume
name, subvol number and brick number. So collision will not happen
when multiple smart volumes are created.


atinmu:
retest this please

atinmu:
The test failure on shd doesn't seem related and I can see the same failure on the latest GD2 smoke as well.

PR #1480, opened 2019-01-14
Add gd2 plugin and glustercli commands for trace management.

### LINKED ISSUE #1368: Tracing: Add functionality to enable/update/disable tracing using glustercli command
In addition to Glusterd2 startup options, add functionality to dynamically update, disable, enable and query the same using Gluustercli commands. See [attached](https://docs.google.com/document/d/1qH9sE58JKN1HMsihodKcfbyyoaAb774xhal9BofUQNY/edit?usp=sharing) document for additional details.

 - Must be able to enable tracing at any point in GD2 i.e., When GD2 node/pod is spawned and When GD2 node/pod is running

- Must be able to temporarily disable tracing in a running GD2 cluster and, Retain the existing tracing configuration that can be enabled later or Deregister existing Jaeger endpoints

- Must be able to update tracing options in GD2 i.e.,Register new tracer endpoints, Modify sampling frequency dynamically (Always/Probabilistic)
    
- Must be able to query and get details about the current tracing state. Status returns information like, 
  * Enabled or disabled
  * Jaeger endpoint(s)
  * Sampling type (Always/ Probabilistic)
  * Sampling frequency
    
- All tracing controls/options are applied globally. No node/pod level controls are possible.

### Observed behavior
NA

### Expected/desired behavior
NA

NA

- Operating system used: fedora27
- Glusterd2 compiled from sources, as a package (rpm/deb), or container: 
- Using External ETCD: (yes/no, if yes ETCD version):no

### Other useful information

- ETCD configuration

### Useful commands

- To get glusterd2 version
    ```
    glusterd2 --version
    ```
- To get ETCD version
    ```
    etcd --version
    ```
- To get output of statedump
    ```
    ```

sseshasa:
The trace management feature is implemented as a GD2 plugin.

The set of commits here progressively implement the gd2 plugin and appropriate commands in 
glustercli to help control tracing. Commands added allows trace enable, update and disable. This initial
set of commits build up the gd2 plugin and transactions/steps to allow enable of trace configuration. The latter commits implement the related glustercli commands like enable, get status, update and disable of tracing feature.

Closes #1368


Madhu-1:
add to whitelist

glustercli/cmd/tracing.go
```diff
@@ -0,0 +1,183 @@
...
+var traceEnableCmd = &cobra.Command{
+	Use:   "enable <jaeger-endpoint> <jaeger-agent-endpoint>",
+	Short: helpTraceEnableCmd,
+	Args:  cobra.ExactArgs(2),
+	Run: func(cmd *cobra.Command, args []string) {
+		jaegerEndpoint := args[0]
+		jaegerAgentEndpoint := args[1]
+
+		if jaegerEndpoint == "" || jaegerAgentEndpoint == "" {
+			err := errors.New("One or more Jaeger endpoints not specified")
```
Madhu-1: error message should start with small case , please update in other places also

glustercli/cmd/tracing.go
```diff
@@ -0,0 +1,183 @@
...
+					"JaegerSampleFraction": flagTraceJaegerSampleFraction,
+				}).Error("Trace Enable failed")
+			}
+			failure(errTraceEnableReqFailed, err, 1)
+		}
+		fmt.Println("Trace enable successful")
+	},
+}
+
+func samplerTypeToString(sampler int) string {
```
Madhu-1: can this be updated as 
```
func samplerTypeToString(type int) string {
sampler:=""	
switch type {
	case 0:
		sampler= "Never"
	case 1:
		sampler= "Always"
	case 2:
		sampler="Probabilistic"
	default:
		sampler="Unknown"
	}
return sampler
}
```

glustercli/cmd/tracing.go
```diff
@@ -0,0 +1,183 @@
...
+	Use:   "status",
+	Short: helpTraceStatusCmd,
+	Run: func(cmd *cobra.Command, args []string) {
+		jaegerConfigInfo, err := client.TraceStatus()
+		if err != nil {
+			failure("Error getting trace status", err, 1)
+		}
+
+		table := tablewriter.NewWriter(os.Stdout)
+		table.SetAlignment(tablewriter.ALIGN_LEFT)
```
Madhu-1: can you paste the sample output of status?

glustercli/cmd/tracing.go
```diff
@@ -0,0 +1,183 @@
...
+		table.Render()
+		fmt.Println()
+	},
+}
+
+var traceUpdateCmd = &cobra.Command{
+	Use:   "update",
+	Short: helpTraceUpdateCmd,
+	Run: func(cmd *cobra.Command, args []string) {
+		if flagTraceJaegerSampler <= 0 || flagTraceJaegerSampler > 2 {
```
Madhu-1: I see the same validation in enable also, can this and below check be moved to a new function to avoid code duplication?

pkg/tracing/tracing.go
```diff
@@ -13,58 +18,269 @@ import (
...
+	Never JaegerSamplerType = iota
+	// 'Always' - Sample every trace
+	Always
+	// 'Probabilistic' - Sample based on sample fraction
+	Probabilistic
 )
 
+// SamplerTypeToString returns string representation of sampler type.
+func SamplerTypeToString(samplerType JaegerSamplerType) string {
+	switch samplerType {
```
Madhu-1: same here, this can be simplified as suggested in one of the comment, 

can we call this function for https://github.com/gluster/glusterd2/pull/1480/files#diff-8918aaf56f805e88174f0f9e8599772bR99

pkg/tracing/tracing.go
```diff
@@ -13,58 +18,269 @@ import (
...
+	JaegerSampler JaegerSamplerType
+	// Jaeger sample fraction to use in case sampler type is probabilistic
+	JaegerSampleFraction float64
+}
+
+// Jaeger exporter
+var jaegerExporter *jaeger.Exporter
+
+// Global Jaeger trace config
+var jaegerTraceConfig *JaegerTraceConfig
```
Madhu-1: as this is a global option. what happens if gd2 restarts?

don't we need to take lock as we are updating global options

plugins/tracemgmt/api/resp.go
```diff
@@ -0,0 +1,18 @@
...
+	TracingEnabled = "enabled"
+
+	// TracingDisabled represents disabled state
+	TracingDisabled = "disabled"
+)
+
+// JaegerConfigInfo represents tracing configration
+type JaegerConfigInfo struct {
+	Status               string  `json:"tracing-status"`
+	JaegerEndpoint       string  `json:"jaeger-endpoint"`
```
Madhu-1: can we embed `SetupTracingReq`  into `JaegerConfigInfo` i see configInfo contains the same fields

plugins/tracemgmt/init.go
```diff
@@ -0,0 +1,62 @@
...
+type Plugin struct {
+}
+
+// Name returns name of plugin
+func (p *Plugin) Name() string {
+	return "tracemgmt"
+}
+
+// RestRoutes returns a list of REST API routes to register with Glusterd
+func (p *Plugin) RestRoutes() route.Routes {
```
Madhu-1: need to generate endpoints.md

plugins/tracemgmt/traceutils/store-utils.go
```diff
@@ -0,0 +1,55 @@
...
+func AddOrUpdateTraceConfig(t *tracemgmtapi.JaegerConfigInfo) error {
+	json, err := json.Marshal(t)
+	if err != nil {
+		return err
+	}
+
+	if _, err := store.Put(context.TODO(), traceMgmtPrefix, string(json)); err != nil {
+		return err
+	}
+	return nil
```
Madhu-1: return err

```
_, err := store.Put(context.TODO(), traceMgmtPrefix, string(json))
return err	
```

plugins/tracemgmt/traceutils/store-utils.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// DeleteTraceConfig deletes the JagegerConfigInfo object from the store
+func DeleteTraceConfig() error {
+	if _, err := store.Delete(context.TODO(), traceMgmtPrefix); err != nil {
```
Madhu-1: same here , return err

plugins/tracemgmt/transactions.go
```diff
@@ -0,0 +1,170 @@
...
+
+// storeTraceConfig uses the passed context key to get
+// trace config and updates it into the store.
+func storeTraceConfig(c transaction.TxnCtx, key string) error {
+	var traceConfig tracemgmtapi.JaegerConfigInfo
+	if err := c.Get(key, &traceConfig); err != nil {
+		return err
+	}
+
+	if err := traceutils.AddOrUpdateTraceConfig(&traceConfig); err != nil {
```
Madhu-1: return err

plugins/tracemgmt/transactions.go
```diff
@@ -0,0 +1,170 @@
...
+		return err
+	}
+	return nil
+}
+
+// Transaction step that stores the trace config
+func txnTracingStoreConfig(c transaction.TxnCtx) error {
+	if err := storeTraceConfig(c, "traceconfig"); err != nil {
```
Madhu-1: return err

plugins/tracemgmt/transactions.go
```diff
@@ -0,0 +1,170 @@
...
+		return err
+	}
+	return nil
+}
+
+// Transaction step that stores the trace config
+func txnTracingUndoStoreConfig(c transaction.TxnCtx) error {
+	if err := storeTraceConfig(c, "oldtraceconfig"); err != nil {
```
Madhu-1: return err

plugins/tracemgmt/transactions.go
```diff
@@ -0,0 +1,170 @@
...
+		return err
+	}
+	return nil
+}
+
+// Transaction step that reverts the trace config
+func txnTracingDeleteStoreConfig(c transaction.TxnCtx) error {
+	if err := traceutils.DeleteTraceConfig(); err != nil {
```
Madhu-1: return err

plugins/tracemgmt/transactions.go
```diff
@@ -0,0 +1,170 @@
...
+
+// Transaction step that disables tracing and resets the tracing config.
+func txnTracingDisable(c transaction.TxnCtx) error {
+	// Disable tracing
+	trace.ApplyConfig(trace.Config{DefaultSampler: trace.NeverSample()})
+
+	// Unregister the old Jaeger exporter if it exists
+	Exporter := tracing.JaegerExporter()
+	if Exporter == nil {
+		return errors.New("No Opencensus Exporter registered")
```
Madhu-1: error message with small case

glustercli/cmd/tracing.go
sseshasa: Sample output:
```
$./glustercli trace status --endpoints "http://192.168.122.10:24007" -v
+------------------------+----------------------------+
|      TRACE OPTION      |           VALUE            |
+------------------------+----------------------------+
| Status                 | enabled                    |
| Jaeger Endpoint        | http://192.168.122.1:14268 |
| Jaeger Agent Endpoint  | http://192.168.122.1:6831  |
| Jaeger Sampler         | 1 (Always)                 |
| Jaeger Sample Fraction | 1.00                       |
+------------------------+----------------------------+
```

glustercli/cmd/tracing.go
sseshasa: Updated here and in all other places as well.

glustercli/cmd/tracing.go
sseshasa: I have removed this function and modified the one under pkg/tracing/tracing.go and used that.

glustercli/cmd/tracing.go
sseshasa: Done. I have moved the common code to a function and used that.

pkg/tracing/tracing.go
sseshasa: There's only one function now under pkg/tracing/tracing.go and that is being used.

pkg/tracing/tracing.go
sseshasa: This global option is always modified under a transaction (under lock) as part of "enable", "update" and "delete" commands. When gd2 restarts on a node for example, the configuration is only read from etcd in the startup sequence and the in-memory config on that node is updated. Therefore, I don't see an issue with respect to this scenario as the node will get the latest configuration if trace config was update while the node was down. On the other hand, if config change happens just after it joins the etcd cluster, it will anyway get the updated configuration as part of the "enable", "update" or "delete" transaction.

plugins/tracemgmt/api/resp.go
sseshasa: I prefer leaving it the way it is for readability. Nesting will add another indirection to access members of the original structure and there are a lot of places where these members are accessed/modified.

Can I take this up as a separate change set? Please let me know.

plugins/tracemgmt/init.go
sseshasa: Generated the new endpoints.md.

plugins/tracemgmt/traceutils/store-utils.go
sseshasa: Done.

plugins/tracemgmt/traceutils/store-utils.go
sseshasa: Done.

plugins/tracemgmt/transactions.go
sseshasa: Done.

plugins/tracemgmt/transactions.go
sseshasa: Done.

plugins/tracemgmt/transactions.go
sseshasa: Done.

plugins/tracemgmt/transactions.go
sseshasa: Done.

plugins/tracemgmt/transactions.go
sseshasa: Done.

PR #1481, opened 2019-01-16
Set heal options to on in glustershd volfile

vpandey-RH:
Set options metadata-self-heal, entry-self-heal and data-self-heal (which are now set to off  by default)  in glustershd e2e tests to successfully heal. 
Change introduced in https://review.gluster.org/#/c/glusterfs/+/21938/.
The above three options are not set to off by default in glustershd volfile.


aravindavk:
@vpandey-RH found an issue in option validation function https://github.com/gluster/glusterd2/blob/master/glusterd2/options/options.go#L386 Please move the `return ErrInvalidArg` after the for loop.

aravindavk:
retest this please

PR #1483, opened 2019-01-17
added support for go modules

oshankkumar:
With the current go version 1.11,Modules allow to build
any project outside the GOPATH.

Refer for more details: [go modules](https://blog.golang.org/modules2019)


oshankkumar:
that is okay even if we are using go1.9X, but If any one using go1.11 and wants to just clone the repo and build glusterd2, then it will be helpful.

Madhu-1:
IMO we should not keep code which we are not testing as part of our CI.

if someone adds a new package using the current dep module, they need to change this also. 
as we don't have a CI to build based on this module it will become a non-tested code for us.

@aravindavk  @kshlm @atinmu

PR #1484, opened 2019-01-17
plugins/gluster-virtblock: Rename to virtual block

kotreshhr:
plugins/gluster-virtblk: Rename block to gluster-virtual block

### DIFF
--- a/plugins/blockvolume/blockprovider/gluster-virtblock/gluster_virtblock.go
+++ b/plugins/blockvolume/blockprovider/gluster-virtblock/gluster_virtblock.go
@@ -1,4 +1,4 @@
-package glusterloopback
+package glustervirtblock
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


Madhu-1:
add to whitelist

plugins/blockvolume/blockprovider/gluster-virtblk/glustervirtblk.go
```diff
@@ -17,28 +17,28 @@ import (
 	"k8s.io/kubernetes/pkg/util/mount"
 )
 
-const providerName = "gluster-loopback"
+const providerName = "gluster-virtblk"
```
aravindavk: Remove `gluster-` prefix and expand blk to block, so that in URL we can use `/v1/blockvolumes/virtblock`

Madhu-1:
@kotreshhr i  think either  you need to  update  commit message or change the folder name
currently it says `plugins/gluster-virtblock: Rename to virtual block`

i would suggest to  change the folder  name to `plugins/gluster-virtblock` as you  mentioned in commit message.

plugins/blockvolume/blockprovider/gluster-virtblk/glustervirtblk.go
kotreshhr: done

kotreshhr:
> @kotreshhr i think either you need to update commit message or change the folder name
> currently it says `plugins/gluster-virtblock: Rename to virtual block`
> 
> i would suggest to change the folder name to `plugins/gluster-virtblock` as you mentioned in commit message.

Done

PR #1490, opened 2019-01-23
xlators default orders and volume default options

### LINKED ISSUE #1485: disable glustershd for GCS 1.0
Currently GD2 doesn't restart self heal daemon for every volume operation which is functionality wise incorrect. However we end up restarting self heal process on a container storage environment, with hundreds of RWX PVCs that's going to cause the performance penalty. PR https://github.com/gluster/glusterd2/pull/1460 attempted to do so and the results were not promising. 

While the actual way to address this problem is to bring in multiplexing support to self heal daemon where with every volume operations the graph is attached/detached to self heal with out it being getting restarted but that work would require some more time before the GCS 1.0 timelines.

The mitigation plan is to disable self heal daemon and allow client self heal to ensure all the healing take place for the files which are accessed.

aravindavk:
Xlator order for client volfile changed to,

```
debug/io-stats
performance/read-ahead
performance/io-threads
performance/nl-cache
performance/quick-read
performance/open-behind
performance/io-cache
performance/readdir-ahead
performance/write-behind
performance/md-cache
features/utime
features/shard
cluster/distribute
```

This PR also disables Self heal by default.(Fixes #1485)

*(edited to remove ":" after "Fixes" so waffle.io tracks properly)*


Madhu-1:
retest this please

PR #1491, opened 2019-01-23
Add block hosting volume defaults

kotreshhr:


oshankkumar:
now we are using these options from cluster-options not from config file, so I think this PR can be closed.

PR #1493, opened 2019-01-23
blockVolume: use cluster-options for host volume configuration

oshankkumar:

### DIFF
--- a/plugins/blockvolume/hostvol_opts.go
+++ b/plugins/blockvolume/hostvol_opts.go
@@ -1,34 +1,37 @@
 package blockvolume
 
 import (
+	"strconv"
+
+	"github.com/gluster/glusterd2/glusterd2/options"
 	"github.com/gluster/glusterd2/pkg/api"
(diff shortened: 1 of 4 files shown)


plugins/blockvolume/hostvol_opts.go
```diff
@@ -1,34 +1,37 @@
...
-// VolumeType represents a volume type
-type VolumeType string
-
 const (
-	// Replica represents a replica volume type
-	Replica VolumeType = "Replica"
+	defaultHostVolsize         = size.GiB * 5
+	defaultHostVolType         = "Replicate"
+	defaultHostVolReplicaCount = 3
+	hostVolautoCreate          = true
```
poornimag: Are two copies of default values necessary(also set in glusterd2/options/cluster.go)? can it be nil values or ?

oshankkumar: when we set values using cluster options, there might be a possibility of error. So in case of any error we can simply use these default values.

PR #1497, opened 2019-02-07
BlockVolume add extra arguments to create and reduce number of cluster locks

poornimag:
With this the block volumes can be created on thin arbiter, shard volumes. 

### DIFF
--- a/glusterd2/volume/volume-utils.go
+++ b/glusterd2/volume/volume-utils.go
@@ -7,6 +7,7 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 	"syscall"
 
@@ -398,4 +399,29 @@ func CleanBricksLoop(volinfo *Volinfo) error {
 		}
 	}
 	return nil
+}
+
+//AddThinArbiter adds thin arbiter option to the volume create request
+func AddThinArbiter(req *api.VolCreateReq, thinArbiter string) error {
+
+	s := strings.Split(thinArbiter, ":")
+	if len(s) != 2 && len(s) != 3 {
+		return errors.New("thin arbiter brick must be of the form <host>:<brick> or <host>:<brick>:<port>")
+	}
+
+	// TODO: If required, handle this in a generic way, just like other
+	// volume set options that we're going to allow to be set during
+	// volume create.
+	req.Options["replicate.thin-arbiter"] = thinArbiter
+	req.AllowAdvanced = true
+	return nil
+}
+
+//AddShard adds shard options to the volume create request
+func AddShard(req *api.VolCreateReq, shardSize uint64) {
+
+	req.Options["features/shard"] = "on"
+	req.Options["features/shard.shard-block-size"] = strconv.FormatUint(shardSize, 10)
+	req.AllowAdvanced = true
+	return
 }
--- a/plugins/blockvolume/api/types.go
+++ b/plugins/blockvolume/api/types.go
@@ -11,11 +11,25 @@ type BlockVolumeInfo struct {
 	HaCount int    `json:"hacount,omitempty"`
 }
 
+// HostVolumeInfo represents block volume info
+type HostVolumeInfo struct {
+	// HostVolReplicaCnt represents the replica count of the block hosting volume
+	HostVolReplicaCnt int `json:"hostvolreplicacnt,omitempty"`
+	// HostVolThinArbPath represents the thin arbiter path
+	HostVolThinArbPath string `json:"hostvolthinarbpath,omitempty"`
+	// HostVolShardSize represents the shard size of the block hosting volume
+	HostVolShardSize uint64 `json:"hostvolshardsize,omitempty"`
+	// Size represents Block Hosting Volume size in bytes
+	HostVolSize uint64 `json:"hostvolsize,omitempty"`
+}
+
 // BlockVolumeCreateRequest represents req Body for Block vol create req
 type BlockVolumeCreateRequest struct {
 	*BlockVolumeInfo
-	Clusters []string `json:"clusters,omitempty"`
-	Auth     bool     `json:"auth,omitempty"`
+	HostVolumeInfo
+	BlockType string   `json:"blocktype,omitempty"`
+	Clusters  []string `json:"clusters,omitempty"`
+	Auth      bool     `json:"auth,omitempty"`
 }
 
 // BlockVolumeCreateResp represents resp body for a Block Vol Create req
--- a/plugins/blockvolume/blockprovider/gluster-virtblock/gluster_virtblock.go
+++ b/plugins/blockvolume/blockprovider/gluster-virtblock/gluster_virtblock.go
@@ -12,7 +12,7 @@ import (
 	"github.com/gluster/glusterd2/pkg/errors"
 	"github.com/gluster/glusterd2/pkg/utils"
 	"github.com/gluster/glusterd2/plugins/blockvolume/blockprovider"
-	blkUtils "github.com/gluster/glusterd2/plugins/blockvolume/utils"
+	"github.com/gluster/glusterd2/plugins/blockvolume/hostvol"
 
 	log "github.com/sirupsen/logrus"
 	config "github.com/spf13/viper"
@@ -95,18 +95,21 @@ func (g *GlusterVirtBlk) CreateBlockVolume(name string, size uint64, hostVolume 
 	blockFileName := hostDir + "/" + name
 	err = utils.ExecuteCommandRun("truncate", fmt.Sprintf("-s %d", size), blockFileName) //nolint: gosec
 	if err != nil {
-		logger.WithError(err).Errorf("failed to truncate block file" + blockFileName)
+		logger.WithError(err).Errorf("failed to truncate block file %s", blockFileName)
 		return nil, err
 	}
 
-	err = utils.ExecuteCommandRun("mkfs.xfs", "-f", blockFileName) //nolint: gosec
-	if err != nil {
-		logger.WithError(err).Errorf("failed to format block file " + blockFileName)
-		return nil, err
+	if blockVolOpts.BlockType != "raw" {
+		fsType := blockVolOpts.BlockType
+		err = utils.ExecuteCommandRun(fmt.Sprintf("mkfs.%s", fsType), "-f", blockFileName) //nolint: gosec
+		if err != nil {
+			logger.WithError(err).Errorf("failed to format block file %s with filesystem %s", blockFileName, fsType)
+			return nil, err
+		}
 	}
 
 	resizeFunc := func(blockHostingAvailableSize, blockSize uint64) uint64 { return blockHostingAvailableSize - blockSize }
-	if err = blkUtils.ResizeBlockHostingVolume(hostVolume, size, resizeFunc); err != nil {
+	if err = hostvol.ResizeBlockHostingVolume(hostVolume, size, resizeFunc); err != nil {
 		logger.WithError(err).Error("failed in updating hostvolume _block-hosting-available-size metadata")
 		return nil, err
 	}
@@ -119,7 +122,7 @@ func (g *GlusterVirtBlk) CreateBlockVolume(name string, size uint64, hostVolume 
 
 	volInfo, err := volume.GetVolume(hostVolume)
 	if err != nil {
-		logger.WithError(err).Errorf("failed to get host volume info " + hostVolume)
+		logger.WithError(err).Errorf("failed to get host volume info %s", hostVolume)
 		return nil, err
 	}
 	key := volume.BlockPrefix + name
@@ -181,14 +184,14 @@ func (g *GlusterVirtBlk) DeleteBlockVolume(name string, options ...blockprovider
 
 	hostDir, err := mountHost(g, blkVol.HostVolume())
 	if err != nil {
-		log.WithError(err).Errorf("error mounting block hosting volume :" + blkVol.HostVolume())
+		log.WithError(err).Errorf("error mounting block hosting volume :%s", blkVol.HostVolume())
 		return err
 	}
 
 	blockFileName := hostDir + "/" + name
 	err = os.Remove(blockFileName)
 	if err != nil {
-		log.WithError(err).Errorf("error removing block :" + blockFileName)
+		log.WithError(err).Errorf("error removing block :%s", blockFileName)
 		return err
 	}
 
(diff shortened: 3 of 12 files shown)


poornimag:
@oshankkumar @aravindavk Please review the patch. Thanks

aravindavk:
add to whitelist

plugins/blockvolume/hostvol/utils.go
```diff
@@ -15,34 +15,70 @@ import (
 	"github.com/gluster/glusterd2/glusterd2/volume"
 	"github.com/gluster/glusterd2/pkg/api"
 	"github.com/gluster/glusterd2/pkg/size"
-
 	log "github.com/sirupsen/logrus"
 )
 
+func filterSize(volinfo *volume.Volinfo, size uint64) bool {
```
oshankkumar: size filter is already present, please use the existing one.

plugins/blockvolume/hostvol/utils.go
```diff
@@ -15,34 +15,70 @@ import (
...
+		//Checking for shard size may not be required
+		return true
+	}
+	return false
+}
+
 // BlockSizeFilter returns a volume Filter, which will filter out volumes
 // haing block-hosting-available-size greater than give size.
-func BlockSizeFilter(size uint64) volume.Filter {
+func HostOptnsFilter(size uint64, h *HostingVolumeOptions) volume.Filter {
```
oshankkumar: write applyArbiter and applyShard separately as a filter function and this function can be omitted.

plugins/blockvolume/hostvol/utils.go
```diff
@@ -15,34 +15,70 @@ import (
...
 		return volumes
 	}
 }
 
 // GetExistingBlockHostingVolume returns a existing volume which is suitable for hosting a gluster-block
-func GetExistingBlockHostingVolume(size uint64) (*volume.Volinfo, error) {
+func GetExistingBlockHostingVolume(size uint64, h *HostingVolumeOptions) (*volume.Volinfo, error) {
 	var (
-		filters     = []volume.Filter{volume.FilterBlockHostedVolumes, BlockSizeFilter(size)}
+		filters     = []volume.Filter{volume.FilterBlockHostedVolumes, HostOptnsFilter(size, h)}
```
oshankkumar: use can use filters like []volume.Filter{volume.FilterBlockHostedVolumes, BlockSizeFilter(size),FilterArbiter,FilterShard}.
FilterArbiter,FilterShard need to be implemented

plugins/blockvolume/hostvol/utils.go
oshankkumar: you can add  FilterArbiter and FilterShard func in this file `github.com/gluster/glusterd2/glusterd2/volume/filters.go`

plugins/blockvolume/hostvol/utils.go
poornimag: I thought of the same, but if i were to apply filter separately then each filter has to loop though the list of all volumes right? So wrote it like this, for each volume apply all filters required. If the number of volumes are huge, this looping of volumes for each filter adds up redundancy? But yeah block hosting volumes are generally less in number, hence not much difference may be. Thought?

oshankkumar: right, if the number of volumes will be more then it will be a problem, okay then we can go with the current implementation.

plugins/blockvolume/hostvol/utils.go
poornimag: Done

plugins/blockvolume/hostvol/utils.go
poornimag: Done

plugins/blockvolume/hostvol/utils.go
poornimag: Done

plugins/blockvolume/api/types.go
```diff
@@ -11,9 +11,23 @@ type BlockVolumeInfo struct {
...
+// HostVolumeInfo represents block volume info
+type HostVolumeInfo struct {
+	// HostVolReplicaCnt represents the replica count of the block hosting volume
+	HostVolReplicaCnt int `json:"hostvolreplicacnt,omitempty"`
+	// HostVolThinArbPath represents the thin arbiter path
+	HostVolThinArbPath string `json:"hostvolthinarbpath,omitempty"`
+	// HostVolShardSize represents the shard size of the block hosting volume
+	HostVolShardSize uint64 `json:"hostvolshardsize,omitempty"`
+	// Size represents Block Hosting Volume size in bytes
+	HostVolSize uint64 `json:"hostsize,omitempty"`
```
Madhu-1: `%s/hostsize/hostvolsize`

plugins/blockvolume/api/types.go
```diff
@@ -11,9 +11,23 @@ type BlockVolumeInfo struct {
...
+}
+
 // BlockVolumeCreateRequest represents req Body for Block vol create req
 type BlockVolumeCreateRequest struct {
 	*BlockVolumeInfo
+	HostVolumeInfo
+	BlkType  string   `json:"blktype,omitempty"`
```
Madhu-1: what is `blktype`?

plugins/blockvolume/blockprovider/gluster-virtblock/gluster_virtblock.go
```diff
@@ -99,21 +99,33 @@ func (g *GlusterVirtBlk) CreateBlockVolume(name string, size uint64, hostVolume
 		return nil, err
 	}
 
-	err = utils.ExecuteCommandRun("mkfs.xfs", "-f", blockFileName) //nolint: gosec
-	if err != nil {
-		logger.WithError(err).Errorf("failed to format block file " + blockFileName)
-		return nil, err
+	if blockVolOpts.BlkType != "none" {
```
Madhu-1: why it is compared with none? can this be compared with an empty string which default none for string?

plugins/blockvolume/blockprovider/gluster-virtblock/gluster_virtblock.go
```diff
@@ -99,21 +99,33 @@ func (g *GlusterVirtBlk) CreateBlockVolume(name string, size uint64, hostVolume
...
+		fsType := blockVolOpts.BlkType
+		err = utils.ExecuteCommandRun(fmt.Sprintf("mkfs.%s", fsType), "-f", blockFileName) //nolint: gosec
+		if err != nil {
+			logger.WithError(err).Errorf("failed to format block file " + blockFileName + " with filesystem " + fsType)
```
Madhu-1: formatting required for `Errorf` 
```
logger.WithError(err).Errorf("failed to format block file %s with filesystem %s" , blockFileName, fsType)
```

plugins/blockvolume/blockprovider/gluster-virtblock/gluster_virtblock.go
```diff
@@ -99,21 +99,33 @@ func (g *GlusterVirtBlk) CreateBlockVolume(name string, size uint64, hostVolume
...
-	if err = blkUtils.ResizeBlockHostingVolume(hostVolume, size, resizeFunc); err != nil {
+	if err = hostvol.ResizeBlockHostingVolume(hostVolume, size, resizeFunc); err != nil {
 		logger.WithError(err).Error("failed in updating hostvolume _block-hosting-available-size metadata")
 		return nil, err
 	}
 
-	if err = clusterLocks.Lock(hostVolume); err != nil {
-		logger.WithError(err).Error("error in acquiring cluster lock")
-		return nil, err
+	for {
```
Madhu-1: infinite loop could cause an issue, which may lead to http timeout

plugins/blockvolume/blockprovider/gluster-virtblock/gluster_virtblock.go
```diff
@@ -142,9 +154,18 @@ func delBlockEntry(hostName string, name string) error {
 		clusterLocks = transaction.Locks{}
 	)
 
-	if err := clusterLocks.Lock(hostName); err != nil {
-		log.WithError(err).Error("error in acquiring cluster lock")
-		return err
+	for {
```
Madhu-1: same here

plugins/blockvolume/hostvol/hostvol_manager.go
```diff
@@ -55,21 +55,35 @@ func (g *glusterVolManager) GetHostingVolumesInUse() []*volume.Volinfo {
...
+func (g *GlusterVolManager) GetOrCreateHostingVolume(name string, minSizeLimit uint64, hostVolumeInfo *api.HostVolumeInfo) (*volume.Volinfo, error) {
 	var (
 		volInfo      *volume.Volinfo
 		clusterLocks = transaction.Locks{}
 	)
 
-	if err := clusterLocks.Lock(path.Join(globalLockID, name)); err != nil {
-		return nil, err
+	for {
+		err := clusterLocks.Lock(path.Join(globalLockID))
```
Madhu-1: same here

poornimag:
> what if the hostvolume is different? then we should not block.

Even when the host volume is different we still need global lock for creating host volume. Consider a situation, where there are two block create requests issued simultaneously, then we have two block host host volumes created. Hence we need a global lock for creating block host volume. The only time we may not need it, is when the bhv is passed a input in the block create request. Is that the case you are talking about?

plugins/blockvolume/hostvol/hostvol_manager.go
oshankkumar: if host volume name is empty, then `path.Join(globalLockID, name)` resolves in `globalLockID` only. So I don't think this change is required.

plugins/blockvolume/api/types.go
poornimag: It will contain the filesystem type the block will be formatted with, if its "none" then the file will not be formatted with any filesystem. The default is "xfs".

plugins/blockvolume/blockprovider/gluster-virtblock/gluster_virtblock.go
poornimag: "none" is used to represent that no formatting is required, if its "" then the default is "xfs".

plugins/blockvolume/api/types.go
Madhu-1: can we make it as `block-type` for more readability and set a default value on server side based on the empty string will be a good close instead of "none"

plugins/blockvolume/blockprovider/gluster-virtblock/gluster_virtblock.go
poornimag: So we try for few iterations, and then return with LockTimeout? Then the csi should handle this error and retry?

plugins/blockvolume/blockprovider/gluster-virtblock/gluster_virtblock.go
poornimag: Done

plugins/blockvolume/blockprovider/gluster-virtblock/gluster_virtblock.go
poornimag: Done

plugins/blockvolume/blockprovider/gluster-virtblock/gluster_virtblock.go
Madhu-1: sorry, am not aware of no formatting us required use case can you explain?

plugins/blockvolume/blockprovider/gluster-virtblock/gluster_virtblock.go
Madhu-1: if volume creation fails, CSI will do a retry

plugins/blockvolume/hostvol/hostvol_manager.go
poornimag: Ah ok. @aravindavk The block host volume creation is already serialized, then how can there be stale bricks? Before creating bhv it checks for alerady existing bhv under lock. So its not root cause for stale bricks?

plugins/blockvolume/blockprovider/gluster-virtblock/gluster_virtblock.go
poornimag: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#volume-mode  This one? VolumeMode can be raw as well, not sure about further deatils.

plugins/blockvolume/api/types.go
poornimag: Sure, may be will use "raw" instead of "none"?

plugins/blockvolume/hostvol/hostvol_manager.go
oshankkumar: parallel req for BHV creation is already taken care of. This is not the root cause for any stale bricks.

poornimag:
@Madhu-1 @oshankkumar Have addressed the comments, and got rid of patches with infinite loop for lock and bhv lock id change. Any other changes required?

plugins/blockvolume/blockprovider/gluster-virtblock/gluster_virtblock.go
```diff
@@ -179,9 +155,10 @@ func (g *GlusterVirtBlk) DeleteBlockVolume(name string, options ...blockprovider
 		return errors.ErrBlockVolNotFound
 	}
 
-	hostDir, err := mountHost(g, blkVol.HostVolume())
+	hostName := blkVol.HostVolume()
+	hostDir, err := mountHost(g, hostName)
 	if err != nil {
-		log.WithError(err).Errorf("error mounting block hosting volume :" + blkVol.HostVolume())
+		log.WithError(err).Errorf("error mounting block hosting volume :" + hostName)
```
Madhu-1: formatting  is required for Errorf

oshankkumar:
LGTM

glusterd2/volume/volume-utils.go
```diff
@@ -399,3 +400,28 @@ func CleanBricksLoop(volinfo *Volinfo) error {
...
+	return nil
+}
+
+//AddShard adds shard options to the volume create request
+func AddShard(req *api.VolCreateReq, shardSize uint64) error {
+
+	req.Options["features/shard"] = "on"
+	req.Options["features/shard.shard-block-size"] = strconv.FormatUint(shardSize, 10)
+	req.AllowAdvanced = true
+	return nil
```
Madhu-1: what's the use of returning an error here? it's always nil

poornimag: Nothing, just to keep it similar to AddThinArbiter.., Will change

poornimag:
@Madhu-1 Anyother changes? Thanks

poornimag:
@aravindavk Can you merge this patch?

PR #1499, opened 2019-02-22
transaction: adding txns under pending-transaction/transaction/ key

oshankkumar:
initiator will add trasactions under pending-transaction/transaction/
key. Txn Engine keeps a watch on this key for new transactions
and execute transactions.

eg.. etcd key for storing details of a single transaction:-
`pending-transaction/transaction/<txn-ID>`

### DIFF
--- a/glusterd2/transactionv2/txnmanager.go
+++ b/glusterd2/transactionv2/txnmanager.go
@@ -22,6 +22,11 @@ import (
 var GlobalTxnManager TxnManager
 
 const (
+	// PendingTxnPrefix is the etcd namespace into which all pending txn info will be stored
+	PendingTxnPrefix = "pending-transaction"
+	// TxnPrefix is etcd key prefix under which details of a single txn will added
+	// eg.. key for storing details of a single transaction:- pending-transaction/transaction/<txn-ID>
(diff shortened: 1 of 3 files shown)


oshankkumar:
retest this please

glusterd2/transactionv2/txnmanager.go
```diff
@@ -229,16 +224,11 @@ func (tm *txnManager) GetTxnByUUID(id uuid.UUID) (*Txn, error) {
 
 // GetTxns returns all txns added to the store
 func (tm *txnManager) GetTxns() (txns []*Txn) {
-	resp, err := store.Get(context.TODO(), tm.getStoreKey(), clientv3.WithPrefix())
+	resp, err := store.Get(context.TODO(), tm.getStoreKey(txnPrefix), clientv3.WithPrefix())
```
kshlm: You should be using `TxnPrefix` with `getStoreKey` here. This works right now because of a `txnPrefix` defined in transaction.go. You should remove that and make use of a common constant.

oshankkumar: done

PR #1503, opened 2019-03-11
Add gluster-rsyslog Alpine container

sseshasa:
This is a lightweight container based on Alpine Linux and run as a
sidecar container in a gcs cluster along with containers running gd2.
This is mainly used to parse gluster specific logs and normalize it for
elastic search. Right now only gd2 logs are parsed to some extent. For
this, an rsyslog configuration along with a rulebase file are included.

### DIFF
--- a/extras/rsyslog-container/rsyslog.conf
+++ b/extras/rsyslog-container/rsyslog.conf
@@ -1,0 +1,26 @@
+module(load="imfile" PollingInterval="10")
+module(load="mmnormalize")
+input(type="imfile"
+      file="/var/log/glusterd2/glusterd2.log"
+      tag="glusterd2"
+      addmetadata="on"
+      reopenOnTruncate="on"
+      discardTruncatedMsg="on"
+      msgDiscardingError="off")
+action(type="mmnormalize"
+       rulebase="/etc/rsyslog.d/gd2.rulebase")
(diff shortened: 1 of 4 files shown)


sseshasa:
@aravindavk @kshlm PTAL. 
@sidharthanup please note the change in repo.

extras/rsyslog-container/rsyslog.conf
```diff
@@ -0,0 +1,26 @@
+module(load="imfile" PollingInterval="10")
+module(load="mmnormalize")
+input(type="imfile"
+      file="/var/log/glusterd2/glusterd2.log"
```
aravindavk: This can be taken from env variable

kshlm:
add to whitelist

extras/rsyslog-container/rsyslog.conf
sseshasa: I experimented a bit by creating a container specific config file that contains the env variable(s) and then source this file when the rsyslog container starts. The env variables can then by accessed from within rsyslog.conf by following a particular syntax. But all this appears to be still in experimental stage as far as rsyslog is concerned as during testing, the rsyslog container did not come up properly. I think it would be better to wait until this feature of rsyslog is stable and keep things as is for now. I will keep on the lookout for this.

